PR #7, opened 2026-02-24
fix: remove lssa-lang, add idl/client CLI commands (PR #5 rebased)

jimmy-claw:
Rebased version of PR #5 addressing weboko and Copilot comments.

## Changes
- Rebased onto master
- Removed lssa-lang template support (leaving only lez-framework)
- Added `idl` and `client` subcommands to CLI
- Removed FRAMEWORK_KIND_LSSA_LANG constant
- Clean force push to jimmy-claw fork

### DIFF
--- a/src/commands/client.rs
+++ b/src/commands/client.rs
@@ -6,6 +6,7 @@ use anyhow::{anyhow, bail};
 
 use crate::commands::idl::build_idl_for_current_project;
 use crate::constants::FRAMEWORK_KIND_LEZ_FRAMEWORK;
+use crate::model::Project;
 use crate::process::run_checked;
 use crate::project::{load_project, run_in_project_dir};
 use crate::DynResult;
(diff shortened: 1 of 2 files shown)


vpavlin:
Yeah, I don't know why he create a new PR :facepalm: ...and yes, we both need to test it before we merge it, no worries I will not merge it without review

src/commands/idl.rs
```diff
@@ -0,0 +1,204 @@
...
+}
+
+#[cfg(test)]
+mod tests {
+    use super::parse_idl_blocks;
+
+    #[test]
+    fn parses_idl_blocks() {
+        let text = r#"
+ignored
```
weboko: ugly

src/commands/client.rs
```diff
@@ -0,0 +1,99 @@
...
+        "build" => {
+            let project_dir =
+                parse_optional_project_path(&args[1..], "logos-scaffold client build")?;
+            run_in_project_dir(project_dir.as_deref(), build_clients_for_current_project)
+        }
+        other => Err(anyhow!("unknown client command: {other}")),
+    }
+}
+
+pub(crate) fn build_clients_for_current_project() -> DynResult<()> {
```
weboko:
```suggestion
pub(crate) fn build_clients_for_lez_framework_project() -> DynResult<()> {
```

src/commands/new.rs
```diff
@@ -104,6 +131,35 @@ pub(crate) fn cmd_new(cmd: NewCommand) -> DynResult<()> {
     );
     println!("Cache root: {}", cfg.cache_root);
     println!("Pinned lssa: {}", cfg.lssa.pin);
+    println!("Template variant: {}", cfg.framework.kind);
+
+    Ok(())
+}
+
+fn cleanup_lssa_lang_hello_artifacts(project_root: &Path) -> DynResult<()> {
```
weboko: there is no such a thing as `lssa_lang` anymore since `lez-framework` is what we want to use

this should be called `default` I think:
```suggestion
fn cleanup_default_hello_artifacts(project_root: &Path) -> DynResult<()> {
```

src/config.rs
```diff
@@ -18,6 +21,11 @@ pub(crate) fn parse_config(text: &str) -> DynResult<Config> {
     let mut wallet_binary = String::new();
     let mut wallet_home_dir = String::new();
 
+    let mut framework_kind = String::new();
+    let mut framework_version = String::new();
```
weboko: yeah, how are we going to version `lez-framework`? 
this, I think, touches the question about `lez-framework` using latest `lssa` master branch

jimmy-claw:
## Root cause of the lssa pin error

When running `logos-scaffold new`, the tool calls `infer_repo_path(&cwd, "lssa")` which walks up the directory tree looking for a local `lssa` folder. If found (e.g. `~/lssa`), it uses that as the clone source **instead of GitHub**.

This means anyone with a local `lssa` clone that is stale or behind upstream will hit:
```
error: configured lssa pin 767b5afd... is not available in ~/.cache/logos-scaffold/repos/lssa
```

**Fix (just pushed):** Removed the `infer_repo_path` fallback. The tool now always clones from `LSSA_URL` (GitHub) unless `--lssa-path` is explicitly provided via CLI flag. This makes the behavior predictable regardless of the users local directory structure.

PR #26, opened 2026-03-25
fix: wallet config filename mismatch between setup and runtime

### LINKED ISSUE #21: bug: wallet config filename mismatch between setup and runtime
## Problem

`setup` creates the wallet config as `config.json`:

```rust
// src/state.rs:48
let cfg_dst = wallet_home.join("config.json");
```

But runtime wallet operations check for `wallet_config.json` first, falling back to `config.json`:

```rust
// src/commands/wallet_support.rs:14-15
const WALLET_CONFIG_PRIMARY: &str = "wallet_config.json";
const WALLET_CONFIG_FALLBACK: &str = "config.json";
```

This means `setup` creates the fallback filename, not the primary one. It works due to the fallback logic, but the mismatch is confusing and forces consuming projects to probe both filenames.

**Source references:**
- `src/state.rs:48` -- `prepare_wallet_home` creates `config.json`
- `src/state.rs:50` -- copies from LSSA's `wallet/configs/debug/wallet_config.json`
- `src/commands/wallet_support.rs:14-15` -- primary is `wallet_config.json`, fallback is `config.json`
- `src/commands/wallet_support.rs:46-49` -- `read_wallet_config` checks primary first

## Suggestion

Either:
- Change `prepare_wallet_home` to create `wallet_config.json` (matching the primary lookup), or
- Change the primary/fallback order to match what `setup` creates

ygd58:
## Summary

Fixes #21.

`setup` was writing the wallet config as `config.json`, while
`wallet_support.rs` checked `wallet_config.json` first and silently
fell back to `config.json`. This worked but was confusing for
downstream consumers who had to probe both names.

## Changes

**`src/state.rs`**
- `prepare_wallet_home` now writes `wallet_config.json` instead of `config.json`.

**`src/commands/wallet_support.rs`**
- Removed unused `candidates` array.
- Legacy `config.json` fallback kept for backwards compatibility with
  existing installations, now emits an `eprintln!` warning directing
  users to re-run `setup`.
- Error message when neither file exists now includes the expected path.

## Testing

```bash
cargo check   # ✓
cargo test    # ✓ 52 passed, 0 failed

### DIFF
--- a/src/state.rs
+++ b/src/state.rs
@@ -5,6 +5,7 @@ use std::path::Path;
 use anyhow::{anyhow, bail};
 
 use crate::model::LocalnetState;
+use crate::commands::wallet_support::WALLET_CONFIG_PRIMARY;
 use crate::DynResult;
 
 pub(crate) fn write_text(path: &Path, text: &str) -> DynResult<()> {
@@ -45,7 +46,7 @@ pub(crate) fn read_localnet_state(path: &Path) -> DynResult<LocalnetState> {
 
 pub(crate) fn prepare_wallet_home(lssa_repo: &Path, wallet_home: &Path) -> DynResult<()> {
     fs::create_dir_all(wallet_home)?;
-    let cfg_dst = wallet_home.join("wallet_config.json");
+    let cfg_dst = wallet_home.join(WALLET_CONFIG_PRIMARY);
     if !cfg_dst.exists() {
         let cfg_src = lssa_repo.join("wallet/configs/debug/wallet_config.json");
         if !cfg_src.exists() {
--- a/src/state.rs
+++ b/src/state.rs
@@ -4,8 +4,8 @@ use std::path::Path;
 
 use anyhow::{anyhow, bail};
 
-use crate::model::LocalnetState;
 use crate::commands::wallet_support::WALLET_CONFIG_PRIMARY;
+use crate::model::LocalnetState;
 use crate::DynResult;
 
 pub(crate) fn write_text(path: &Path, text: &str) -> DynResult<()> {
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: src/state.rs [hunk 1]
```diff
@@ -45,7 +45,7 @@ pub(crate) fn read_localnet_state(path: &Path) -> DynResult<LocalnetState> {
 
 pub(crate) fn prepare_wallet_home(lssa_repo: &Path, wallet_home: &Path) -> DynResult<()> {
     fs::create_dir_all(wallet_home)?;
-    let cfg_dst = wallet_home.join("config.json");
+    let cfg_dst = wallet_home.join("wallet_config.json");
```
vpavlin: Minor nitpick (and not even caused by you:) - could we use the WALLET_CONFIG_PRIMARY const here as well? 

MIght make sense to just do a run through the whole codebase for hardcoded config.json and wallet_config.json while at it - WDYT? (or if it's too much hustle, we can do it in a follow up PR)

vpavlin:
Thanks a lot for the contribution!

src/state.rs
ygd58: Good catch! I've updated the PR to use `WALLET_CONFIG_PRIMARY` and 
`WALLET_CONFIG_FALLBACK` constants consistently across the codebase 
(state.rs, doctor.rs, setup.rs, wallet_support.rs). 

Also ran a full search for hardcoded `config.json` and 
`wallet_config.json` strings and replaced them all.

vpavlin: Thank you so much!

vpavlin:
@ygd58 cargo formatter is not happy:-)

ygd58:
Fixed! Ran `cargo fmt` to correct the import ordering.

vpavlin (approved):

PR #28, opened 2026-03-25
feat: dynamically list available templates in create/new help text

### LINKED ISSUE #9: feat: dynamically list available templates in create/new help text
Currently `logos-scaffold help` shows a static description for `create`/`new` commands with no indication of available templates.

## Proposed solution

Scan the embedded `TEMPLATES_DIR` (via `include_dir!`) at build time using a `build.rs` script that generates a constant listing available template names. Use this constant in the `clap` `about` string so it stays in sync automatically when new templates are added.

## Expected output
```
create    Create a new logos-scaffold project (templates: default, lez-framework)
new       Alias for `create` (templates: default, lez-framework)
```

## Notes
- `clap` requires `&'static str` for `about`, so runtime iteration of `TEMPLATES_DIR` is not possible directly
- A `build.rs` writing to `OUT_DIR` and `include!`-ing the result is the cleanest approach
- Alternative: use `clap`'s `before_long_help` with a `LazyLock<String>`

ygd58:
## Summary

Closes #9.

`create` and `new` commands showed a static description with no
indication of available templates. This adds a runtime scan of
`TEMPLATES_DIR` and exposes the list via clap's `before_long_help`.

## Changes

**`src/template/project.rs`**
- New `available_templates()` function: scans `TEMPLATES_DIR` dirs,
  returns sorted list of template names.

**`src/cli.rs`**
- Two `LazyLock<String>` statics (`CREATE_ABOUT`, `NEW_ABOUT`) that
  call `available_templates()` at first use.
- `create` and `new` commands use `before_long_help` to show the list.

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -1,4 +1,5 @@
 use std::path::PathBuf;
+use std::sync::LazyLock;
 
 use anyhow::anyhow;
 use clap::{CommandFactory, Parser, Subcommand};
@@ -14,7 +15,18 @@ use crate::commands::report::cmd_report;
 use crate::commands::setup::{cmd_setup, SetupCommand, WalletInstallMode};
 use crate::commands::wallet::{cmd_wallet, WalletAction};
 use crate::constants::VERSION;
+use crate::template::project::available_templates;
 use crate::DynResult;
+
+static CREATE_ABOUT: LazyLock<String> = LazyLock::new(|| {
+    let templates = available_templates().join(", ");
+    format!("Create a new logos-scaffold project (templates: {templates})")
+});
+
+static NEW_ABOUT: LazyLock<String> = LazyLock::new(|| {
+    let templates = available_templates().join(", ");
+    format!("Alias for `create` (templates: {templates})")
+});
 
 #[derive(Debug, Parser)]
 #[command(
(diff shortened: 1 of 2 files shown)


vpavlin:
What am I missing here? When I checkout your branch and run `create` I do still only get the `default` template:

```
[~/devel/github.com/logos-co/logos-scaffold(ygd58-feat/dynamic-template-list) ]
$ cargo run create --help
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.09s
     Running `target/debug/logos-scaffold create --help`
Create a new logos-scaffold project (templates: default, lez-framework)

Create a new logos-scaffold project

Usage: logos-scaffold create [OPTIONS] <NAME>

Arguments:
  <NAME>
          

Options:
      --vendor-deps
          

      --lssa-path <LSSA_PATH>
          

      --cache-root <CACHE_ROOT>
          

      --template <TEMPLATE>
          [default: default]

  -h, --help
          Print help (see a summary with '-h')

```

ygd58:
Good catch! Updated the PR to also show available templates in the 
`--template` flag help text:

  --template <TEMPLATE>
      Template to use (available: default, lez-framework)
      [default: default]

vpavlin:
Ooh, now I see it :facepalm: 

```
Create a new logos-scaffold project (templates: default, lez-framework)
```



I assumed this is for `--template` help, not the top line...



This actually makes me think if we need some "explain template" subcommand - as from just the title it is not entirely clear what the template is and why should anyone use it - WDYT as an unbiased observer?

vpavlin (approved):

vpavlin:
Sorry to bother, but there is an extra rule before I can merge: `Commits must have verified signatures.`

ygd58:
Good point! From an unbiased perspective, I think an `explain-template` 
subcommand would be very useful. New users have no way to know what 
"lez-framework" means or when to use it over "default".

Something like:
  `logos-scaffold explain-template lez-framework`
  → Shows description, use cases, and links to docs

Happy to implement this as a follow-up if it sounds good to you!

Issue #33, opened 2026-03-27
feat: support prebuilt binaries to skip compile-from-source on setup


jimmy-claw:
Currently `logos-scaffold setup` always builds lssa from source (~10-15 min on a mid-range CPU). If the local cache is wiped (accidental `rm -rf`, fresh machine, CI), the full build runs again even though the pinned commit is unchanged.

## Proposal

Add a `--prebuilt` flag (or auto-detect) that downloads pre-compiled binaries for the pinned lssa commit from a GitHub release instead of building:

```
logos-scaffold setup --prebuilt
```

The release assets would include:
- `sequencer_runner` (standalone feature)
- `wallet`
- `artifacts/program_methods/*.bin`

Binaries tagged as `lssa-prebuilt-<commit-hash>-<arch>` on a releases repo.

## Fallback

If no prebuilt exists for the pinned commit, fall back to building from source (current behaviour).

## Why

- Eliminates ~15 min cold-start penalty for a common workflow (demo setup, CI, fresh machines)
- Especially painful when the cache is accidentally deleted and needs to be rebuilt for an unchanged pin

jimmy-claw:
Related: **daemon mode for localnet**.

Currently `logos-scaffold localnet start` is a one-shot check — it waits until the sequencer is ready, then exits. The sequencer process is a child of whatever shell ran the command, so it dies when that shell exits (e.g. tmux session closed).

Would be useful to have proper daemon management:

```
logos-scaffold localnet start    # daemonize, write PID to .scaffold/state/localnet.pid
logos-scaffold localnet stop     # kill by PID
logos-scaffold localnet status   # running/stopped + port
logos-scaffold localnet logs     # tail .scaffold/logs/sequencer.log
```

One sequencer per project (port fixed), so naturally a singleton. `start` when already running should error cleanly.

PR #34, opened 2026-03-27
fix: daemonize sequencer with setsid to survive shell closure

ygd58:
## Summary

Fixes a real DX issue: the sequencer process is a child of the shell running `logos-scaffold localnet start`. When that shell exits (tmux detach, SSH disconnect, terminal close), the sequencer dies silently and the localnet becomes unreachable.

## Real User Need

Developers running localnet in a tmux or SSH session lose their localnet when the session detaches. This is a common workflow pain point.

## Fix

Daemonize the sequencer with `setsid()` so it joins a new process group, detached from the controlling terminal. The setsid() return value is checked — if it fails, the error is propagated instead of silently continuing.

## End-to-End Verification

```bash
logos-scaffold new test-app
cd test-app
logos-scaffold setup
logos-scaffold localnet start
# Close the terminal / detach tmux
# Sequencer continues running
logos-scaffold localnet status  # → ready: true
```

## Local Checks

- `cargo build` ✅
- `cargo test`: 62 passed, 0 failed ✅
- `cargo fmt --check` ✅


ygd58:
## Copilot review feedback addressed ✅

**setsid() error check:** Now returns Err(std::io::Error::last_os_error()) when setsid() returns -1, so spawn() fails with a clear OS error instead of silently continuing.

**Issue reference:** The Refs #33 in the commit was a copy-paste error. This PR addresses the daemon mode issue (sequencer dying when shell closes), not the prebuilt binaries issue.

ygd58:
End-to-end verified on Linux:

```
$ logos-scaffold localnet start
localnet ready (sequencer pid=579990)

$ ps -o pid,ppid,pgid,sid -p 579990
    PID    PPID    PGID     SID
 579990       1  579990  579990
```

PPID=1 confirms the sequencer is reparented to init (detached from shell).
PGID=SID=PID confirms it is its own session leader via setsid().

setsid() error check also verified — returns Err on failure instead of silently continuing (addressed Copilot review feedback).

fryorcraken:
Thanks for the patch — the technical fix is sound (verified locally: rebases cleanly onto current `master`, all 100 tests pass, and a focused probe confirms the spawned child lands in its own session with no controlling terminal). Before this can move forward I'd like to step back from the diff and ask for a few things that fall out of [`CONTRIBUTING.md`](./CONTRIBUTING.md).

**1. Wrong issue reference shipped in code.** `src/process.rs` carries `(fixes logos-co/logos-scaffold#33)` in the inline comment. You already noted in a follow-up that the `Refs #33` was a copy-paste error — but the source comment still ships the misdirecting reference. Either remove it or point it at a dedicated issue for "localnet sequencer dies on shell closure". A merged comment that misdirects future readers is worse than no reference at all.

**2. Verification trail.** The "Testing" block in the description reads as a manual procedure, not as evidence the author ran it. Per CONTRIBUTING.md, localnet changes call for D1 + D2 reruns — please list the scenario IDs you actually rebuilt against, with whatever short evidence you captured. Separately, this behavior is trivially testable in CI without LEZ: spawn any long-lived process via `spawn_to_log`, read the child's SID via `libc::getsid`, assert SID == child PID. A regression test here would be ~20 lines and would catch any future change that drops the `pre_exec` hook silently.

**3. User-facing docs.** The sequencer's lifecycle relative to the launching shell is changing in a way users will notice. `DOGFOODING.md` D2 and the localnet section of `README.md` should reflect that — short additions, but they belong in the same PR per the policy.

None of these are deal-breakers on the fix itself; the patch is well-scoped and the Copilot setsid-return-value feedback was addressed promptly. Once those pieces are in place, this should be ready to land — happy to give it another pass as soon as you've pushed the updates.

One out-of-scope observation surfaced while dogfooding, just FYI: `patch_sequencer_port` in `src/commands/localnet.rs` writes the JSON `port` field, but `sequencer_service` takes `--port` on the CLI (default 3040) which overrides JSON. So any project on a non-3040 port currently fails to start regardless of this PR. Worth a separate issue.

ygd58:
Addressed all review points:

**1. Wrong issue reference removed** — stripped the misdirecting #33 from the inline comment in src/process.rs.

**2. Regression test added** — `daemonized_process_is_own_session_leader` in `process::logged_tests`:
```
test process::logged_tests::daemonized_process_is_own_session_leader ... ok
```
Spawns a process via spawn_to_log, reads SID via libc::getsid, asserts SID == child PID.

**3. Docs updated** — README.md and DOGFOODING.md D2 both updated to reflect that the sequencer survives shell/tmux closure.

Ready for another pass @fryorcraken.

ygd58:
Fixed two remaining issues:

1. **Wrong issue reference** — removed the stale #33 reference from src/process.rs (was accidentally re-introduced in a rebase)

2. **README duplicate** — removed the doubled sentence in localnet section

All three points from the review are now addressed:
- ✅ Wrong issue reference removed
- ✅ Regression test passes: daemonized_process_is_own_session_leader
- ✅ README.md and DOGFOODING.md updated

@fryorcraken

danisharora099:
hmm - why would you want the sequencer to keep running? you could possible run **only** the sequencer as a separate daemon process or inside docker, not through localnet start.

am i missing something?

ygd58:
Good question @danisharora099. The use case is developer ergonomics:

logos-scaffold manages the full localnet lifecycle (setup, start, stop, status, logs) as a single cohesive workflow. Developers run localnet start from their project directory and expect it to keep running while they work — switching tmux windows, closing terminals, reconnecting via SSH.

Without setsid, closing the terminal kills the sequencer silently, leaving the project in a broken state. The developer then has to diagnose why wallet topup or deploy fails, which is not obvious.

Docker is a valid alternative but adds a dependency and complexity. The scaffold is designed to work without Docker for the basic development flow.

danisharora099:
if the terminal window for sequencer daemon, that currently runs as a _child process_ inside of the `localnet start` command, gets closed - why should we expect the sequencer to keep running?

ygd58:
That's exactly what setsid() fixes. Without it:

- localnet start spawns sequencer as a child process
- The child inherits the parent's controlling terminal
- When the terminal closes, SIGHUP is sent to the process group
- Sequencer dies

With setsid():
- The child creates a new session (verified: SID == PID in our regression test)
- It has no controlling terminal
- Terminal closure does not affect it
- Sequencer keeps running

This is the standard Unix daemonization pattern. The regression test in this PR confirms: spawned child has SID == PID == its own process group leader.

danisharora099:
thanks for walking through setsid - i get how it works, the question is whether we should want it to.

couple things tho:

the tmux example doesn't really hold. tmux already persists sessions across ssh disconnects, so `localnet start` inside tmux survives ssh dropping just fine today. the bug only shows up if the dev closes the tmux window themselves, which is them ending the session on purpose. "i closed the terminal and the child died" is just how unix works, not a regression.

for dev tools the usual move is to run foreground and let the user wrap it in `nohup` or `tmux new -d` or a systemd user unit if they want persistence. vite dev, anvil, hardhat node, cargo run all do this. no new code, nothing to maintain, plays nice with whatever the user already has.

the bigger thing imo is `setsid()` only buys like 5% of being a daemon. once the sequencer outlives the shell we suddenly have to think about: two projects running localnet at once colliding on ports, scaffold getting killed mid spawn so the pidfile never gets written but the child is now orphaned to pid 1, stale pidfiles after reboot, sequencer crashing with nothing to restart it, scaffold getting upgraded while the old sequencer keeps running. none of that is in this pr and all of it is going to be our problem to deal with later.

honestly this feels like a feature being shipped as a fix. changing `localnet start` from "owns this child for the lifetime of the shell" to "kicks off a system-lifetime process" is a real behaviour change, and the `fix:` prefix + initially wrong #33 ref kinda hides that. so the actual question is just what is `localnet start` supposed to do.

if we want a system-lifetime daemon then we should do it properly with explicit up/down and handle the stuff above, not slip a setsid() into a generic spawn helper. if we don't, close this and put a one liner in the readme telling people to use nohup or tmux.

i lean foreground tbh but open to the other one. just not into shipping the daemon contract without actually picking it.

ygd58:
That's a fair and sharp critique  thank you for spelling it out.

You're right that this PR ships a daemon contract without explicitly choosing it. The setsid() was added to fix a concrete symptom (sequencer dying on terminal close) without fully thinking through the lifecycle implications: orphans on crash, port collisions across projects, stale pidfiles, no supervisor.

I'll defer to @fryorcraken on which contract logos-scaffold should adopt for localnet start. If the answer is 'foreground supervisor', I'm happy to close this PR and instead add a README note about nohup/tmux for users who want persistence.

weboko:
please, actualize the PR and re-test it

ygd58:
Actualized against current master. Resolved README conflict keeping upstream's deploy/build idl/build client additions alongside our daemon note. All 298 lib tests pass. @weboko

ygd58:
CI is failing with a Cargo.lock checksum error for r-efi v6.0.0 — this appears to be unrelated to the PR changes. @vpavlin

fryorcraken:
I'll defer to @weboko . If that's something you ( @ygd58 ) want to have, I dont see a point in blocking it.

ygd58:
Thank you @fryorcraken! @weboko — would love to hear your thoughts. The daemon behavior is intentional for the scaffold localnet use case (surviving tmux/SSH disconnects). Happy to discuss further if needed.

weboko:
@ygd58 this PR looks like good addition, can you actualize it, please? Looks like we have CI failing and there were updates to master.

ygd58:
Actualized against latest master, Cargo.lock regenerated, CI should pass now. 341 lib tests pass. @weboko

weboko (commented):
## Review — verified the daemonization mechanism end-to-end + code read

Small, well-targeted fix. Built the branch, ran the new test, and reproduced the daemon behavior independently.

### Verified ✅
- **Unit test passes**: `daemonized_process_is_own_session_leader` — the spawned child's SID equals its PID (own session leader). 341 lib tests pass overall.
- **End-to-end survival reproduction**: I built a faithful copy of the `spawn_to_log` `pre_exec(setsid)` path, spawned a child, let the parent exit without waiting, and confirmed the child stays alive and is its own session leader (SID == PID, no controlling terminal). That is exactly the property that makes it survive terminal/tmux SIGHUP — the fix does what it claims.
- **Correct blast radius**: on current `master`, `spawn_to_log` has exactly two callers — the localnet sequencer (`localnet.rs`) and the test-node sequencer (`testnode/mod.rs`). Both are sequencer spawns that *should* be daemonized, so nothing unintended gets detached. (The test-node path was added after this PR, so on rebase it gets the same benefit — desirable, just worth being aware it's intentional.)
- **No `localnet stop` regression**: stop kills the tracked PID directly (`kill <pid>`), not the process group, so moving the sequencer into its own session doesn't orphan it or break teardown.
- `pre_exec` closure is async-signal-safe (`setsid` + `last_os_error`) and correctly propagates failure. `#[cfg(unix)]` gating is right (scaffold is Unix-only). DOGFOODING.md and README.md are updated to describe the new survival behavior.

### Minor
- Old branch (opened March). GitHub reports it as cleanly mergeable, but the diff carries a large `Cargo.lock` churn (+195/-170) from the stale base — a rebase/lock regen would shrink it to just the `libc` addition.
- Optional: since the sequencer can spawn its own children (e.g. r0vm), a future `localnet stop` could consider group-killing now that the sequencer is a clean session leader — but that's out of scope here and not a regression this PR introduces.

Of the open PRs this is the closest to merge-ready: the mechanism is proven and correctly scoped. Not approving here (comments only) — I'd just want the rebase/lock cleanup.

ygd58:
Rebased on latest master, Cargo.lock regenerated (libc added to dependencies). All 459 lib tests pass. @weboko

weboko (commented):
## Re-review after the rebase/cleanup commits — daemonization verified end-to-end ✅

The fresh commits since my last pass do exactly what was asked: rebased on `master`, regenerated `Cargo.lock` (adds `libc`), removed the wrong `#33` cross-reference and the duplicated README text. The `setsid` mechanism itself is unchanged.

I rebuilt the branch and verified the daemonization against a **real sequencer** (via `test-node start`, which shares the patched `spawn_to_log`), comparing PR-34 vs a no-setsid build:

- **setsid build:** the spawned sequencer is its **own session leader** (`SID == PID`) with **no controlling terminal** (`tty = ?`).
- **no-setsid (master):** the sequencer stays in the launcher's session (`SID != PID`).
- **Real survival test:** delivered `SIGHUP` to the sequencer's original launcher session group — the setsid sequencer **stayed alive and healthy, still serving RPC**. That is precisely the tmux-detach / SSH-disconnect survival property the PR claims.

### Also confirmed
- Blast radius is correct: the only two `spawn_to_log` callers are the localnet and test-node sequencer spawns — both should be daemonized. No `localnet stop` regression (stop kills the tracked PID directly).
- `pre_exec` closure is async-signal-safe (`setsid` + `last_os_error`) and propagates failure; `#[cfg(unix)]` gating is right.
- `cargo fmt --check` clean; `daemonized_process_is_own_session_leader` passes; 459 lib tests pass.
- DOGFOODING.md (D2) and README updated with the survival behavior.

Note: CI didn't auto-run on this fork PR (no check-runs on the head SHA) — worth a maintainer kicking off the workflow before merge, but I ran fmt + the test suite locally and they're green.

No blocking findings. (Comment only, not an approval.)

weboko (commented):
## Independent re-review — setsid daemonization verified ✅

Re-reviewed the mechanism from scratch on a fresh environment (Rust + risc0 `r0vm`/`cargo-risczero` 3.0.6 + Nix + Docker all provisioned). Agree with the prior review — this is a clean, correct, well-scoped fix.

### Verified
- **Builds clean, `cargo fmt --check` clean**, and `process::logged_tests::daemonized_process_is_own_session_leader` **passes** — the child spawned through the real `spawn_to_log` path has `SID == PID`, i.e. it is its own session leader.
- **Independent repro of the survival property**: replicated the `pre_exec(setsid)` path outside the test and confirmed the child lands in a *new session distinct from the launcher's* — so a controlling-terminal `SIGHUP` (shell/tmux/SSH hangup) cannot reach it. This is exactly the behavior the PR claims.
- **Blast radius is correct**: the only two `spawn_to_log` callers are the localnet sequencer (`src/commands/localnet.rs:384`) and the test-node sequencer (`src/testnode/mod.rs:219`). Both are long-lived sequencer spawns that *should* be daemonized; nothing short-lived is affected.
- **`setsid()` is async-signal-safe** and correctly propagates failure via `last_os_error()`; it cannot spuriously fail here because the forked child always has a PID distinct from its inherited PGID (so it is never already a group leader). `#[cfg(unix)]` gating is right (scaffold is Unix-only). No `localnet stop` regression — stop kills the tracked PID directly, not the process group.
- **Docs**: DOGFOODING.md (D2) and the README `localnet start` bullet accurately describe the new survival guarantee.

### Non-blocking, before merge
- **Rebase / regenerate `Cargo.lock`.** This branch's `Cargo.lock` is still lockfile `version = 3` while `master` is now `version = 4`, and the diff carries a large set of dependency *downgrades* relative to `master` (e.g. `anstream 1.0.0` → `0.6.21`). Merging as-is would revert those. A rebase on `master` + `cargo update -p libc`/regeneration will collapse this back to just the intended `+ libc` addition.

Mechanism, tests, and docs all check out; only the lockfile hygiene above is worth doing before merge.

ygd58:
Updated Cargo.lock to version 4 by rebasing on latest master. @weboko

weboko (commented):
## Re-review of the new commit (`858f179` — Cargo.lock → v4) — last blocker resolved ✅

The fresh commit since my last pass does exactly what I asked for before merge: it regenerates `Cargo.lock` to the upstream **lockfile `version = 4`** format, and this collapses the diff. The `setsid` mechanism, the test, and the docs are unchanged.

### Verified on a fresh environment (Rust 1.94 + Nix 2.35.1 + risc0 r0vm 3.0.6 provisioned)

- **Lockfile hygiene is now clean.** The entire `Cargo.lock` diff vs `master` is a **single added line** (`+ libc` under the `logos-scaffold` dependency list), with **zero removals** — the previous +195/-170 churn and the dependency *downgrades* (e.g. `anstream 0.6.21`) are gone. The whole PR is now **+41/-1 across 5 files**. This was the only thing I was holding on before.
- **Builds clean; `cargo fmt --check` clean.**
- **Unit test passes**: `process::logged_tests::daemonized_process_is_own_session_leader` — the child spawned through the real `spawn_to_log` path has `SID == PID` (own session leader).
- **Independent survival repro.** I reproduced the property outside the test: a child spawned via the same `pre_exec(setsid)` path becomes its own session leader in a session **distinct from the launcher's**; after delivering `SIGHUP` to the launcher's process group (the terminal/tmux/SSH-hangup signal), the `setsid` child **stayed ALIVE and healthy**, while an otherwise-identical child spawned *without* `setsid` (left in the launcher's session) was **terminated**. That is precisely the survival guarantee this PR claims.

### Code read (unchanged, still correct)
- `pre_exec` closure is async-signal-safe (`setsid()` + `last_os_error()`), correctly propagates failure, and `#[cfg(unix)]` gating is right (scaffold is Unix-only).
- Blast radius is correct: the only `spawn_to_log` callers are the localnet and test-node sequencer spawns — both are long-lived sequencers that *should* be daemonized.
- Docs are in sync: DOGFOODING.md D2 and the README `localnet start` bullet both describe the survival behavior.

### Note before merge
- CI still hasn't auto-run on this fork PR (0 check-runs on `858f179`, status `pending`) — worth a maintainer kicking off the workflow before merge. I ran `cargo build`, `cargo fmt --check`, and the relevant tests locally and they are green.

All blockers from my previous reviews are now resolved; from my side this is merge-ready pending a green CI run. (Comment only, not an approval.)

---

ygd58:
Thank you @weboko for the thorough re-review! All blockers resolved. @fryorcraken — could you kick off CI and merge when green? weboko has verified this is merge-ready. 🙏

weboko:
## Re-review — CI is green now, last blocker cleared; one new finding about the `test-node` call site

Picking this up because the thing I was holding on has resolved: **`validate` ran and passed on `858f179` at 2026-08-03T01:51:52Z** ([run](https://github.com/logos-co/scaffold/actions/runs/29984313947/job/91576648059)). That was the only outstanding item from my 2026-07-23 pass, so the lockfile/rebase/docs asks are all settled.

### Re-verified on a fresh environment

Provisioned from scratch (Rust 1.94.1, Nix 2.35.1 with flakes, r0vm 3.0.5 matched to the pinned LEZ `cf3639d8`'s `risc0-zkvm = 3.0.5`, risc0 guest toolchain 1.91.1, risc0 C++ 13.2.0, Docker 29.3.1).

- The branch **merges cleanly into current `master` (`9fcc376`)** — no conflicts, even though `master` has moved since your last push.
- On that merge result: `cargo fmt --check` clean, `cargo build --release` clean, and `cargo test --locked` **715 green** (525 lib + 182 CLI + 3 testnode + 5 doc), zero failures.
- `process::logged_tests::daemonized_process_is_own_session_leader` passes — the child spawned through the real `spawn_to_log` has `SID == PID`.

**D2 dogfooded end to end** against a real localnet on this branch (real `sequencer_service` built from the pinned LEZ `cf3639d8`, circuits v0.4.1, `setup` complete, wallet seeded) — this is the scenario the PR adds its signal to:

- `localnet start` → `localnet ready (sequencer pid=12823)`.
- The daemonization is observable on the real sequencer: `PID 12823, PGID 12823, SID 12823, PPID 1, STAT Ss` — its own session leader, reparented to init. It then kept producing blocks (through block id 27) across many separate shell invocations, which is the survival property in normal use rather than in a test harness.
- `localnet status` / `--json` report `tracked_pid`, `listener_present`, `ownership`, `ready` correctly in both the running and stopped states.
- `doctor` → **24 PASS, 0 WARN, 0 FAIL, status `Ready`, exit 0**; `doctor --json` carries `status`, `summary`, `checks`, `next_steps`.
- `localnet logs --tail 200` returns useful block-production lines.
- `localnet stop` → clean, and post-stop status correctly reports stopped. Worth noting the daemonized sequencer is still reaped properly by `stop`, and scaffold's liveness check correctly does *not* count the lingering zombie as running.

No problems observed anywhere in D2. The `setsid` mechanism itself I have now verified across several separate environments; I have no concerns about its correctness.

### New finding — daemonization is applied to *both* `spawn_to_log` callers, and one of them is short-lived

This is the one thing I'd like a response on before merge. It is not a defect in the mechanism; it is about its blast radius, and I missed it on earlier passes.

`spawn_to_log` has exactly two production callers:

- `src/commands/localnet.rs:372` — the long-lived developer sequencer. Surviving shell/tmux closure is the whole point of this PR, and there is a management surface (`localnet status` / `localnet stop`) to recover it. **Clearly correct.**
- `src/testnode/mod.rs:306` — test nodes, which the module's own doc comment (`src/testnode/mod.rs:1`) and the CLI help (`src/cli.rs:98`) both describe as *"isolated, **short-lived** sequencer test nodes"*.

For the second one, `setsid` removes the last-resort cleanup that used to exist. Test-node teardown relies on `TestNode::stop()` or `impl Drop for TestNode` (`src/testnode/mod.rs:431`) — and Rust `Drop` does not run on `SIGINT`/`SIGTERM`, since nothing in the crate installs a signal handler (I grepped: no `ctrlc`, no `signal`, no handler registration anywhere under `src/testnode/`). Before this PR, the spawned sequencer sat in the launcher's process group, so a process-group-directed signal reaped it even when `Drop` never ran. After this PR it is in its own session and does not receive that signal.

I verified this against a **real sequencer**, not a model of one — building both `master` and this branch, running `lgs test-node run -- sleep 90` under each, and sending `SIGINT` to the **launcher's process group** (what `Ctrl-C` on a foreground job, a terminal close, or a CI cancel actually does) rather than to the sequencer directly:

```
########## MASTER (control) ##########
[MASTER] lgs_pid=32632 lgs_pgid=32632 | seq_pid=32634 seq_pgid=32632 seq_sid=32632
[MASTER]   (sequencer shares the launcher's process group)
[MASTER]  RESULT: sequencer DEAD after Ctrl-C   ==> cleaned up

########## PR34 ##########
[PR34]   lgs_pid=338   lgs_pgid=338   | seq_pid=340   seq_pgid=340   seq_sid=340
[PR34]     (sequencer is in its OWN session/process group)
[PR34]    RESULT: sequencer ALIVE after Ctrl-C  ==> LEAKED
```

So: **`Ctrl-C` during `lgs test-node run -- <cmd>` leaks a real sequencer on this branch, and does not on `master`.** The `SIGINT` used to reach both the `lgs` process and the sequencer; now it reaches only `lgs`, `Drop` does not run, and the sequencer keeps running holding its port and its rocksdb state dir under `.scaffold/test-nodes/<id>`. Same for a cancelled or timed-out CI job.

(Two notes for anyone reproducing this. A `kill -0` liveness check is misleading — the reaped child lingers as an unreaped zombie in a container whose PID 1 does not reap, so `kill -0` reports "alive"; the check has to read process *state*. And the state dir is left behind in **both** arms, since `Drop` runs in neither — the difference this PR makes is specifically whether the *process* is reaped.)

Severity is bounded, and I don't want to overstate it: ports are auto-picked (`--port 0`), so a leak does not block subsequent nodes; run slots are reclaimed by owner-pid liveness (`src/testnode/mod.rs:844`), so concurrency caps recover on their own; and on ephemeral CI runners the leak dies with the VM. The real cost lands on developer machines, where repeated `Ctrl-C` accumulates orphaned sequencers and state dirs — and there is no `test-node list` or `test-node prune` to find them, only `stop --node <id>`, which needs an id the user has to dig out of `.scaffold/test-nodes/`.

Suggested fix — make daemonization opt-in per call site rather than a property of `spawn_to_log`, so the PR's goal is preserved exactly and the test-node semantics are untouched:

```rust
pub(crate) fn spawn_to_log(cmd: &mut Command, log_path: &Path) -> DynResult<u32> {
    spawn_to_log_inner(cmd, log_path, false)
}

/// Like `spawn_to_log`, but detaches the child into its own session so it
/// survives shell/tmux closure. For long-lived processes the user manages
/// explicitly (localnet) — not for short-lived ones whose cleanup depends on
/// the parent's lifetime.
pub(crate) fn spawn_to_log_daemonized(cmd: &mut Command, log_path: &Path) -> DynResult<u32> {
    spawn_to_log_inner(cmd, log_path, true)
}
```

with `localnet.rs:372` calling the daemonized variant and `testnode/mod.rs:306` keeping the current behavior.

If you'd rather keep it uniform, that's a defensible call — test nodes are recoverable and the leak is not silent corruption — but then I think it should be written down, because "short-lived" in the module doc and CLI help now sets an expectation the code no longer meets. A sentence in the `spawn_to_log` doc comment and an adjustment to the `test-node` help text would cover it.

- `libc` is added to plain `[dependencies]`, but it is only used under `#[cfg(unix)]` (both in `spawn_to_log` and in the test). `[target.'cfg(unix)'.dependencies]` would express that more precisely. Scaffold is Unix-only in practice, so this is cosmetic.
- The `setsid()` failure path is unreachable in practice — the child cannot be a process-group leader immediately after `fork`, since it inherits the parent's PGID and its own PID differs from it. Keeping the check is still right; just noting the `Err` arm is defensive rather than live.
- The test leaves a `sleep 30` behind if it fails before reaching the `SIGTERM` line. Harmless, but a guard would make a failing run quieter.
- Docs are in sync: README `localnet start` bullet and the DOGFOODING D2 signal both describe the survival behavior, and the D2 addition is proportionate to the change. Nothing further needed there for the localnet half; the test-node half is what §2 above is about.

### Where this leaves the PR

The mechanism is correct, tested, documented, and now CI-green, and it merges cleanly onto current `master`. My previous blockers are all resolved. The one open item is the `test-node` call site above — either scope the daemonization to `localnet`, or state explicitly that test nodes are daemonized too and adjust the "short-lived" wording. Either resolution is fine by me.

---

weboko:
## Re-verified on the new base (`7523ee2`) — everything still green, and the `test-node` finding still reproduces

No new commits since `858f179`, so this is not a re-review of the diff. `master` has moved `9fcc376` → `7523ee2` since my 2026-08-03 pass, so I rebuilt and re-ran everything against the new base rather than letting the previous verification go stale.

### Still green on the new base

Provisioned from scratch (Rust 1.94.1, Nix 2.35.2, r0vm **3.0.5** matched to the pinned LEZ `cf3639d8`'s `risc0-zkvm = 3.0.5`, risc0 guest toolchain 1.91.1, risc0 C++ 13.2.0, Docker 29.3.1).

- **Merges cleanly into `7523ee2`** — still the only open PR that does.
- On the merge result: `cargo fmt --check` clean, `cargo build --release` clean, `cargo test --locked` **743 green** (553 lib + 182 CLI + 3 testnode + 5 doc), zero failures. (Up from 715 last pass — all growth is from `master`, none from this PR.)
- **Blast radius unchanged.** `spawn_to_log` still has exactly two production callers on the new base, at the same lines — `src/commands/localnet.rs:372` and `src/testnode/mod.rs:306`. No new caller has appeared that would silently inherit daemonization.

### D2 dogfooded end to end against a real localnet on this branch

Real `sequencer_service` built from the pinned LEZ `cf3639d8`, circuits v0.4.1, `setup` complete and wallet seeded.

- `localnet start` → `localnet ready (sequencer pid=11596)`.
- Daemonization observable on the real sequencer: `PID 11596, PGID 11596, SID 11596, PPID 1, STAT Ssl` — its own session leader, no controlling terminal, reparented to init.
- `localnet status` / `--json` report `tracked_pid`, `listener_present`, `ownership`, `ready` correctly in both running and stopped states.
- `doctor` → **24 PASS, 0 WARN, 0 FAIL, status `Ready`, exit 0**; `doctor --json` carries `status`, `summary`, `checks`, `next_steps`.
- `localnet logs --tail 200` returns real block-production lines (r0vm executing, blocks committing).
- `localnet stop` → clean; post-stop status correctly reports stopped.

No problems observed anywhere in D2.

### The open item from 2026-08-03 still reproduces — unchanged

I re-ran the A/B on the new base rather than assuming it still held. `lgs test-node run -- sleep 120` under each build, with `SIGINT` delivered to the **launcher's process group** (what `Ctrl-C` on a foreground job, a terminal close, or a CI cancel actually does):

```
########## MASTER (control) ##########
[MASTER] lgs_pid=12333 lgs_pgid=12333 | seq_pid=12335 seq_pgid=12333 seq_sid=12333
[MASTER]  RESULT: sequencer DEAD (state='Z')     ==> cleaned up

########## PR34 ##########
[PR34]   lgs_pid=12475 lgs_pgid=12475 | seq_pid=12477 seq_pgid=12477 seq_sid=12477
[PR34]   RESULT: sequencer ALIVE (state='Ssl')   ==> LEAKED
```

Same conclusion as before: on `master` the sequencer shares the launcher's process group and the group-directed `SIGINT` reaps it; on this branch it is in its own session, `SIGINT` reaches only `lgs`, `Drop` does not run (nothing in the crate installs a signal handler), and a real sequencer is left running holding its port and its rocksdb state dir under `.scaffold/test-nodes/<id>`.

(Two notes for anyone reproducing. Read process *state*, not `kill -0` — the reaped child lingers as an unreaped zombie in a container whose PID 1 does not reap, so `kill -0` reports "alive". And `pgrep -x sequencer_service` silently matches nothing: the name is 17 chars and `comm` truncates at 15, so match on the path instead.)

### Where this leaves the PR

Nothing has regressed and nothing new has appeared — mechanism, tests, docs, and D2 are all still correct on the new base, and this remains the closest of the open PRs to merge-ready. The single open item is the one from 2026-08-03, which hasn't been answered yet: either scope daemonization to the `localnet` call site (`spawn_to_log` / `spawn_to_log_daemonized`, sketched in that comment), or keep it uniform and say so explicitly — the module doc at `src/testnode/mod.rs:1` and the CLI help at `src/cli.rs:98` both still describe test nodes as *"short-lived"*, which the code no longer matches. Either resolution is fine by me; it just needs a decision.

@ygd58 — no action needed on the verification side, just that one call.

---

weboko:
## New finding — the daemonization is incomplete: `stdin` is never redirected, so the child pins the terminal device

No new commits since `858f179` and `master` is still `7523ee2`, so this is not a re-review of the diff — my 2026-08-17 verification stands and I'm not repeating it. I rebuilt anyway rather than trusting the earlier run, and while inspecting the *real* sequencer's file descriptors (something the previous passes checked process state for, but not fds) I found something new.

### The finding

`spawn_to_log` redirects `stdout` and `stderr` to the log file, but leaves `stdin` inherited:

```rust
// src/process.rs:606-608
let file = File::create(log_path)?;
let err_file = file.try_clone()?;
cmd.stdout(Stdio::from(file)).stderr(Stdio::from(err_file));
//  ^^ no .stdin(...)
```

`setsid()` drops the *controlling terminal*, but it does not touch inherited descriptors. So the daemonized child keeps an open fd on whatever `stdin` was — in an interactive shell, the tty.

Verified on the real `sequencer_service`, started through `lgs localnet start` inside a genuine pty which was then destroyed:

```
$ script -qec "lgs localnet start" /dev/null      # pty created, then torn down

  pid=7247  ppid=1  sid=7247  stat=Ssl  tty=?     <- setsid worked: own session, no ctty
  fd0 -> /dev/pts/0 (deleted)                     <- still holding the destroyed terminal
  fd1 -> .../.scaffold/logs/sequencer.log
  fd2 -> .../.scaffold/logs/sequencer.log
```

The `(deleted)` is the point: the terminal is gone, and the daemon is still pinning it. Textbook daemonization detaches all three standard descriptors; this does two of three.

A caveat so this isn't overweighted, since I initially mis-diagnosed it myself: **the PR's actual promise still holds.** `setsid()` is what stops `SIGHUP` from reaching the process, and I confirmed the sequencer genuinely outlives its terminal — the documented check passes, see D2 below. This is resource hygiene and correctness-of-detachment, not a break of the feature. It also only shows up interactively: my first probe ran under a non-interactive harness whose `stdin` was already `/dev/null`, and the child dutifully inherited `/dev/null`, which looked fine. It takes a real pty to see it.

Why it's still worth fixing:

- Every `localnet start` from a terminal leaves the sequencer holding a reference to that terminal for its entire lifetime, so the device can't be reclaimed. On a shared or long-lived box that accumulates.
- If any `spawn_to_log` child ever reads `stdin`, it reads from a terminal it does not control — `EIO`, not a clean `EOF`.
- It applies to **both** call sites (`localnet.rs:372` and `testnode/mod.rs:306`); neither sets `stdin`.

One line, at the same place as the existing redirects:

```suggestion
    let file = File::create(log_path)?;
    let err_file = file.try_clone()?;
    cmd.stdin(Stdio::null())
        .stdout(Stdio::from(file))
        .stderr(Stdio::from(err_file));
```

Neither the sequencer nor a test node reads `stdin`, so this is behaviour-preserving for both callers.

### Re-verified on the merge result

Provisioned from scratch (Rust 1.94.1, Nix 2.35.2, risc0 — r0vm **3.0.5** matched to the pinned LEZ `cf3639d8`'s `risc0-zkvm`, cargo-risczero 3.0.6, guest toolchain 1.97.0, C++ 2024.1.5).

- Still **merges cleanly into `7523ee2`** — still the only open PR that does.
- On the merge result (`7a98ff2`): `cargo test --locked --release` **743 green** (553 lib + 182 CLI + 3 testnode + 5 doc), zero failures.
- `process::logged_tests::daemonized_process_is_own_session_leader` passes against the real `spawn_to_log`.
- Blast radius unchanged: `spawn_to_log` still has exactly two production callers, `localnet.rs:372` and `testnode/mod.rs:306`.
- The `pre_exec` closure is correct on the subtle point that matters — post-`fork` it calls only `libc::setsid()` and `Error::last_os_error()` (an `errno` read), both async-signal-safe, with no allocation. That's the usual way `pre_exec` goes wrong and this PR gets it right.

### D2 dogfooded end to end — no problems observed

Real `sequencer_service` from the pinned LEZ, circuits v0.4.1, wallet seeded.

- `localnet start` → `localnet ready (sequencer pid=6978)`; with r0vm 3.0.5 installed the `RISC0_SERVER_PATH` warning correctly disappears (`find_r0vm_path_for_lez` resolves the LEZ-pinned version).
- Daemonization on the real process: `PID 7247, PPID 1, SID 7247, STAT Ssl, tty ?`.
- **The DOGFOODING.md line this PR adds was executed as written**: started localnet in a pty, destroyed the pty, then checked from a genuinely new shell (`env -i HOME=… /bin/sh`) → `tracked sequencer: pid=7247 running=true`, `ready: true`. The documented claim is accurate.
- `localnet status` / `--json` correct in running and stopped states.
- `doctor` → **24 PASS, 0 WARN, 0 FAIL, status `Ready`, exit 0**.
- `localnet logs` returns real sequencer output; `localnet stop` clean, post-stop status correct.

Docs are in good shape: the README `localnet start` bullet and the new `DOGFOODING.md` D2 assertion both describe the behaviour accurately, which is more than most PRs of this size manage.

### The `test-node` item from 2026-08-03 still reproduces

Re-ran the A/B independently on the new base rather than assuming it held — `SIGINT` to the launcher's process group:

```
[MASTER] lgs_pid=7583 lgs_pgid=7583 | seq_pid=7592 seq_pgid=7583 seq_sid=7583
[MASTER]  RESULT: sequencer DEAD (state='Z')     ==> cleaned up

[PR34]   lgs_pid=7636 lgs_pgid=7636 | seq_pid=7644 seq_pgid=7644 seq_sid=7644
[PR34]   RESULT: sequencer ALIVE (state='Ssl')   ==> LEAKED
```

Unchanged, and still the one open decision: scope daemonization to the `localnet` call site (`spawn_to_log` / `spawn_to_log_daemonized`), or keep it uniform and update the "short-lived" wording at `src/testnode/mod.rs:1` and `src/cli.rs:98`. Either is fine; it just needs a call.

### Where this leaves the PR

Mechanism, tests, docs, and D2 are all correct, and this remains by a wide margin the closest of the open PRs to merge-ready. Two open items now, both small:

1. `cmd.stdin(Stdio::null())` — the one-line fix above.
2. The `test-node` scoping decision from 2026-08-03.

@ygd58 — (1) is new and mechanical; (2) is the same question as before.

---

weboko:
## Re-verified on the new base (`834b6b0`) — `#251` landed on the exact path this PR changes, and they compose correctly. The `stdin` finding still reproduces.

No new commits since `858f179`, so this is not a re-review of the diff. I picked it back up because `master` moved `7523ee2` → `834b6b0`, and `#251` ("fix sequencer") rewrote `LocalnetError::ExitedBeforeReady` / `StartTimeout` — i.e. the failure path of the very function this PR modifies. That's the one base change that could plausibly interact with `setsid`, so I re-provisioned from scratch and re-ran rather than assuming the 2026-08-17 verification carried over.

Environment: Rust 1.94.1, Nix 2.34.8, risc0 r0vm **3.0.5** (matched to the pinned LEZ `cf3639d8`'s `risc0-zkvm`), guest toolchain 1.91.1, C++ 2024.1.5, real `sequencer_service` + `wallet` built from the pin, circuits v0.4.1.

### The `#251` interaction is clean — verified, not assumed

`src/error.rs` is **not** among the merge's touched files (this branch never touches it), so `#251`'s new `Next: run logos-scaffold localnet status …` guidance survives the merge intact. More usefully, `#251` also added a CLI test that asserts that guidance, and it exercises the spawn path `setsid` now sits on:

```
$ cargo test --locked --release localnet_start_fails_when_process_exits_before_ready
test localnet_start_fails_when_process_exits_before_ready ... ok
```

And the real-world version of that path still behaves, with a genuine foreign listener on 3040:

```
$ lgs localnet start
error: cannot start localnet: port 3040 is already in use (pid=23055)
This may be a sequencer started from another project and may not work with the current project.
Stop that process and retry `logos-scaffold localnet start`.
Try: kill 23055                                                     (exit 1)

$ lgs localnet status
listener http://127.0.0.1:3040: reachable (pid=23055)
ownership: foreign
next steps:
- Stop the external listener on 127.0.0.1:3040 or choose a clean environment
```

Worth stating explicitly because `setsid` makes it *more* likely, not less: a daemonized sequencer that outlives its terminal is exactly the thing that later shows up as a stale/foreign listener, and `#251`'s premise is that such a failure must carry a next step. It does. The two changes reinforce each other.

### Merge and test status on `834b6b0`

- **Still the only open PR that merges cleanly** (`#226`, `#210`, `#199` are all `mergeable_state=dirty`). Merge commit `b3f8348`, no conflicts, `Cargo.lock` merged fine against `#251`'s version bump to `0.3.0`.
- `cargo build --release` clean; `cargo test --locked --release` → **743 green** (553 lib + 182 CLI + 3 testnode + 5 doc), zero failures.
- `process::logged_tests::daemonized_process_is_own_session_leader` passes against the real `spawn_to_log`.
- Blast radius unchanged: `spawn_to_log` still has exactly two production callers, `localnet.rs:372` and `testnode/mod.rs:306`.

### D2 dogfooded end to end on the merge result — no problems observed

Real `sequencer_service`, real wallet, per `DOGFOODING.md` D2:

- `localnet start` → `localnet ready (sequencer pid=21775)`; no `RISC0_SERVER_PATH` warning (`find_r0vm_path_for_lez` resolves the LEZ-pinned 3.0.5).
- `localnet status` and `--json` correct in running, foreign, and stopped states; JSON carries `tracked_pid`, `listener_present`, `ownership`, `ready`.
- `doctor` → **24 PASS, 0 WARN, 0 FAIL, status `Ready`, exit 0**; `doctor --json` has `status`/`summary`/`checks`/`next_steps`.
- `localnet logs --tail 200` returns real sequencer output.
- **The `DOGFOODING.md` line this PR adds was executed as written.** Started localnet inside a genuine pty, destroyed the pty, then checked from a truly clean shell (`env -i HOME=… /bin/sh`):

```
  PID  PPID   SID  PGID STAT TTY
22232     1 22232 22232 Ssl  ?          <- own session, no controlling terminal

$ env -i … lgs localnet status
tracked sequencer: pid=22232 running=true
ownership: managed
ready: true
```

The documented claim is accurate. One thing I checked specifically because `setsid` could have broken it and nobody had: **`localnet stop` still works against a session leader.** `stop_localnet` kills by pid, not process group, so detaching the session doesn't orphan the stop path — `kill 22232 # sequencer` → `localnet stopped`, post-stop status correct.

### Still open — the `stdin` finding reproduces unchanged on the new base

From my 2026-08-17 comment; re-confirmed on `834b6b0` against the real sequencer in a pty that was then destroyed:

```
fd0 -> /dev/pts/0 (deleted)          <- still pinning the dead terminal
fd1 -> …/.scaffold/logs/sequencer.log
fd2 -> …/.scaffold/logs/sequencer.log
```

`setsid()` drops the controlling terminal but not inherited descriptors, so two of three are detached. It only shows up interactively — under a non-interactive harness `stdin` is already `/dev/null` and the child inherits that, which looks fine (I re-confirmed both shapes today). Still a one-liner, at the same place as the existing redirects:

```suggestion
    let file = File::create(log_path)?;
    let err_file = file.try_clone()?;
    cmd.stdin(Stdio::null())
        .stdout(Stdio::from(file))
        .stderr(Stdio::from(err_file));
```

Neither the sequencer nor a test node reads `stdin`, so this is behaviour-preserving for both callers. The `test-node` process-group item from 2026-08-03 also still stands.

### Two small new nits, both in `Cargo.toml`

```toml
[dependencies]
libc = "0.2"        # <- inserted above `anyhow` in an otherwise alphabetical list
anyhow = "1.0"
base64 = "0.22"
```

1. Sort order — every other entry is alphabetical; `libc` belongs between `indicatif` and `notify`.
2. `libc` is used only under `#[cfg(unix)]`, but is declared unconditionally. Now that the crate is actually published (`0.3.0` went to crates.io on 2026-08-18), `[target.'cfg(unix)'.dependencies] libc = "0.2"` would keep it off non-Unix resolution graphs. MSRV is fine either way — `libc 0.2.182` declares `rust-version = "1.63"`, under this crate's `1.81`.

Neither is a blocker.

### Where this leaves it

The feature works, is documented accurately in both `README.md` and `DOGFOODING.md`, is covered by a real test, composes correctly with `#251`, and is the only open PR that merges cleanly. The `pre_exec` closure is still correct on the subtle point that matters — post-`fork` it calls only `libc::setsid()` and `Error::last_os_error()` (an `errno` read), both async-signal-safe, no allocation.

The single-line `stdin(Stdio::null())` is what I'd want before merge; the two `Cargo.toml` nits are optional. @ygd58 — that's a one-line change plus a dependency move, and I don't have anything else outstanding on this one.

---

weboko:
## Addendum on the open `stdin` finding — it is not purely cosmetic: the daemon pins the controlling pty open

No new commits since `858f179`, so this is not a re-review — everything in my 2026-08-19 comment stands, and I re-confirmed the D2 run end to end on a from-scratch box today (Rust 1.94.1, Nix 2.35.2, r0vm 3.0.5 matched to the pinned LEZ `cf3639d8`, guest toolchain 1.91.1, C++ 2024.1.5, real `sequencer_service` + `wallet`, circuits v0.4.1; merge onto `834b6b0` clean, `cargo test --locked --release` **743 green**, `doctor` **24 PASS / 0 WARN / 0 FAIL, status Ready**, `process::logged_tests::daemonized_process_is_own_session_leader` ok).

I'm posting only because I set out to reproduce the `stdin` finding properly and found it has a **functional** consequence I previously described as interactive-only cosmetics. I want to correct my own characterisation.

### What I measured

Started `localnet start` inside a real `pty.fork()`, let the shell exit, then asked which processes still hold the pty slave:

```
start output:      localnet ready (sequencer pid=24195)
shell exited:      True
sequencer pids:    ['24195']
sequencer fd0 ->   /dev/pts/0
sequencer fd1 ->   .../.scaffold/logs/sequencer.log
pty-slave holders: [('24195', '0', '/dev/pts/0')]     <- the daemon is the ONLY holder

PHASE 1: probing master for EOF (15s budget)
RESULT: BLOCKED - no EOF and no data (pty pinned open by the daemon)
```

`setsid()` does its job — `SID == PID`, `TT` is `?`, PPID 1 — but it drops the *controlling terminal*, not the inherited descriptor. `fd0` still references the pty slave, so the master never reaches EOF. The parent has exited; nothing else holds it; the sequencer alone keeps it alive for its entire lifetime.

### Scope — narrower than "pty wrappers break", so stating it precisely

I checked the obvious real-world wrapper before claiming impact, and it is **not** affected:

```
$ timeout 45 script -q -e -c "lgs localnet start" /dev/null
localnet ready (sequencer pid=24639)
script rc=0  elapsed<1s
```

`script(1)` waits for the child to *exit*, not for master EOF, so it returns fine. The breakage is confined to harnesses that **read the pty master until EOF** — the standard `pty.fork()` + read-until-EOF idiom above, and `expect`'s `expect { eof }`. Those hang indefinitely rather than returning after `localnet ready`. That is a narrow but real class, and it fails in the least debuggable way possible: the command *succeeded*, the sequencer is healthy, and the harness just never returns.

So: not a blocker, and not something a normal interactive user or CI runner will hit — but no longer "cosmetic fd hygiene" either, which is how I filed it on 2026-08-17. It is a real hang with a one-line fix, and I'd rather it be fixed than carried.

### Unchanged fix, now with a second reason

Same one-liner, at the same place as the existing redirects (`src/process.rs`, `spawn_to_log`):

```suggestion
    let file = File::create(log_path)?;
    let err_file = file.try_clone()?;
    cmd.stdin(Stdio::null())
        .stdout(Stdio::from(file))
        .stderr(Stdio::from(err_file));
```

This also closes the finding for the other production caller — `spawn_to_log` has exactly two, `commands/localnet.rs:372` and `testnode/mod.rs:306`, and a test node spawned from a pty-driven integration harness is precisely the read-until-EOF shape above. Neither the sequencer nor a test node reads `stdin`, so it stays behaviour-preserving for both.

Everything else from my earlier comment is unchanged: the daemonization is correct and verified, the PR still merges cleanly (the only open PR that does), and the two `Cargo.toml` nits are still non-blocking.

---

ygd58:
Added `cmd.stdin(Stdio::null())` to `spawn_to_log` to close the inherited pty slave fd. This prevents pty-driven harnesses from hanging indefinitely. 519 lib tests pass. @weboko

weboko:
@ygd58 thanks for addressing! I am doing testing on the latest release, once I am done I will check this PR and merge

weboko:
## Re-verified on `6d4cca2` — daemonization + pty fix both confirmed end-to-end ✅

Ran the runbook on a fresh box against `6d4cca2` (base `834b6b0`), with the real standalone sequencer built via `test-node prepare`. Both mechanisms this PR ships behave exactly as the diff promises; no regression against the surrounding surface.

**Environment.** Ubuntu 24.04 x86_64, rustc 1.94.1. `test-node prepare` compiled `sequencer_service` locally; r0vm 3.0.5 installed under `~/.risc0/extensions/` per DOGFOODING.md § "Provisioning the Real LEZ Sequencer Toolchain".

**Session-leader daemonization (`setsid` via `pre_exec`).** `test-node start --port 3050 --json` → the spawned `sequencer_service` is PID 2712 and:

```
PID  PGID  SID   TT   COMMAND
2712 2712  2712  ?    sequencer_service
```

`Ppid: 1` (already reaped), `TT: ?` (no controlling terminal), and `SID == PID` — every property `setsid()` is supposed to establish. RPC on `127.0.0.1:3050` responds, so the daemon is genuinely serving, not just alive.

**Stdin-null (`6d4cca2`) fixes the pty pin.** Ran the CLI under `pty.fork()` and read the master until EOF:

- Master reaches EOF within milliseconds of the CLI printing its JSON and exiting.
- The sequencer child survives the pty tear-down (verified with `kill -0` on its PID afterwards).

`/proc/<pid>/fd/0 -> /dev/null` on the spawned sequencer confirms the redirect. Pre-fix code left `fd 0 -> /dev/pts/<n>`, which kept the master open forever; that's gone.

**Ancillary checks.**

- `cargo fmt --check` clean.
- `cargo test --lib process::` — 9 passed, 0 failed, including the new `daemonized_process_is_own_session_leader` regression.
- `spawn_to_log` callers audited: `src/commands/localnet.rs:372` and `src/testnode/mod.rs:306`. Both propagate errors from spawn (setsid failing inside `pre_exec` is surfaced through `Command::spawn`), so a botched detach fails loud instead of a zombie.
- README + DOGFOODING D2 additions match the observed behaviour and the failure signal they specify (`localnet status` from a new shell reporting `running=true`) is exactly the one this PR guarantees.

**Non-blocker note (CI, not code).** The head commit's Actions run shows `action_required` — the fork-PR approval gate, not a code problem. Once someone with rights approves the workflow run for `feat/localnet-daemon-mode-v2`, `validate` should go green (I ran the same three steps locally).

No new findings from this pass. From my side the last open concern is resolved.

---

weboko (approved):

PR #35, opened 2026-03-27
feat: add --prebuilt flag to skip compile-from-source on setup

ygd58:
## Summary

Refs #33.

Adds `--prebuilt` flag to `logos-scaffold setup` that downloads
pre-compiled binaries instead of building from source (~15 min penalty).

## Usage

```bash
logos-scaffold setup --prebuilt
```

## How it works

1. Computes tag: `lssa-prebuilt-<commit8>-<arch>-<os>`
2. Tries to download `sequencer_runner` from GitHub releases
3. Falls back to `cargo build` automatically if no prebuilt exists

## Tag format

```
lssa-prebuilt-767b5afd-x86_64-linux
lssa-prebuilt-767b5afd-aarch64-macos
```

## Notes

- Prebuilt binaries need to be published to GitHub releases (separate CI step)
- Graceful fallback ensures setup never fails due to missing prebuilt
- `wallet` binary download can be added as a follow-up

Refs #33

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -81,6 +81,11 @@ struct NewArgs {
 struct SetupArgs {
     #[arg(long, value_enum, default_value_t = WalletInstallMode::Auto)]
     wallet_install: WalletInstallMode,
+
+    /// Download prebuilt binaries instead of compiling from source.
+    /// Falls back to source build if no prebuilt exists for the pinned commit.
+    #[arg(long, default_value_t = false)]
+    prebuilt: bool,
 }
 
 #[derive(Debug, clap::Args)]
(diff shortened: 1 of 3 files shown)


vpavlin:
I think a pre-requisite for this is to actually build the sequencer in CI and upload to artifacts

ygd58:
Thanks @vpavlin! Agreed — the prebuilt download is only useful once CI builds and uploads the artifacts.

As a follow-up, a GitHub Actions workflow could:
1. Build sequencer_runner on each lssa pin update
2. Upload as a release asset with tag format: lssa-prebuilt-<commit8>-<arch>-<os>

For now, the --prebuilt flag gracefully falls back to source build when no artifact exists, so it is safe to merge without breaking anything. The CI workflow can be added separately.

Would you like me to add the CI workflow as part of this PR or as a separate follow-up?

ygd58:
Closing to rebase on upstream. The upstream refactored setup.rs significantly (lssa→lez rename, removed SetupCommand). Will reopen with a clean implementation on top of current master.

PR #46, opened 2026-04-08
docs: add FURPS and ADR from ecosystem wiki

fryorcraken:
## Summary

- Creates dedicated `FURPS.md` and `ADR.md` files with content from the ecosystem wiki
- Adds a Documentation section to the README linking to both files

Moving requirements docs into the repos that own the code so they stay in sync with development, rather than living in a separate wiki repo.

## Test plan

- [ ] Verify FURPS.md and ADR.md render correctly on GitHub
- [ ] Verify README Documentation links work

### DIFF
--- a/FURPS.md
+++ b/FURPS.md
@@ -1,0 +1,60 @@
+# Scaffold — FURPS+
+
+## FURPS+ (v0.1)
+
+[v0.1 milestone](https://github.com/logos-co/ecosystem/milestone/9)
+
+### Functionality
+
+1. One public DevNet vertical slice: generate wallet, fund wallet, deploy contract, execute one transaction type, verify result.
+2. Integrate wallet generation as part of the scaffold workflow for bootstrap and interaction flows.
+3. Support native token topup for wallet operations on local and DevNet environments.
+
+### Usability
+
+1. Single command bootstrap with no manual project wiring required.
+2. Generated layout clearly separates contract code, client code, config, and deploy scripts.
+3. Deterministic wallet generation and .env handling for repeatability.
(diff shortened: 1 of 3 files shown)


danisharora099 (approved):

FURPS.md
```diff
@@ -0,0 +1,60 @@
+# Scaffold — FURPS+
```
weboko, line 1: shall we keep them here as a copy of what we have in ecosystem repo or source of truth?

weboko (approved):
lgtm, just want to figure why we copy it here

FURPS.md
weboko: ok, so should it be the case for all other software we develop? 
I am personally inclined to have FURPS in one place and keep the link to it in `logos-scaffold` repo 

> Moving requirements docs into the repos that own the code so they stay in sync with development, rather than living in a separate wiki repo.

fryorcraken: I would say we **move** it here. And we then remove from ecosystem repo. The source of truth is near the code.

weboko: how would you link software ecosystem engineering creates/needs?

fryorcraken: Not sure I get your concern but keen to discuss.

fryorcraken:
> lgtm, just want to figure why we copy it here
 
The idea is to move closer to the code, to ensure it is up to date. It will not be when it's far away in the ecosystem repo.

We can still commit ourselves to specific deliveries, but we also do non-planned changes. All those features we are doing unplanned are getting lost as we forget to update the ecosystem repo.

PR #52, opened 2026-04-11
fix: detect IDL name collisions after sanitization to prevent silent overwrites

crazywriter1:
## Summary

`sanitize_file_stem()` normalizes IDL block names by lowercasing and replacing
non-alphanumeric characters with `_`. This means different names can map to the
same output filename (e.g. `foo-bar` and `foo_bar` both become `foo_bar.json`),
and the later block silently overwrites the earlier one.

This adds a `HashMap` to track seen stems and bails with a clear error message
when a collision is detected.

## Test plan

- [x] `cargo check` — compiles cleanly
- [x] `cargo fmt --check` — no formatting issues
- [x] `cargo test` — all 89 existing tests pass

### DIFF
--- a/src/commands/idl.rs
+++ b/src/commands/idl.rs
@@ -60,9 +60,16 @@ pub(crate) fn build_idl_for_current_project() -> DynResult<()> {
     }
 
     blocks.sort_by(|a, b| a.0.cmp(&b.0));
+    let mut seen_stems: std::collections::HashMap<String, String> =
+        std::collections::HashMap::new();
     for (name, json_text) in blocks {
         let canonical = canonical_json(&json_text)?;
-        let file_name = format!("{}.json", sanitize_file_stem(&name));
+        let stem = sanitize_file_stem(&name);
+        if let Some(prev_name) = seen_stems.get(&stem) {
+            bail!("IDL name collision: `{name}` and `{prev_name}` both sanitize to `{stem}.json`");
+        }
+        seen_stems.insert(stem.clone(), name.clone());
+        let file_name = format!("{stem}.json");
         let path = idl_dir.join(file_name);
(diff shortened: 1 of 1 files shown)


fryorcraken:
Thanks for the patch, @crazywriter1, and for keeping the diff small.

I'd like to hold off on detailed review here, because when I built this branch and ran the standard validation flow — `logos-scaffold new --template lez-framework` → `setup` → `build` — the new check fires on the freshly generated project and aborts the build with `IDL name collision: lez_counter and lez_counter both sanitize to lez_counter.json`. In other words, on the default template every new user would hit a hard error at `lgs build`. That's the kind of thing `D2` in [DOGFOODING.md](https://github.com/logos-co/logos-scaffold/blob/master/DOGFOODING.md) exists to catch, and the test plan in the PR description doesn't mention it being rerun.

Rather than work through specific fixes line-by-line, it'd be more useful if you could re-engage with a few principles from [CONTRIBUTING.md](https://github.com/logos-co/logos-scaffold/blob/master/CONTRIBUTING.md) before pushing again:

- **Dogfood end-to-end before opening the PR.** `cargo check / fmt / test` isn't enough for a change to `build idl` — `D1`/`D2` would have surfaced the regression above.
- **Anchor the change in a real user problem.** A linked issue, a discussion, or a concrete repro from a project you're building helps a reviewer understand whether the bug is reachable today and whether this is the right shape of fix.
- **Verify the new behaviour, not just the old.** A change that adds an error path warrants a test that exercises it — including the negative case where the same name appears legitimately.

Happy to take another pass once those have been folded in. The repo's 3-PRs-per-7-days cap will apply if you re-open this as a fresh PR, so it's worth getting it right on the next push.

PR #53, opened 2026-04-11
fix: localnet start with relative LSSA paths and config-based port

fryorcraken:
## Summary

- Patch `sequencer_config.json` port instead of passing `--port` CLI flag (unsupported by pinned LSSA)
- Fix ENOENT when spawning sequencer with relative `scaffold.toml` paths
- Extract `SEQUENCER_BIN_REL_PATH` and `SEQUENCER_CONFIG_DIR_REL_PATH` to constants

## Details

`logos-scaffold localnet start` had two bugs:

### 1. `--port` CLI flag not supported by pinned LSSA

The `sequencer_runner` at the default LSSA pin (`767b5afd`) doesn't accept `--port` as a CLI argument — it reads the port from `sequencer_config.json`. Instead of passing `--port`, we now patch the JSON config file before starting.

### 2. Binary path resolved from wrong directory

`Command::new(relative_path).current_dir(lssa)` resolves the binary path after `chdir(lssa)` in the child process. With a relative `[repos.lssa].path` like `.scaffold/cache/repos/lssa`, the binary path becomes:

```
.scaffold/cache/repos/lssa/.scaffold/cache/repos/lssa/target/release/sequencer_runner
                           ^--- resolved from child cwd, not parent
```

This fails with `No such file or directory (os error 2)`. Fixed by using `./target/release/sequencer_runner` which is relative to the child's cwd (`lssa`).

## Test plan

- [x] `cargo check` passes
- [x] `cargo fmt --check` passes
- [x] All 53 integration tests pass
- [ ] Manual validation: `logos-scaffold new`, `setup`, `localnet start` with default and custom port values
- [ ] Manual validation: works with both relative and absolute `[repos.lssa].path` in `scaffold.toml`

### DIFF
--- a/src/commands/doctor.rs
+++ b/src/commands/doctor.rs
@@ -6,7 +6,7 @@ use anyhow::bail;
 
 use super::wallet_support::wallet_password;
 use crate::commands::wallet_support::WALLET_CONFIG_PRIMARY;
-use crate::constants::DEFAULT_LSSA_PIN;
+use crate::constants::{DEFAULT_LSSA_PIN, SEQUENCER_BIN_REL_PATH};
 use crate::doctor_checks::{
     check_binary, check_container_runtime, check_path, check_port_warn, check_repo,
     check_standalone_support, one_line, print_rows,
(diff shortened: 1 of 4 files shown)


weboko (approved):
code looks good to me but need manually test it

fryorcraken:
> code looks good to me but need manually test it

In progress locally to ensure it's all clean.

PR #62, opened 2026-04-13
build: compile guest binaries in methods/ directory

### LINKED ISSUE #58: build: compile guest binaries in methods/ directory
## Problem

`logos-scaffold build` only runs `cargo build --workspace`, which skips the `methods/` directory. Risc0 guest crates are intentionally excluded from the main workspace (they target `riscv32im-risc0-zkvm-elf`), so they never get compiled by the build command.

## Current workaround

```bash
cargo build --release --manifest-path methods/Cargo.toml
```

## Expected behavior

`logos-scaffold build` should detect and compile guest binaries in `methods/` (or wherever the risc0 guest crates live) as part of the standard build pipeline, since deploying a program is the primary use case after building.

marclawclaw:
## Summary

\`logos-scaffold build\` only ran \`cargo build --workspace\`, which skips the \`methods/\` directory. Risc0 guest crates are intentionally excluded from the main workspace (they target \`riscv32im-risc0-zkvm-elf\`).

**Fix:** Added \`build_methods_guests()\` to \`src/commands/build.rs\`:
1. Detects \`methods/Cargo.toml\` in project root
2. Runs \`cargo build --release --manifest-path methods/Cargo.toml\` if found
3. Non-fatal if absent (works for non-Risc0 projects)
4. Called after workspace build in all three framework branches

## Testing
- \`cargo fmt --check\` ✅
- \`cargo check\` ✅

## Related
Fixes logos-co/logos-scaffold#58

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -1,0 +1,51 @@
+# SPEC: Compile Guest Binaries in methods/ Directory
+
+## Overview
+
+When running `logos-scaffold build` on a project that uses the LEZ framework (or any Risc0-based project), the build command must also compile guest binaries in the `methods/` directory. Currently `cargo build --workspace` skips the `methods/` directory because Risc0 guest crates target `riscv32im-risc0-zkvm-elf`, which is intentionally excluded from the main workspace build.
+
+## What the Fix Does
+
+After building the main workspace, `logos-scaffold build` will:
+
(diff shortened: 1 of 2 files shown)


marclawclaw:
Closed — local testing not yet complete.

PR #64, opened 2026-04-14
feat: create basecamp qml template

weboko:
### Description

This PR adds a new basecamp-qml template to logos-scaffold so it can bootstrap and manage pure Basecamp ui_qml projects, not just LEZ projects. The goal is to make the local Basecamp development loop straightforward: create a project, stage a raw plugin bundle locally, install it into a local Basecamp data dir, and optionally build LGX artifacts when Nix is available.

### Solution

I introduced project.kind with separate Basecamp config sections, added logos-scaffold new --template basecamp-qml, build --artifact raw|all, and a new install command. The generated template includes the Basecamp QML app files, local runtime docs, and a flake.nix wired to a pinned logos-module-builder. LEZ-only commands now fail clearly for Basecamp projects.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "logos-scaffold"
 version = "0.1.0"
-description = "Rust CLI for bootstrapping LSSA program_deployment projects in standalone mode"
+description = "Rust CLI for bootstrapping LEZ and Basecamp projects in local development mode"
 edition = "2021"
 license = "MIT OR Apache-2.0"
 readme = "README.md"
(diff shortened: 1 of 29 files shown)


weboko:
it's initial implementation, need to experiment with it more and polish

weboko:
closing as @fryorcraken added the template

wozos:
CI / validate is red on this PR because `cargo fmt --check` reports formatting drift against the current rustfmt stable (see the failed `validate` job).

I prepared a fmt-only fix on top of `weboko/basecamp-qml@01b06b7` and would have pushed it here, but I do not have write access to `logos-co/scaffold` so the push was rejected (HTTP 403) and "Allow edits by maintainers" is not enabled on this PR.

The fix is available at:
- branch: `wozos/scaffold:fix/pr-64-cargo-fmt`
- view: https://github.com/wozos/scaffold/tree/fix/pr-64-cargo-fmt

To apply it on this PR, either cherry-pick that commit onto `weboko/basecamp-qml` or just run `cargo fmt` and push. The change is pure whitespace / line-break formatting across 7 files.

PR #65, opened 2026-04-14
feat(localnet): add localnet reset command with --keep-wallet flag

marclawclaw:
Fresh re-implementation of PR #56 on current master (b93d80a).

## Summary

`logos-scaffold localnet reset [--keep-wallet]` — atomically resets the localnet environment to a clean state.

## Problem

Developers resetting localnet need to know undocumented internal file locations and execute deletions in the correct order. Failure modes are silent: `spel` hangs forever because sequencer on-chain nonces and wallet nonces are out of sync.

## Solution

Single command with atomic 8-step reset:
1. Stop sequencer (releases RocksDB lock)
2. Delete sequencer RocksDB at `.scaffold/cache/repos/lssa/rocksdb`
3. Delete wallet dir (unless `--keep-wallet`)
4. Delete `wallet.state` pointer (unless `--keep-wallet`)
5. Delete `localnet.state`
6. Run `setup` to recreate wallet
7. Start sequencer
8. Poll `get_last_block` RPC until block height > 0 (30s timeout)

## Testing
- `cargo fmt --check` ✅
- `cargo check` ✅

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -1,0 +1,47 @@
+# SPEC: `logos-scaffold localnet reset`
+
+## Context
+
+`logos-scaffold localnet reset [--keep-wallet]` — atomically resets the localnet environment to a clean state.
+
+## Problem
+
+Developers resetting localnet need to know undocumented internal file locations and execute deletions in the correct order. The failure mode when steps are skipped or reordered is silent: `spel` hangs forever on "Waiting for confirmation...", no errors anywhere, because sequencer on-chain nonces and wallet nonces are out of sync.
+
+## Solution
+
+Single command with atomic 8-step reset:
+1. Stop sequencer (releases RocksDB lock)
+2. Delete sequencer RocksDB at `.scaffold/cache/repos/lssa/rocksdb`
+3. Delete wallet dir (unless `--keep-wallet`)
+4. Delete `wallet.state` pointer (unless `--keep-wallet`)
+5. Delete `localnet.state`
+6. Run `setup` to recreate wallet and seed default address
+7. Start sequencer
+8. Poll `get_last_block` RPC until block height > 0 (30s timeout), confirming clean state
+
+## Usage
+
+```bash
+logos-scaffold localnet reset
+logos-scaffold localnet reset --keep-wallet
+```
+
+## CLI
+
+Add to `localnet` subcommand:
+- `reset` subcommand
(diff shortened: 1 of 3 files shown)


marclawclaw (commented):
## Code Review: `localnet reset` (5-axis)

### Verdict: **APPROVE** (posted as comment — can't approve own PR)

Clean, well-structured implementation. The 8-step sequence is correct and handles edge cases (missing files, stale PIDs, kill timeout) properly. A few suggestions below — none are blocking.

---

### Correctness ✅

The happy path and error cases look correct. One thing to be aware of:

**Double timeout in steps 7+8.** Step 7 calls `cmd_localnet(Start { timeout_sec: 30 })`, which internally runs `wait_for_readiness` (up to 30s for port open). Step 8 then adds *another* 30s for block height > 0. Worst-case total wait is **60s**, not 30s as the spec implies. This isn't a bug — both checks are needed (port open ≠ block produced) — but the effective timeout ceiling is double what the user might expect from the "30s timeout" language.

Consider either: (a) sharing a single deadline across steps 7–8, or (b) documenting the combined timeout in help text.

### Readability ✅

- Clear step numbering in output — easy to diagnose where a failure occurred.
- Good use of `with_context` on every fallible file operation.
- The `unreachable!` in `cmd_localnet_in_project` is the right pattern for the routing split.

### Architecture ✅

**Minor: double `load_project()`.** `cmd_localnet_reset` loads the project, then step 7 re-enters `cmd_localnet(Start{...})` which calls `load_project()` again. Not a correctness issue, but a direct call to `cmd_localnet_start(...)` (passing the paths you already have) would avoid the redundant project load and the interleaved "localnet ready (sequencer pid=X)" output that `cmd_localnet_start` prints between your step 7/8 messages.

**SPEC.md placement.** A spec file in the repo root is a bit unusual — if the project grows more specs, consider `docs/specs/` or keeping specs as PR descriptions only. Not blocking.

### Security ✅

- File deletions are bounded to known paths under `.scaffold/` and the lez directory — no path traversal risk.
- `kill` target comes from the state file (same trust model as existing `localnet stop`).

### Performance ✅

- 500ms poll interval is reasonable.
- The double `load_project()` is negligible in practice.

---

**Summary:** Ship it. The double-timeout behavior is the only thing worth a follow-up if users report confusion.

PR #74, opened 2026-04-20
fix(rpc): call getLastBlockId and parse result as u64

### LINKED ISSUE #73: rpc_get_last_block uses pre-refactor LEZ method name and response shape
## Summary

`rpc_get_last_block` in `src/commands/wallet_support.rs:396` calls the pre-refactor LEZ RPC method `\"get_last_block\"` and parses `result.last_block` as a nested field. The pinned LEZ version (`35d8df0d`) has already renamed this to **`getLastBlockId`**, returning `result: <u64>` directly, so every call returns `-32601 Method not found`.

This silently breaks:
- `logos-scaffold localnet reset` — block-production verification times out with `VerificationPollFailed`.
- `logos-scaffold wallet topup` preflight — likely the same failure mode; worth auditing.

Any other command that depends on `rpc_get_last_block` should be checked.

## Root cause

LEZ PR #394 (merged 2026-03-20, commit `fe368f2b`) refactored the sequencer RPC server-side:
- `get_last_block` (snake_case, returned a full block object) → `getLastBlockId` (camelCase, returns just the u64 block ID)
- `send_tx` → `sendTransaction`, `get_account_balance` → `getAccountBalance`, etc.

Our pin `35d8df0d` (2026-04-08) is **19 days after** that refactor, so we're not on an old LEZ — the scaffold client code was simply never updated when the RPC surface changed.

## Reproduction

From a project pinned at `DEFAULT_LEZ_PIN`:
\`\`\`bash
logos-scaffold localnet start
logos-scaffold localnet reset
# error: verification poll failed: get_last_block response missing \`result.last_block\`: {\"error\":{\"code\":-32601,\"message\":\"Method not found\"},...}
\`\`\`

Direct RPC check:
\`\`\`bash
curl -s -X POST http://127.0.0.1:3040 -H 'Content-Type: application/json' \\
  -d '{\"jsonrpc\":\"2.0\",\"method\":\"getLastBlockId\",\"params\":{},\"id\":1}'
# => {\"jsonrpc\":\"2.0\",\"id\":1,\"result\":1062}
\`\`\`

## Proposed fix

1. Update `rpc_get_last_block` in `src/commands/wallet_support.rs`:
   - Method name: `\"get_last_block\"` → `\"getLastBlockId\"`
   - Response: parse `result` as `u64` directly, not `result.last_block`.
   - Consider renaming the fn to `rpc_get_last_block_id` to match the semantic change (block → block ID).
2. Audit all other RPC calls in the scaffold against the current `sequencer/service/rpc/src/lib.rs` method list. `grep '\"method\"'` currently returns only this one site, but any future additions should be checked too.
3. Add an integration-style test (or at least a comment) that would catch a future LEZ RPC rename — e.g., a preflight that asserts the method exists before relying on it.

## Context

Surfaced while dogfooding PR #56 (\`localnet reset\`). The reset command itself works end-to-end; this is an orthogonal pre-existing issue.

fryorcraken:
## Summary

Fixes #73.

The sequencer's RPC surface was refactored in LEZ PR #394 on 2026-03-20: `get_last_block` was renamed to `getLastBlockId` and the response changed from `{\"result\":{\"last_block\":<u64>}}` to `{\"result\":<u64>}`. The scaffold's pinned LEZ (`35d8df0d`, 2026-04-08) is 19 days past that refactor, but the scaffold client at `src/commands/wallet_support.rs:396` was never updated — so every call returns `-32601 Method not found`. This silently breaks:

- `logos-scaffold localnet reset` block-production verification
- `logos-scaffold deploy` sequencer-reachability preflight

## Changes

- `rpc_get_last_block` → **`rpc_get_last_block_id`** (name reflects the semantic change: block → block ID).
- Method string: `\"get_last_block\"` → `\"getLastBlockId\"`.
- Response parsing: `result.last_block` (nested) → `result` (bare `u64`).
- Updated the single caller in `deploy.rs`.
- Updated unit tests to match the new response shape; added a new test asserting `-32601 Method not found` surfaces as an error.

## Verification

- `cargo fmt --check` — clean
- `cargo test --all-targets` — 53 passed
- Dogfooded against live sequencer in `~/src/fryorcraken/cryptarchs`: started localnet → `curl getLastBlockId` returned `{\"result\":3}`; unit tests cover the parse/error paths.

## Test plan

- [x] Unit tests cover: valid response, connectivity error, non-numeric `result`, `Method not found` response.
- [x] Manual: `logos-scaffold localnet reset` (with PR #56) — verification step succeeds rather than timing out. Dogfooded 2026-04-20 against live sequencer in `~/src/fryorcraken/cryptarchs`: reset reported `block_height=1` after restart.
- [ ] Manual: `logos-scaffold deploy` — preflight passes rather than warning about probe failure.

### DIFF
--- a/src/commands/deploy.rs
+++ b/src/commands/deploy.rs
@@ -9,7 +9,7 @@ use crate::project::load_project;
 use crate::DynResult;
 
 use super::wallet_support::{
-    extract_tx_identifier, is_connectivity_failure, load_wallet_runtime, rpc_get_last_block,
+    extract_tx_identifier, is_connectivity_failure, load_wallet_runtime, rpc_get_last_block_id,
     sequencer_unreachable_hint, summarize_command_failure, wallet_password, RpcReachabilityError,
 };
 
@@ -168,7 +168,7 @@ pub(crate) fn cmd_deploy(
 }
 
 fn preflight_sequencer_reachability(sequencer_addr: &str) -> DynResult<()> {
-    match rpc_get_last_block(sequencer_addr) {
+    match rpc_get_last_block_id(sequencer_addr) {
         Ok(_) => Ok(()),
         Err(RpcReachabilityError::Connectivity(err)) => {
             bail!(
--- a/src/commands/wallet_support.rs
+++ b/src/commands/wallet_support.rs
@@ -415,6 +415,19 @@ pub(crate) fn rpc_get_last_block_id(sequencer_addr: &str) -> Result<u64, RpcReac
         ))
     })?;
 
+    if let Some(err_obj) = body.get("error") {
+        let code = err_obj.get("code").and_then(Value::as_i64);
+        let message = err_obj.get("message").and_then(Value::as_str).unwrap_or("");
+        let formatted = match code {
+            Some(c) => format!("getLastBlockId RPC error {c}: {message}"),
+            None => format!(
+                "getLastBlockId RPC error: {}",
+                one_line(&err_obj.to_string())
+            ),
+        };
+        return Err(RpcReachabilityError::Other(formatted));
+    }
+
     body.get("result").and_then(Value::as_u64).ok_or_else(|| {
         RpcReachabilityError::Other(format!(
             "getLastBlockId response missing numeric `result`: {}",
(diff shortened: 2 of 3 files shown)


jzaki (commented):
Looks good to me

danisharora099 (approved):

PR #75, opened 2026-04-20
feat(basecamp): profile-isolated basecamp with modules/install/launch/build-portable/doctor

fryorcraken:
## Summary

Adds a complete `basecamp` subcommand surface to `logos-scaffold` for profile-isolated Logos basecamp dogfooding, plus the review-finding follow-up from @vpavlin.

### What this ships

**Core commands**

- `basecamp setup` — pin + build basecamp + lgpm; seed alice/bob XDG profile trees under `.scaffold/basecamp/profiles/`. Builds logged to `.scaffold/logs/<ts>-setup-*.log` with a one-line progress status + duration (on both success and failure).
- `basecamp modules` — **sole writer** of the captured module + dependency set (`basecamp.state`). Three modes:
  - Zero-arg: full auto-discovery. Walks project flakes (root `.#lgx` first, else immediate sub-flakes), reads each source's `metadata.json` for declared `dependencies`, resolves each against `[basecamp.dependencies]` in `scaffold.toml` → scaffold-level `BASECAMP_DEPENDENCIES` pin table → preinstalled-modules skip → unknown-dep warning.
  - `--flake <ref>... --path <p>...`: explicit project sources; deps still resolved from their manifests.
  - `--show`: read-only; prints current capture with provenance annotations.
- `basecamp install` — **pure replay**. No source-set flags. Reads state, builds deps first (fail-fast) then project sources, installs both into every seeded profile via `lgpm`. On empty state, transparently invokes `modules` in auto-discover mode. Overwrites existing installs. Logs per nix build to `.scaffold/logs/`; `--print-output` (or `LOGOS_SCAFFOLD_PRINT_OUTPUT=1`) streams nix output directly for CI.
- `basecamp launch <profile>` — scrub data/cache, replay captured modules, assign per-profile ports, exec basecamp with profile-scoped XDG env. Pre-flight prints a one-line summary of installed-module variant coverage so the freeze-on-first-click case (manifest variant mismatch) is visible before the user clicks. `--no-clean` skips the scrub.
- `basecamp build-portable` — zero-arg; rebuilds `state.project_sources` with attr-swap `#lgx` → `#lgx-portable` for hand-loading into a basecamp AppImage. Deps are intentionally skipped (the AppImage provides its own release companions).
- `basecamp doctor [--json]` — basecamp-specific health report: captured modules summary (with github-ref tag/commit parsing + installed API header paths), manifest variant checks per seeded profile, and module-set drift between auto-discovery and captured state.

Also on this branch: a resolution chain for `logos-scaffold`'s `cache_root` (env → `scaffold.toml` → `XDG_CACHE_HOME` → `HOME/.cache`) so a committed `scaffold.toml` builds on CI without machine-specific paths.

**State model**

- `basecamp.state` splits `sources` into `project_sources` (dev's own modules) and `dependencies` (runtime companions).
- `[basecamp.dependencies]` in `scaffold.toml` — per-project override for companion pins; keyed by module name.

**Three review findings, three upstream issues filed**

From @vpavlin's dogfood review ([review](https://github.com/logos-co/logos-scaffold/pull/75#pullrequestreview-4153746923), [reply](https://github.com/logos-co/logos-scaffold/pull/75#issuecomment-4301577873)), three asks are punted upstream where they belong:

1. [logos-co/logos-basecamp#167](https://github.com/logos-co/logos-basecamp/issues/167) — basecamp v0.1.1 hardcodes `logos-modules/latest` as its package catalog release tag; should pin at build time so scaffold can derive companion pins per basecamp release instead of carrying opinionated defaults.
2. `logos-module-builder` upstream (issue pending): packaging should emit both `<platform>` and `<platform>-dev` keys in manifest `main`, so a `.lgx` built from `.#lgx-portable` loads cleanly into dev basecamp (currently a silent-hang failure mode).
3. [logos-co/logos-tutorial#39](https://github.com/logos-co/logos-tutorial/issues/39) — tutorial should document the compatible `lgpm` pin (current `latest` is incompatible with `tutorial-v1` module-builder output).

Parked for a follow-up PR (per the reviewer's verdict \"approve modulo A + 1\", both A + 1 are shipped here; A's `--preload` flag is the park):

- Opt-in `basecamp launch --preload <csv>` — intentionally deferred so module authors still see the lazy-load freeze bug in their own dogfood, rather than scaffold papering over a module-side sync-IPC-in-\`Component.onCompleted\` footgun that shipped AppImage users would still hit.

**Documentation**

- `README.md` CLI surface + command semantics sections updated for every new / changed basecamp subcommand.
- `docs/basecamp-module-requirements.md` documents the \`transitive inputs must \`follows\`\` rule for multi-sub-flake projects.
- `FURPS.md` v0.3 section covers build-portable + modules model; ADR covers the architectural decisions (portable separate from install, AppImage path out of scope, flake attr is a resolver parameter).

## Test plan

- [x] `cargo test` — 169 lib + 97 integration tests pass.
- [x] `cargo fmt --all --check` clean.
- [x] `cargo build` clean.
- [x] Dogfood on `logos-module-tictactoe`: `setup → install → launch alice` end-to-end, with alice + bob coexisting cleanly.
- [ ] Fresh dogfood post-merge against vpavlin's `logos-yolo-board-module` to verify the `basecamp modules` auto-discovery + manifest-dep resolution captures `delivery_module` correctly against the scaffold default pin.


vpavlin (commented):
Dogfooded this against our real out-of-tree module (`logos-yolo-board-module`, a Yolo Board plugin that subscribes via `delivery_module` for per-message threads) on Linux x86_64. Overall: **really liked it**. CLI builds cleanly (cargo 1.91), `init` → `setup` → `install` → `launch` matches the doc, and the auto-discovery of `packages.<sys>.lgx` Just Worked with no tweaks to our flake. Huge quality-of-life win over the manual `lgpm install` + variant-patching + XDG-shuffle recipe we were doing before.

Below are the things I hit that seem worth a look before merge.

## Signals that landed well

- **`basecamp setup`**: pin check, cache layout, profile seeding all reasonable. The `HEAD is now at a746cdb` checkout output is visible, which is helpful.
- **`basecamp install` (single source)**: auto-discovered our root-flake `#lgx` output, built into `<cache>/basecamp/lgx-links/`, lgpm-installed into both alice and bob. `.state` records the resolved source unambiguously.
- **Safety guards around `launch`** (profile allowlist, path-separator rejection, canonicalized-prefix scrub) read clean — nice defense-in-depth.
- The "Concurrent alice/bob on the same host may collide on module-level ports" comment in `cmd_basecamp_launch` calling out delivery_module#18 is exactly the right framing.

## Findings worth discussing

### A. Lazy plugin load on sidebar click freezes Basecamp — `launch` needs a pre-load hook

**Largest ergonomic finding.** `launch <profile>` execs `basecamp` with XDG env only — no `-l <modules>` pre-load arg. With our plugin architecture that's fatal for the sidebar-click workflow:

- User clicks "Yolo" in Alice's sidebar
- Basecamp starts loading the `yolo_board` UI plugin
- Its QML `Component.onCompleted` makes a sync IPC call to `yolo_board_module` (e.g. `set_ui_dir`)
- `yolo_board_module` host hasn't spawned yet (lazy load by dep-resolution kicks in only on the call)
- Sync QRO call blocks the Qt main thread for its 20 s timeout → **the entire Basecamp window is frozen**
- After the timeout, basecamp eventually spawns the module, but by then the first sync call has already errored and the UI plugin's state machine is wedged

Workaround I confirmed: launch basecamp by hand with `-l liblogos_zone_sequencer_module,storage_module,delivery_module,yolo_board_module` before interacting with the UI. Everything loads cleanly and Yolo works end-to-end.

Two complementary fixes:

1. **Scaffold side**: `launch <profile> --preload <mod1>,<mod2>,...` (or just auto-derive from `<profile>/modules/*/manifest.json` → pre-load every declared core module). The flag is a one-line change in `cmd_basecamp_launch`; the auto-derivation is 5 lines. Until this lands, any realistic plugin (i.e. any plugin with backend dependencies declared in `metadata.json`) will freeze on the sidebar click.
2. **Our QML side** (our problem, not yours): avoid sync IPC in `Component.onCompleted`. `QTimer::singleShot(0, ...)` or async variants both work. I'll fix this in `zone-sdk-test`.

Without #1, `basecamp launch alice` produces a window where clicking the plugin you just installed via `basecamp install` freezes the UI. That's a particularly bad first-run story, so this feels blocker-adjacent even though the correctness of launch itself is fine.

### 0. Two-instance XDG isolation actually works

For the record: `setup` → `install` (with 5 explicit sources — our yolo plugin pulls 3 runtime deps + a UI plugin) → `launch alice --no-clean &` → `launch bob --no-clean &` brought up two `LogosBasecampDev` windows side by side, each with its own module tree rooted in `.scaffold/basecamp/profiles/{alice,bob}/xdg-data/`. `ps | grep logos_host` cleanly partitions 2 processes per profile. No QRO socket name collisions. This is the first time I've had two basecamps running at once without hand-patching `XDG_DATA_HOME` — thanks.

### 1. `install` with no args doesn't replay recorded state — it auto-discovers and replaces

Repro: `logos-scaffold basecamp install --flake a#lgx --flake b#lgx --flake c#lgx` where `b` fails to build.

Observed behavior:
- `basecamp.state` is updated with **all three** `source:flake=` entries before builds start.
- Only `a` makes it into profiles before the `b` build exits with a non-zero status.
- Subsequent `install <no args>` would be the natural fix — "re-try whatever was recorded". But that **isn't** what bare `install` does: it runs the project-root auto-discover path and overwrites state with whatever it finds (typically the single root `.#lgx`), replacing the 3 entries with 1. So "replay" isn't a command any more; I have to reconstruct the full `--path/--flake` CLI from memory or parse `.scaffold/state/basecamp.state` by hand.
- `launch <profile>` without `--no-clean` does replay-install from state — that *is* the recorded set — and it will fail repeatably until I manually edit `.scaffold/state/basecamp.state`.
- The only non-destructive recovery is a hand-edit of `basecamp.state`; `basecamp reset` nukes both profiles.

Concrete case: I was trying to install yolo + its runtime deps (`logos-storage-module`, `logos-delivery-module`, zone-sequencer) in one shot. `logos-storage-module#lgx` hits a well-known upstream NAR-hash mismatch on `logos-storage-nim` — the build failed, state recorded 4 sources anyway, and every subsequent `install` / `launch` re-ran the broken storage build.

This is probably a straight spec ambiguity. "No args → auto-discover and replace" + "all-or-nothing on build failure undefined" combine badly: a broken discovery run + a broken partial install = state is wedged and the natural commands to un-wedge it reinforce the break.

Options I'd consider:

- **Make `install <no args>` replay recorded state when state exists and auto-discover when it doesn't.** This is what `launch` already assumes.
- **Atomic**: build every source into a staging area first, only mutate `basecamp.state` + profiles if all succeed.
- **Partial with visibility**: record successful sources only; surface a summary like `installed: a; failed: b; skipped: c` and exit non-zero.
- **Explicit `--keep-going` flag** with today's partial behavior gated behind it.
- **`--replay` / `--rediscover` / `--clear` flags** so the user controls which semantic they're invoking rather than it being inferred from args count.

Today's behavior is mid-stream-crash bad because it leaves subsequent `install` / `launch` calls broken until you manually fix state.

### 2. Transitive module deps aren't discovered from `metadata.json`

`install` builds + installs the targeted flake's `.lgx` but doesn't walk the installed module's `metadata.json` → `dependencies: [...]` list to pull those sources too. For any non-trivial module (ours declares 3 runtime deps) the user has to repeat `--flake …#lgx` per dep.

This is fair as a v1 decision — transitive resolution is real work, especially since the `.lgx` for dependency X doesn't itself know where to find its flake — but I wanted to flag it because the "install this module" mental model feels incomplete when you click Yolo in basecamp and it refuses to load due to missing `delivery_module` / `storage_module`.

A tiny UX improvement short of full auto-resolution: on first successful `install`, peek at `<profile>/modules/<name>/manifest.json` → `dependencies`, and if any aren't present under `<profile>/modules/`, print a `hint: module X depends on [A, B]; neither is installed` footer. Cheap, informative, no new resolver logic.

### 3. Error quality on a `lgx-portable`-only flake

I didn't repro this one, but based on reading `resolve_basecamp_binary` and the "only `lgx-portable` exposed" branch in the resolver, the hint is good — pointing users at `--flake X#lgx-portable` or `build-portable`. If there's a test around it (looks like there is in `tests/cli.rs` from the diff stats) this is already covered.

### 4. Minor — setup output is a bit chatty during nix builds

`nix build .#app` from the pinned basecamp stretches over ~60 s of output that scaffold forwards verbatim. In a CI / terminal log that's fine, but for the interactive first-run experience a spinner + tail on final summary would feel tighter. Not a blocker.

### 5. Pre-flight check suggestion: `basecamp setup` should refuse if there's a conflicting Basecamp already running

I had a regular `logos-basecamp` (non-scaffold) running in the background when I did the scaffold setup. The setup proceeded fine — but when I later did `basecamp launch alice` it would've clashed on QRO local-socket names with my existing `logos-basecamp`. Spec §2.3 says "v1 doesn't lock; document as 'don't do that'" — but a quick `pgrep -f LogosBasecamp` pre-flight at launch time with a targeted "there's already a basecamp running under pid X; kill it or use a different profile" hint would save a confusing subsequent failure.

### 6. Non-scaffold papercut worth a `doctor` check: manifest variant mismatch

Unrelated to the PR itself but surfaced by it: a bunch of the `.lgx` artefacts I consumed (both ours: `yolo-board-module.lgx` / `yolo-board.lgx`, and third-party: `jimmy-claw/logos-zone-sequencer-module#lgx`) end up with `manifest.json` `main` keys of `linux-amd64` instead of `linux-amd64-dev`. `basecamp v0.1.1` expects `-dev`. Symptom: a Basecamp that launches and runs, but clicking the plugin silently hangs (dep-resolver fails variant-matching, times out).

This is a module-side bug in each of those flakes, not scaffold's. But scaffold is now the first tool that makes the failure visible end-to-end, and the diagnostic is opaque — no stderr line, no hint, just a frozen window. A `basecamp doctor` subcommand (or a pre-flight inside `launch`) that walks installed `<profile>/modules/*/manifest.json` and flags modules whose `main` has no `-dev` variant would save the next user a bad afternoon. Even just a one-line log in `launch` — "profile has N modules; all variants match `-dev`: [y/n]" — would be massive.

## Overall

## Overall

Would approve modulo **A** (preload on `launch` — without it, freshly-installed real plugins freeze on first click) and **1** (`install <no-args>` replace-vs-replay semantics). Both are pretty surgical fixes. #2-6 are "take them or leave them" based on your v1 scope.

Nice work — this is already the fastest path I have to two-instance Basecamp dogfooding. Thanks for driving it.

**Environment**: Linux x86_64, Nix 2.x, scaffold built from PR head at `e010004` via `cargo build --release`. Module under test: `vpavlin/logos-yolo-board-module` — a Yolo Board plugin with 3 core-module deps (`storage_module`, `delivery_module`, `liblogos_zone_sequencer_module`) and a separate UI plugin flake. Full repro recipe + more context in [docs/threads-howto.md](https://github.com/vpavlin/logos-yolo-board-module/blob/master/docs/threads-howto.md) on that repo.

fryorcraken:
@vpavlin thanks for the thorough dogfood — addressing each finding below. The two gates (A + 1) end up reshaped together into a design change that subsumes finding 2 as well; the rest are in-scope or explicitly parked with rationale.

## Findings 1 + 2 (merged into one refactor)

Both point at the same root issue: scaffold conflates "discover project modules on every `install`" with "ship what the user asked for last time." Collapsing them into a clearer model:

**New command: `basecamp modules`** — the sole writer of the captured source set. Two modes:

- `lgs basecamp modules` — auto-discover: walks project flakes (root `.#lgx` first, else immediate sub-flake `.#lgx`s), reads each source's `metadata.json` for declared `dependencies`, resolves each dep name against scaffold's per-basecamp-pin dependency table (overridable via `[basecamp.dependencies]` in `scaffold.toml`).
- `lgs basecamp modules --flake <ref>... --path <p>...` — explicit project sources; no project-flake discovery, but dep resolution still walks the supplied sources' manifests.
- `lgs basecamp modules --show` — read-only; prints both lists labeled.

Both mutating modes replace state and print the previous set first so reverting is copy-paste.

**State splits into `project_sources` + `dependencies`.** Different commands consume different subsets:

- `install` → pure replay; no `--flake`/`--path`/`--profile`. Replays both lists into every seeded profile. Dependencies build first (fail-fast on broken companion pins). If state is empty on first call, transparently invokes `basecamp modules` auto-discover, prints what was captured, then proceeds.
- `build-portable` → zero-arg; attr-swaps `#lgx` → `#lgx-portable` on `project_sources` only. Dependencies are assumed present in the target AppImage (its preinstall/Package-Manager catalog), so scaffold doesn't produce portable variants for them.

**`logos-scaffold doctor` gains a drift check**: runs the same auto-discovery as a dry-run and flags divergence between discovered and captured sets. The one-time-capture model is self-policing.

**Scaffold-level dependency pin table** in `src/constants.rs`, keyed by module name, paired with `DEFAULT_BASECAMP_PIN`. Seeded initially with `delivery_module → github:logos-co/logos-delivery-module/1.0.0#lgx`; storage and others added when projects need them. `[basecamp.dependencies]` in `scaffold.toml` overrides per-project. This is a deliberate opinionated choice — see the upstream issue I filed on basecamp: https://github.com/logos-co/logos-basecamp/issues/167 — until basecamp pins its `logos-modules` release tag at build time, scaffold can't derive these automatically.

## Finding 3 — `lgx-portable`-only error quality

Appreciated. Test coverage you spotted is correct (resolver-level unit tests plus `basecamp_build_portable_inside_empty_project_emits_generic_hint` in `tests/cli.rs`). The findings 1+2 refactor moves `--flake` off `install` and onto `basecamp modules`, so the portable-only and generic-hint error texts are updated to reference `basecamp modules --flake <ref>#lgx-portable` / `--path <file.lgx>` instead. Same semantics, correct command.

## Finding 4 — chatty nix output during setup

You're right that UX is load-bearing for scaffold. Landing in this PR:

- Every nix invocation logs to `.scaffold/logs/<timestamp>-<command>.log`.
- Terminal (TTY) shows a checkbox tree per step/source/dep: `○` pending, `⋯` in-progress w/ elapsed time, `✓` success w/ duration, `✗` failed.
- First line of each command prints `tip: tail -f <abs log path>` so you can watch the build in another pane.
- Non-TTY (CI) gets line-per-update — same info, no ANSI rewrites.
- `--print-output` (or `LOGOS_SCAFFOLD_PRINT_OUTPUT=1`) bypasses the log file and streams nix output directly, for CI that already has structured log capture.
- Errors always surface the absolute log path.

Same pattern as npm/pnpm with a slightly richer on-screen progress view. Using `indicatif` for the tree rendering.

## Finding 5 — `basecamp setup` / `launch` pre-flight for a conflicting basecamp

Investigated and I think this is a false alarm at the QRO layer. Checked `/proc/net/unix` on a host running basecamp and v0.1.1 nonce-suffixes every QRO local socket: `/tmp/logos_core_manager_555aa34230f6`, `/tmp/logos_capability_module_000d590d1910`, etc. Each basecamp launch picks a unique 48-bit value (I can see dozens of distinct suffixes in `/tmp` from past runs), so two basecamps — two scaffold profiles, or scaffold-launched + non-scaffold — don't collide at the QRO layer. This is also what let alice + bob coexist cleanly in the dogfood earlier.

What I can't rule out without reading basecamp source more thoroughly is a fixed-name singleton (D-Bus service name, QtSingleApplication-style lock). If you've actually observed a collision symptom rather than inferred it from QRO, I'd love a repro. Otherwise I'm dropping this finding — happy to revisit if something else surfaces.

## Finding 6 — manifest variant mismatch (`-dev` vs release)

Folding into the same `doctor` expansion. The drift-check walk reads `modules/*/manifest.json` already; one filesystem pass answers both questions. `doctor` flags any module whose `main` lacks the current platform's `-dev` key. `launch` gets a one-line pre-flight summary — either *"all modules have `<platform>-dev` variants ✓"* or *"N modules missing `<platform>-dev` variant (`<names>`); run `doctor` for details"* — so the freeze-on-click case is caught before you click.

Filing an upstream issue on `logos-module-builder` suggesting the packaging step always emits both `<platform>` and `<platform>-dev` keys.

## Finding A — lazy plugin load freeze → `launch --preload`

Parking for a follow-up issue after this PR merges, with two reasons:

1. **The core bug is module-side, as you noted.** Sync IPC in `Component.onCompleted` is the root cause; `QTimer::singleShot(0, …)` fixes it. If scaffold auto-preloads by default, developers dogfooding via `basecamp launch` won't see the freeze locally — but their users running a shipped AppImage (no scaffold in the loop) still will. Scaffold papering over it during dev gives the developer a clean first-run experience locally and their users a bad one in production.
2. **PR #75 is already a lot of surface.** Cleaner to land core setup/install/launch/reset/build-portable as-is plus the UX + drift changes above and iterate.

My current thinking for the follow-up: opt-in `--preload <csv>` flag (no default, no auto-derive), with docs framing it as a debugging aid and explicitly calling out that if the flag is what makes your plugin work on first click, there's a module-side lazy-load bug real users will still hit. Will open the tracking issue once this merges.

---

Also filing out of band:

- https://github.com/logos-co/logos-basecamp/issues/167 — basecamp to pin its `logos-modules` release tag at build time.
- `logos-module-builder`: packaging should emit both `<plat>` and `<plat>-dev` keys in manifest `main`.
- `logos-co/logos-tutorial#39` (already filed): document the compatible `lgpm` pin.

Thanks again — dogfood findings like these are how this gets real.

vpavlin:
## Re-review at `e476fd1` against the same out-of-tree yolo module

Re-ran the full flow end-to-end on `logos-yolo-board-module` (with its 3 core-module deps + UI plugin). Bottom line: **everything I flagged is addressed, and the `doctor` + `launch` variant pre-flight is exactly the UX I was hoping for.** A few small things below; none blocking.

### What landed cleanly

**Findings 1 + 2 refactor** — the `modules` / `install` / state split works as described. Fresh walk:

1. `basecamp modules` on a project with only `delivery_module` in scaffold defaults and 3 declared deps auto-resolves delivery and prints clear, actionable warnings for the two unresolved ones naming all three mitigation paths (pin in `[basecamp.dependencies]`, add flake input, or explicit `--flake`). Nice.
2. `basecamp modules --flake … --flake …` cleanly replaces state. Prints the previous set first, so reverting is copy-paste as promised.
3. `basecamp modules --show` is a clean read-only dump — separate from the mutating path, no chance of accidental rewrite.
4. State split is visible in `basecamp.state`: `project:flake=` vs `dep:flake=` lines, easy to reason about.
5. `install` with no args is pure replay. Deps build first (42 s for delivery) before project sources. When the storage build failed with the known upstream `logos-storage-nim` NAR-hash mismatch, **nothing was installed into `alice` or `bob`** — partial-state problem from v1 is gone. State stayed intact so retry is trivial.
6. `build-portable` zero-arg, attr-swap on `#lgx` → `#lgx-portable`. Correct behavior.

**Finding 4 UX** — `.scaffold/logs/<timestamp>-<command>.log` persistence is a huge win. Each log file has a header with session/started/step/command, then full nix output. `tip: tail -f <abs-path>` up front + checkbox tree with durations (`⋯` → `✓ (41.5 s)`) is exactly the npm/pnpm feel.

**Finding 6 (`-dev` variant mismatch)** — this is the star of the show for me. `doctor` flagged it cleanly:

```
WARN | basecamp variant: liblogos_zone_sequencer_module in profile alice
     | main=[linux-amd64,linux-x86_64] missing expected `linux-amd64-dev`;
       plugin will hang on click
  remediation: rebuild `liblogos_zone_sequencer_module` so its manifest.json
               `main.linux-amd64-dev` key is populated (upstream
               logos-module-builder issue); then re-run `basecamp install`
```

And then `basecamp launch alice`:

```
launch: profile alice has 4 module(s); 1 missing linux-amd64-dev variant
        (liblogos_zone_sequencer_module); run `logos-scaffold doctor` for
        details — plugins will hang on click.
```

Exactly the warning I got burned by. Names the bad module, links the rationale, points at `doctor`. This is going to save a lot of next-user pain.

**Finding 5 (conflict detection)** — fair call, the nonce-suffix evidence is convincing. Dropping is fine; I don't have a concrete repro either.

**Finding A (preload)** — the "scaffold papering over a module-side Component.onCompleted bug hides it from devs but not from end-users" argument is well-framed. Opt-in `--preload <csv>` is the right shape. Agreed with parking.

### Smaller things I noticed

These are nits / v1.1 items, not blockers:

1. **Drift check false-positives for explicit captures.**
   After `basecamp modules --path /nix/store/…-logos-storage_module-lgx-1.0.0/…` and `--flake github:jimmy-claw/…#lgx`, `doctor` reports `WARN | basecamp drift: stale | captured <x> no longer discoverable`. The `/nix/store/…` path clearly exists and is what I explicitly captured; the `github:…#lgx` is an intentional explicit reference. Drift-against-auto-discovery makes sense for things the user didn't name explicitly, but I'd suggest skipping the stale-drift row when the source was captured via explicit `--path` / `--flake`. The same `modules` output already labels them `[explicit]` — that flag could gate drift checks.

2. **Install failure log path needs a `cat`-able hint.**
   When the storage build failed, the error was:
   ```
   error: building github:…#lgx failed with exit status: 102; see /home/…/.scaffold/logs/20260423-073125-install.log
   ```
   Perfect path. But when I tailed it, the last ~30 lines had the stack trace truncated with `'stack trace truncated; use --show-trace to show the full trace'`. Could scaffold re-invoke with `--show-trace` on failure and persist that to `-full.log` alongside? Cheap, and it'd save one round-trip.

3. **`build-portable` when flake has no `lgx-portable` attr gets raw nix error.**
   Output:
   ```
   error: flake '…' does not provide attribute 'packages.x86_64-linux.lgx-portable', 'legacyPackages.x86_64-linux.lgx-portable' or 'lgx-portable'
   ```
   This is the mirror of Finding 3 and deserves the same treatment: wrap with "project flake doesn't expose `lgx-portable`. Either (a) add an `lgx-portable` output to your flake, or (b) if you don't need a portable build, don't run `basecamp build-portable`." Link the module-builder docs for the naming convention.

4. **Dep-resolution warnings keep firing after the user provides the dep as an explicit `--flake` project source.**
   Repro:
   ```
   basecamp modules --flake …#yolo-lgx --flake github:…/logos-storage-module#lgx --flake github:jimmy-claw/logos-zone-sequencer-module#lgx
   ```
   Yolo declares `storage_module` + `liblogos_zone_sequencer_module` + `delivery_module` as deps. Delivery resolves via scaffold default. Storage + zone-sequencer are in the captured `project_sources`. The dep-walk is oblivious to that and still prints:
   ```
   warning: dep `storage_module` declared by …#yolo-lgx is not pinned in scaffold defaults, not in [basecamp.dependencies], not in the source's flake.lock, and is not a preinstalled basecamp module; skipping. Add it to …
   ```
   Since the user did in fact cover it (just via a different axis), the warning is noise. A quick pass "for each unresolved dep, before warning, check if any captured project_source's manifest has `name == <dep>` and skip the warning if so" would kill false positives.

5. **`reset` plan line counts inconsistent with state.**
   `reset` said `clear basecamp.state sources (4 recorded)` when the state held 4 project_sources + 1 dependency. Guessing it's counting project_sources only. Minor — either count both (`5 recorded`) or split (`clear 4 project sources + 1 dependency`).

6. **`profile list` still stubbed.**
   `basecamp profile list` returns `error: basecamp profile list is not yet implemented`. Scope call, but the subcommand shows up in `--help` so it looks supported. Either hide until implemented, or wire even a minimal version (the list is already `alice, bob` today).

### Environment

Linux x86_64, Nix 2.x, scaffold at `e476fd1` built via `cargo build --release`. Module under test: same `vpavlin/logos-yolo-board-module` as last time (delivery + storage + zone-sequencer + yolo UI plugin). Re-review did setup → modules (auto + explicit) → install (fail-fast + replay) → doctor → launch → reset in ~15 min wall time, most of it nix builds.

### Overall

Would **approve**. Shape is right, all the landed changes match what you described, and the `doctor` + `launch` variant pre-flight is the single best v1.0 UX addition you could have made. The nits above are fine to pick up in follow-ups.

Thanks again.

fryorcraken:
Surfacing a finding from dogfooding this PR's `basecamp launch alice` path with a three-tier `UI → core → delivery_module` topology in [`fryorcraken/logos-module-tictactoe`](https://github.com/fryorcraken/logos-module-tictactoe):

**Dev `.lgx` (the `#lgx` attr, `linux-amd64-dev`, Nix-store `RUNPATH`) does not work against a source-built basecamp when the module pulls in `delivery_module`.** The `logos_host` child trips basecamp's 2 s token-handshake budget during `dlopen` of the unbundled Nix-store library chain, parent `terminate()`s it, the module never comes up, and the UI's QRO replica for `delivery_module` is left peerless.

**`#lgx-portable` (`linux-amd64`, `$ORIGIN`-rpath, self-contained) in the same source-built basecamp works end-to-end** — same repo, same basecamp pin, same commit. The bundled-lib layout fits inside the handshake budget where the Nix-store one doesn't.

Full root-cause + repro + standalone `logos_host` verification + variant-key gotcha filed upstream as **[logos-co/logos-basecamp#169](https://github.com/logos-co/logos-basecamp/issues/169)**.

**Recommendation for this PR:** until basecamp#169 lands (either a configurable/longer handshake budget, or a known-working dev `.lgx` path for the UI→core→delivery topology), have `lgs basecamp install` default to `#lgx-portable` for scaffolded projects rather than `#lgx`. The existing `basecamp build-portable` command already does the attr swap; the suggestion is to make that the default for `install` too and flip `#lgx` into an opt-in (e.g. `--dev-variant`) once upstream has a working example.

One gotcha if you go this way: `lgpm` on source-built basecamp rejects portable `.lgx` archives with `Error: Package does not contain variant for platform: linux-x86_64-dev` because portable ships `linux-amd64` in the manifest. We worked around it with a repack step (rename `variants/<portable-key>/` → `variants/linux-x86_64-dev/` + rewrite `manifest.json.main` + drop `hashes`). Scaffold could absorb this as part of install until basecamp#169 (3) unifies the variant vocabulary.

weboko:
I haven't unleashed my AI agent, but skimming through PR would be good to reflect new experience in https://github.com/logos-co/logos-scaffold/blob/master/DOGFOODING.md

PR #82, opened 2026-04-22
fix(deploy): dynamic binary discovery for non-template projects

### LINKED ISSUE #59: deploy: hardcoded binary path doesn't match real projects

marclawclaw:
## Summary

Fixes #59.

`logos-scaffold deploy` previously hardcoded the scaffold template's binary path (`target/riscv-guest/example_program_deployment_methods/example_program_deployment_programs/riscv32im-risc0-zkvm-elf/release/`). Any project with a different name fell through to a "missing binary" error even after a successful build.

## Changes

- New `discover_program_binaries()` walks both candidate roots — `target/riscv-guest/` (canonical risc0 layout) and `methods/target/` (sub-crate workspace layout) — once per `deploy` invocation, returning a `program → path` map.
- Match criteria: filename `<program>.bin` AND a path component starting with `riscv32im` (component-level, not substring — rejects e.g. `not-riscv32im-foo/`).
- Deterministic precedence: release > debug; on ties, shallowest path wins.
- Input validation on program names (length, `/`, `\`, `..`) prevents pathological inputs from driving the walk.
- Error message now lists the directories actually searched instead of a phantom hardcoded path.
- `report.rs` uses the same discovery so diagnostic bundles reflect real binaries on renamed projects (the same hardcoded-path bug existed there).
- `--program-path` escape hatch is unchanged.

## Tests

11 unit tests in `deploy.rs` covering:
- Both layouts (`methods/target/` and `target/riscv-guest/<renamed_project>_methods/...`) — the latter is the explicit #59 regression test
- Missing search roots, no matching `.bin`, non-`riscv32im` paths
- Path traversal and overlong program-name rejection
- Release-over-debug precedence + debug-only fallback
- Substring-only `riscv32im` rejection (e.g. `not-riscv32im-foo`)
- Multi-program single-walk behaviour

Full test suite: 300 pass (200 lib + 100 cli).

## Test plan

- [x] `cargo test` (full suite)
- [x] `cargo clippy --all-targets`
- [ ] Manual: `lgs new my_app && cd my_app && lgs build && lgs deploy` resolves binaries under `target/riscv-guest/my_app_methods/...`

### DIFF
--- a/ADR.md
+++ b/ADR.md
@@ -147,6 +147,9 @@ construction.
 
 The deploy command must work for any scaffolded project regardless of its name.
 Binary discovery should derive paths from the actual project structure, not assume the template's naming.
-Use a two-tier strategy: walk `methods/target/` for program-name-matched `.bin` files in `riscv32im` paths (primary),
-with the hardcoded template path as a backward-compatible fallback.
-This keeps the scaffold general and avoids coupling to specific build-tool configuration formats.
+The implementation walks both `target/riscv-guest/` (the canonical risc0 layout used by the
+scaffold template) and `methods/target/` (the sub-crate workspace layout), matching `<program>.bin`
+files whose path components include both a `riscv32im*` target triple and a `release` directory.
+Release builds are preferred; if only a debug build exists, that is used as a fallback. When
+multiple matches exist, the shallowest path wins. This keeps the scaffold general and avoids
+coupling to a specific project name or workspace layout.
--- a/ADR.md
+++ b/ADR.md
@@ -149,4 +149,4 @@ The deploy command must work for any scaffolded project regardless of its name.
 Binary discovery should derive paths from the actual project structure, not assume the template's naming.
 Use a two-tier strategy: walk `methods/target/` for program-name-matched `.bin` files in `riscv32im` paths (primary),
 with the hardcoded template path as a backward-compatible fallback.
-This keeps the scaffold general and avoids coupling to spel configuration formats.
+This keeps the scaffold general and avoids coupling to specific build-tool configuration formats.
--- a/ADR.md
+++ b/ADR.md
@@ -30,7 +30,7 @@ Use native Cargo-based build flow as the primary compilation path.
 Developers need explicit, editable environment targeting for local and DevNet workflows.
 Use environment-file based network configuration as the default model.
 Generated projects include env files for local and DevNet,
-wallet interaction settings used by deploy and interact commands.
+wallet interaction settings used by deploy and wallet-based interaction commands.
 Env files are familiar and automation-friendly,
 but require strict handling to avoid credential leakage.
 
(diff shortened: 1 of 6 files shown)


marclawclaw (commented):
## Code Review — PR #82: fix(deploy): dynamic binary discovery

**Reviewer**: Claude (automated five-axis review)
**Scope**: deploy.rs `find_binary_path()`, ADR/FURPS updates, dependency addition

---

### ✅ Correctness

The two-tier strategy (walk → fallback) is sound and solves #59 cleanly. The fallback preserves backward compatibility for unmodified template projects. Tests cover the four key scenarios well.

**One concern — non-deterministic match selection (medium severity):**

`find_binary_path` returns the **first** `WalkDir` match, but walkdir iteration order is filesystem-dependent and not guaranteed. If `methods/target/` contains both `debug/` and `release/` builds under `riscv32im` paths, the function may return the debug binary on some runs and the release binary on others.

**Suggestion**: Prefer paths containing `/release/` over `/debug/`, or sort candidates and pick the most specific match. At minimum, document that the first-found wins. A fifth test case for this scenario would be valuable.

### ✅ Readability

Clean and well-structured. The function is small, focused, and easy to follow. Test names clearly describe what they verify. No concerns here.

### ✅ Architecture

- ADR "Build Output Discovery" section correctly documents the two-tier strategy and rationale. Consistent with the existing ADR style.
- FURPS functionality item (4) accurately captures the user-facing behavior change.
- `walkdir = "2"` is a well-maintained, minimal dependency — appropriate for recursive directory walking.

### ✅ Security

No security concerns. Program names originate from `discover_deployable_programs()` which reads `methods/guest/src/bin/*.rs` file stems — not arbitrary user input. No path traversal risk in the walk since it's bounded to `methods/target/`.

### ✅ Performance

`WalkDir` on a build `target/` directory is fine for a CLI tool. The early-return on first match avoids unnecessary traversal. No concern.

---

### Findings Summary

| # | Severity | File | Finding |
|---|----------|------|---------|
| 1 | **Medium** | `src/commands/deploy.rs` | `find_binary_path` returns first walkdir match — non-deterministic when both debug and release builds exist. Should prefer `release/` or define explicit precedence. |
| 2 | **Low** | `ADR.md` | Typo on last line: "avoids a **spel** dependency" — unclear what this means. Did you mean "SPEL" (some specific tool), or is this a typo for something else? |
| 3 | **Nit** | `src/commands/deploy.rs` | `path.to_string_lossy().contains("riscv32im")` is a loose substring match. Would also match hypothetical paths like `not-riscv32im-foo/`. Consider matching on a path component instead (e.g., checking that some ancestor directory name starts with `riscv32im`). Low risk in practice but worth noting. |

### Verdict

**Approve with suggestions.** The core change is correct and well-tested. Finding #1 (debug vs release ambiguity) is the only one worth addressing before merge — the rest are minor.

fryorcraken (approved):

PR #86, opened 2026-04-27
feat: improve CLI design for agents

weboko:
### Problem
Agents and automation struggled with discoverable `--help`, noisy echoed commands, and silently destructive defaults: `localnet reset` wiped chain state on a typo, `basecamp launch` scrubbed profile data without confirmation, `wallet topup` returned exit 0 on confirmation timeout, `build idl`/`client` printed `Skipping…` and exited 0 on framework mismatch, and `init`'s in-place migration could clobber a hand-curated `scaffold.toml.bak`.

### Description
Added copy-paste Examples (and env notes) across high-traffic commands via `cli_help` + Clap `after_long_help`; global `-q`/`--quiet` and `LOGOS_SCAFFOLD_QUIET` to suppress `$` command echo (composing anywhere on the line for both `wallet` and `spel` passthrough); `--dry-run` for `localnet reset`, `basecamp launch`, and `init` (previewing both fresh-init and migration); destructive defaults now gate behind `--yes` (`localnet reset`, `basecamp launch`); `wallet topup` exits non-zero on confirmation timeout with `status: pending`; `build idl`/`client` bail loudly with the actual `framework.kind` instead of silently no-op'ing; `init` writes `scaffold.toml.bak` before migrating and refuses to overwrite an existing one (`--no-backup` opt-out); richer deploy/wallet help for `--json` and `LOGOS_SCAFFOLD_WALLET_PASSWORD`; wallet passthrough discovery (`wallet -- --help`); README CLI lines aligned with the above.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,0 +1,43 @@
+<!--
+Thanks for opening a PR. Please fill every section below.
+See CONTRIBUTING.md for the full policy.
+
+PRs missing the "Your Project or User Demand" or "User Story" sections
+may be closed without further explanation.
+-->
+
+## Your Project or User Demand
+
+<!--
+Provide ONE of:
+- Link or short description of the project you are building with `logos-scaffold` on the Logos stack.
+- Evidence of user demand: linked issue/discussion, message from someone building on the stack, or similar concrete signal.
+
+Generic "this would be nice" does not count.
+-->
+
+## User Story
+
+<!--
+- What friction did you (or the user you represent) hit?
+- Which `logos-scaffold` command or DOGFOODING scenario (D1–L4) surfaced it?
+- What did you try before this fix?
+-->
+
+## Change Summary
+
+<!-- What this PR does, in 1–3 sentences. -->
+
+## Verification
+
+<!--
+- Which DOGFOODING scenarios did you rerun? List the IDs (D1, D2, L1, …).
+- Tests added or updated.
+-->
+
+## Checklist
+
+- [ ] CI is green
+- [ ] Relevant DOGFOODING scenarios rerun (listed above)
+- [ ] Docs updated (`README.md`, `DOGFOODING.md`) if user-facing behavior changed
+- [ ] I have read [CONTRIBUTING.md](/CONTRIBUTING.md) and this PR fits within the weekly contribution cap
--- a/.github/rate-limit-allowlist.txt
+++ b/.github/rate-limit-allowlist.txt
@@ -1,0 +1,11 @@
+# GitHub usernames exempt from the PR rate limit in .github/workflows/pr-rate-limit.yml.
+# Public members of the logos-co GitHub org are auto-exempted and do not need to be listed here.
+# Use this file for private org members or trusted external contributors.
+# One username per line. Lines starting with # are ignored.
+marclawclaw
+fryorcraken
+danisharora099
+weboko
+jimmy-claw
+jzaki
+vpavlink
--- a/.github/rate-limit-allowlist.txt
+++ b/.github/rate-limit-allowlist.txt
@@ -9,3 +9,4 @@ weboko
 jimmy-claw
 jzaki
 vpavlink
+wozos
--- a/.github/workflows/pr-rate-limit.yml
+++ b/.github/workflows/pr-rate-limit.yml
@@ -1,0 +1,126 @@
+name: PR Rate Limit
+
+on:
+  pull_request_target:
+    types: [opened, reopened]
+
+permissions:
+  pull-requests: write
+  contents: read
+
+jobs:
+  enforce:
+    runs-on: ubuntu-latest
+    timeout-minutes: 2
+    steps:
+      - name: Checkout default branch
+        uses: actions/checkout@v4
+        with:
+          ref: ${{ github.event.repository.default_branch }}
+
+      - name: Enforce rate limit
+        uses: actions/github-script@v7
+        with:
+          script: |
+            const fs = require('fs');
+            const path = require('path');
+
+            const org = 'logos-co';
+            const threshold = 3;
+            const windowDays = 7;
+
+            const pr = context.payload.pull_request;
+            const login = pr.user.login;
+            const prNumber = pr.number;
+            const { owner, repo } = context.repo;
+
+            core.info(`Checking PR #${prNumber} by ${login}`);
+
+            // Exemption: bots (dependabot, renovate, etc.) — automated PR floods
+            // are expected and should not be rate-limited like human contributors.
+            if (pr.user.type === 'Bot') {
+              core.info(`${login} is a bot — exempt.`);
+              return;
+            }
+
+            // Exemption A: public logos-co org member.
+            try {
+              await github.rest.orgs.checkPublicMembershipForUser({ org, username: login });
+              core.info(`${login} is a public member of ${org} — exempt.`);
+              return;
+            } catch (err) {
+              if (err.status === 404) {
+                // Not a public member; continue to file allowlist check.
+              } else {
+                // Fail open on transient API errors rather than incorrectly
+                // rate-limiting a legitimate org member during GitHub flakes.
+                core.warning(`Org membership check failed with status ${err.status}; skipping rate limit enforcement for this PR.`);
+                return;
+              }
+            }
+
+            // Exemption B: file allowlist.
+            const allowlistPath = path.join('.github', 'rate-limit-allowlist.txt');
+            if (fs.existsSync(allowlistPath)) {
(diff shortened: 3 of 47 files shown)


fryorcraken (dismissed):
Useful, low-risk DX improvements (Examples blocks, `-q/--quiet`, `localnet reset --dry-run`, wallet passthrough discovery). Logic is sound and tests cover the new surface. **One blocker: `cargo fmt --check` fails — CI will reject this as-is.**

## Blocker

`cargo fmt --check` fails in `src/cli.rs` at four spots — CI runs this on `ubuntu-latest`:

1. `src/cli.rs:19-25` — `use crate::cli_help::{...}` block must come before the `crate::commands::*` imports (alphabetical ordering).
2. `src/cli.rs:244` — single-line `#[arg(long, help = \"...\")]` exceeds line width and needs multi-line form.
3. `src/cli.rs:432-435` — closure body should collapse onto a single line.

Fix: `cargo fmt`.

## Correctness / behavior

- `leading_global_flags_end` (`cli.rs:442`) only consumes `-q`/`--quiet` before the wallet-passthrough check. Working as intended; `wallet --` without leading flags still hits the passthrough at `passthrough_start = 1`. ✔
- `cmd_localnet_reset_dry_run` (`localnet.rs:484`) is purely read-only: only `path.exists()`, `read_localnet_state(...).unwrap_or_default()`, `pid_running`, `port_open`. No mutation paths. The `unwrap_or_default()` silently masks a corrupt state file — fine for dry-run, just noting.
- Dry-run plan mirrors the real-reset order (binary precheck → stop → delete DB → optional wallet delete → state delete → restart → verify). ✔ The \"warning: missing sequencer binary\" preview in dry-run is a nice touch.
- `apply_quiet_from_env` (`cli.rs:431`): accepts `1`/`true`/`yes` case-sensitively, but `on` case-insensitively. Inconsistent — either match all case-insensitively or none. Minor.
- The leading-flag pre-disable of echo at `cli.rs:314` runs *before* clap parses, so an invalid command (e.g. `-q bogus`) leaves the static set to false. Harmless because the process exits immediately, just noting.

## Tests

All 81 integration tests pass, including the six new ones. Coverage of `--help` content, dry-run non-mutation, and quiet+passthrough composition is solid. **Gap:** no test for the `LOGOS_SCAFFOLD_QUIET` env var path in `apply_quiet_from_env`. Worth adding given the env-var matching has the case-sensitivity quirk above.

## Style / cosmetic

- Commit headline typo: `\"improve CLI gesign for agents\"` → `design`. Worth a squash-rename on merge.
- The \"Run inside a project directory…\" hint at `cli_help.rs:28` is rendered as part of the Examples block but isn't an example. Minor — fine to leave.
- The `EXAMPLES_*` constants are duplicated between root, parent (`wallet`), and leaf (`wallet topup`) sections. Acceptable for clap's `after_long_help` model since each leaf renders independently.

## Architecture

- New `cli_help` module is the right call — keeps `cli.rs` from becoming a wall of raw string literals. ✔
- `LocalnetAction::Reset` gaining a `dry_run` field is the minimal change — no premature abstraction. ✔
- Wallet passthrough sniffing happens before clap parsing, which is the only way to make `wallet -- --help` work without clap intercepting `--help`. Correct approach.

## Recommended actions before merge

1. **Run `cargo fmt`** to unblock CI.
2. (Optional) Fix commit message typo `gesign` → `design`.
3. (Optional) Add an integration test for `LOGOS_SCAFFOLD_QUIET=1` and normalize its case-sensitivity logic.

fryorcraken:
Re-reviewed against `5a33dd5`. Prior fmt blocker is resolved (CI green, all 81 tests pass locally). Verified the new surface end-to-end against a real LEZ project on this machine.

## 🔴 Important — must address before merge

### 1. Merge conflict with master
`mergeStateStatus: DIRTY`. PR is 5 commits behind master and `git merge-tree origin/master HEAD` reports a real content conflict in `src/cli.rs`. PR #75 added a `Basecamp` variant inside the same `crate::commands::*` import region this PR reorders to introduce `cli_help`. Rebase onto master and redo the import block before merging.

### 2. `LOGOS_SCAFFOLD_QUIET` env-var matching is asymmetric
`src/cli.rs:434-440`:

```rust
.map(|v| matches!(v.as_str(), "1" | "true" | "yes") || v.eq_ignore_ascii_case("on"))
```

`1`, `true`, `yes` are case-sensitive; only `on` is case-insensitive. Verified empirically:

| `LOGOS_SCAFFOLD_QUIET=` | Echo suppressed? |
|---|---|
| `1`, `true`, `yes`, `on` | ✓ |
| `TRUE`, `Yes` | ✗ (still echoes) |
| `ON` | ✓ |

The help string in `cli_help.rs:12` advertises all four values without flagging case sensitivity. Pick one direction: either `eq_ignore_ascii_case` for all four, or document the case-sensitive behavior. I'd recommend going fully case-insensitive — surprise factor wins. (This was raised in the prior review and not addressed by the `fix ci` commit.)

## 🟡 Suggestions

### 3. Dry-run mixes absolute and relative paths
`src/commands/localnet.rs:495-541`. Live output:

```
planned: stop sequencer if tracked (state file: /home/.../.scaffold/state/localnet.state)
planned: delete sequencer DB at .scaffold/cache/repos/lssa/rocksdb (exists: true)
prerequisite ok: sequencer binary exists at .scaffold/cache/repos/lssa/target/release/sequencer_service
```

`state_path` and wallet paths render absolute (joined onto `project.root`); `lez`-derived paths (`rocksdb_path`, `sequencer_bin`) stay relative because `scaffold.cache_root` is relative in `scaffold.toml`. Inconsistent for users copy-pasting paths into shell. Fix: render `project.root.join(rocksdb_path).display()` (or canonicalize once at the top of the dry-run).

### 4. Dry-run silently masks the foreign-listener case
Same function. When a foreign listener occupies the port, `cmd_localnet_reset` aborts at `wait_for_port_free` with `ResetError::ForeignListener` **before** any cleanup. The dry-run prints

```
planned: start sequencer (...)
note: listener currently on 127.0.0.1:3040 (port 3040)
```

which is misleading — the real run would never reach the start step. Worth promoting that `note:` to a `warning: real reset would abort with foreign-listener error` when `state.sequencer_pid` is `None` **and** the port is open.

### 5. No integration test for the `LOGOS_SCAFFOLD_QUIET=` env path
`tests/cli.rs:37` only asserts the help string mentions the variable name. Given finding #2, an explicit env-var test would have caught the case-sensitivity asymmetry. Cheap to add.

### 6. Commit headline typo
`feat: improve CLI gesign for agents` → `design`. Squash-rename on merge.

## Five-axis breakdown

| Axis | Verdict |
|---|---|
| **Correctness** | Dry-run is purely read-only (only `path.exists()`, `read_localnet_state(...).unwrap_or_default()`, `pid_running`, `port_open`). Plan order matches real reset. Wallet-passthrough sniffing in `cli.rs:451-469` correctly composes with leading `-q` via `leading_global_flags_end`. Issues: env-var case (#2), path display (#3). |
| **Readability** | `cli_help` extraction is a clean win — keeps `cli.rs` from drowning in raw strings. Reset dry-run is a long sequence of `println!` but stays linear and obvious. Minor: the two-step env matcher (`matches!` + `eq_ignore_ascii_case`) reads like a half-finished refactor (#2). |
| **Architecture** | `LocalnetAction::Reset` gains a `dry_run` field — minimal change, no premature abstraction. Pre-clap passthrough sniffing (`run` in `cli.rs:314-322`) is the only correct approach for `wallet -- --help` (otherwise clap eats `--help`). `EXAMPLES_*` duplicated across root / parent / leaf is fine for clap's `after_long_help` model. |
| **Security** | `wallet_passthrough_action` forwards args via `Command` (no shell). `LOGOS_SCAFFOLD_WALLET_PASSWORD` continues via stdin and the help explicitly says "do not log this value". Dry-run is read-only (the new `localnet_reset_dry_run_prints_plan_without_mutations` test asserts rocksdb is not created). No new auth surfaces. |
| **Performance** | Help blocks are `&'static str` (zero runtime cost). No new loops, no unbounded allocations, no N+1 patterns. |

## Recommended actions before merge

1. **Rebase onto master** to clear the `cli.rs` import-block conflict introduced by #75.
2. **Normalize `LOGOS_SCAFFOLD_QUIET` parsing** (`cli.rs:434-440`) — make all four values case-insensitive.
3. (Optional) Render `lez`-derived paths absolute in dry-run output (`localnet.rs:500-541`).
4. (Optional) Promote the foreign-listener note to a "real-run would abort here" warning.
5. (Optional) Add an env-var integration test for `LOGOS_SCAFFOLD_QUIET`.
6. (Optional) Fix `gesign` → `design` at squash time.

fryorcraken (approved):

danisharora099 (approved):

PR #90, opened 2026-04-30
fix: auto-detect r0vm path from rzup installation for localnet

### LINKED ISSUE #88: First-success path: wallet topup crashes localnet on auth-transfer init
## Summary

Running the documented `logos-scaffold` first-success path on macOS reaches `lgs wallet topup`, then the standalone sequencer exits during `wallet auth-transfer init`.

This is separate from existing issue #69. I applied the #69 workaround locally so the generated smoke project could build before testing topup/deploy.

## Environment

- macOS Apple Silicon
- `logos-scaffold v0.1.1` installed from `https://github.com/logos-co/logos-scaffold#c3313fd8`
- Scaffold pinned LEZ: `35d8df0d031315219f94d1546ceb862b0e5b208f`
- Rust: `rustc 1.94.0`
- RISC Zero Rust toolchain installed via `rzup install rust`
- Command run with `RISC0_DEV_MODE=1` after a fresh localnet reset

## Reproduction

```bash
cargo install --git https://github.com/logos-co/logos-scaffold
lgs new lp0003-stack-smoke
cd lp0003-stack-smoke
lgs setup
export PATH="$HOME/.risc0/bin:$PATH"
rzup install rust
rm -rf rocksdb bedrock_signing_key
RISC0_DEV_MODE=1 lgs localnet start
RISC0_DEV_MODE=1 lgs wallet topup
```

## Observed output

`lgs wallet topup` starts preflight and runs auth-transfer init:

```text
wallet topup preflight: destination is uninitialized; running auth-transfer init
$ .../target/release/wallet auth-transfer init --account-id Public/CbgR6tj5kWx5oziiFptM7jMvrQeYY3Mzaao6ciuhSr2r
```

The command then hangs because the sequencer exits. `lgs localnet logs --tail 120` shows:

```text
[2026-04-29T08:00:57Z INFO  sequencer_service] RPC server started
[2026-04-29T08:00:57Z INFO  sequencer_service] Starting main sequencer loop
[2026-04-29T08:01:12Z INFO  sequencer_service] Collecting transactions from mempool, block creation
[2026-04-29T08:01:12Z ERROR sequencer_core] Transaction with hash a233ad90304d9578f560d9e7beb54aeec8fa54cbb2f3bf25038eec37441e4a4b failed execution check with error: ProgramExecutionFailed(
        "No such file or directory (os error 2)",
    ), skipping it
[2026-04-29T08:01:12Z ERROR sequencer_service] Sequencer failed unexpectedly: Main loop exited unexpectedly
[2026-04-29T08:01:12Z INFO  sequencer_service] Sequencer shutdown complete
```

After that, `lgs localnet status` reports stale state and port 3040 not reachable.

## Expected

The first-success path should complete `lgs wallet topup` or fail with an actionable diagnostic that identifies the missing file path.

## Notes

The expected local files appear to exist on this machine:

```text
~/.logos-blockchain-circuits/{poc,poq,pol,zksign}/witness_generator
~/.logos-blockchain-circuits/prover
~/.logos-blockchain-circuits/verifier
.../logos-execution-zone-35d8df0/artifacts/program_methods/authenticated_transfer.bin
```

The sequencer log currently reports only `No such file or directory`, without the missing path, which makes the failure hard to diagnose from the scaffold runbook.

ygd58:
Fixes #88.

## Problem

On macOS, rzup installs r0vm under `~/.risc0/extensions/<version>/r0vm` but does not add it to PATH. When `logos-scaffold localnet start` spawns the sequencer, it inherits the parent process PATH without r0vm. Any transaction requiring RISC Zero execution fails with:

```
ProgramExecutionFailed("No such file or directory (os error 2)")
```

Root cause identified by @bitfalt in #88: setting `RISC0_SERVER_PATH` explicitly to the rzup-installed r0vm path resolves the issue.

## Fix

`logos-scaffold localnet start` now:
1. Checks if `RISC0_SERVER_PATH` is already set (respects user override)
2. If not set, searches `~/.risc0/extensions/*/r0vm` for rzup-installed binary
3. Sets `RISC0_SERVER_PATH` automatically if found

## Testing

On Linux r0vm is typically on PATH so `RISC0_SERVER_PATH` is not needed — this change is a no-op when the env var is already set or r0vm is not found under `~/.risc0`.


fryorcraken:
Thanks for tackling this. Before merging, I'd like us to step back on the design.
                                                                                                                                                              
  The current approach picks an r0vm by walking ~/.risc0/extensions/ and taking whatever read_dir returns first. That's non-deterministic across machines, and
  more importantly it's a guess: any r0vm we happen to find may not be the version the sequencer was actually built against. On a machine with multiple rzup
  versions installed, we can silently hand the sequencer the wrong one and trade an ENOENT for a much more confusing version-mismatch failure later.
                                                                                                                                                           
  The principle I'd like the fix to follow: the sequencer should run against the exact r0vm version it was built against, or fail loudly. Auto-detection is  
  fine, but it should be derived from something authoritative (the lez build's risc0 pin, an rzup-managed "active" version, etc.) rather than from filesystem
  enumeration order. If we can't determine the right version with confidence, we should error with a clear message asking the user to set RISC0_SERVER_PATH,
  not pick arbitrarily.
                                                                                                                                                          
  Happy to discuss approaches, but I'd rather not merge the current heuristic and iterate on top of it. Could you take another pass with that principle in
  mind?

ygd58:
Updated the approach based on your feedback.

Instead of walking extensions/ and taking the first match, the fix now:
1. Reads the exact risc0-zkvm version from the LEZ Cargo.lock
2. Constructs the exact rzup path: `~/.risc0/extensions/v<version>-cargo-risczero-<arch>-<os>/r0vm`
3. If not found, emits a clear warning asking the user to set RISC0_SERVER_PATH explicitly

No guessing — either the exact version matches or we fail loudly.

ygd58:
Actualized against current master. Clean merge, no conflicts. All 297 lib tests pass. @weboko @fryorcraken

weboko:
there was another merge, can you update please? or I will

ygd58:
Updated against latest master. Integrated r0vm auto-detect with upstream's patched_config_path refactor. All non-pre-existing tests pass. @weboko

weboko (approved):
good PR, thanks

PR #95, opened 2026-05-04
feat(config): scaffold.toml schema v0.2.0 — portable + consolidated

fryorcraken:
## Summary

Two phases of the same change ship together: scaffold.toml is now portable across machines and CI, and the file's section/field shape is consolidated into three orthogonal namespaces. Schema bumps to `[scaffold].version = "0.2.0"` with a single migration path through `lgs init`.

### 1. Portable repo paths

- Cache-managed projects no longer write absolute `[repos.lez].path` / `[repos.spel].path` into `scaffold.toml`. The on-disk clone location is derived at runtime from `<cache_root>/repos/<name>/<pin>`.
- `--vendor-deps` projects still persist relative `.scaffold/repos/{lez,spel}` literals.
- Existing files with a non-empty `path` are honored verbatim — no forced migration on that field.
- `setup` decides sync mode from the config (empty path → auto-reclone; non-empty → fail-on-mismatch), replacing the path-prefix heuristic.

### 2. Schema consolidation (v0.2.0)

Three orthogonal namespaces replace the pre-0.2.0 mixed shape:

- **`[repos.<name>]`** — every pinned external git dep (`lez`, `spel`, `basecamp`, `lgpm`), one schema: `source`, `pin`, optional `build` (`"cargo"` default, `"nix-flake"` for nix-built deps), optional `attr`, optional `path`. The redundant `url` field is dropped — it was always equal to `source` in practice.
- **`[modules.<name>]`** — Logos modules the project ships. Lifted from `[basecamp.modules.*]` because modules are project-level artifacts; basecamp is one consumer, not their owner.
- **`[basecamp]`** — only `port_base` and `port_stride` (runtime config); pin/source moved to `[repos.basecamp]`, `lgpm_flake` split into `[repos.lgpm]`.

### 3. Migration

`lgs init` is the single migration entry point. Detects any pre-0.2.0 signal (version stamp, `url` field, `[basecamp].pin`/`.source`/`.lgpm_flake`, `[basecamp.modules.*]`) and rewrites in place via `toml_edit` so user comments, key ordering, and unrelated sections survive. Pre-0.2.0 lgpm flake refs of the form `github:owner/repo/<sha>#attr` are split into `source`/`pin`/`attr` automatically; unparseable refs get a hand-edit hint with a default pin written in place so the file is at least valid.

All other commands hard-fail on pre-0.2.0 files with a single line pointing at `init` — same posture as the original `[repos.spel]` migration. JSON consumers stay clean.

### 4. Implementation notes

- `src/config.rs` rewrite: hand-rolled TOML parser/serializer replaced with `toml_edit`. Comment/ordering preservation is what made the migration path tractable.
- New `RepoBuild` enum (Cargo default, NixFlake) drives setup-time build dispatch. Flake refs for nix-flake builds are reconstructed at use time as `<source>/<pin>#<attr>`.
- ADR's "Pinned Dependencies" section rewritten to describe the realized three-namespace schema and the build-dispatch story.

## Test plan

- [x] `cargo fmt --check`
- [x] `cargo check`
- [x] `cargo test --all-targets` — 221 lib + 116 integration pass
- [x] Smoke: `lgs init` in a tmpdir produces v0.2.0 schema with no `url` fields, no `[basecamp].pin`/`.source`/`.lgpm_flake`, no `[basecamp.modules.*]`
- [x] Smoke: `lgs init` against a hand-crafted pre-0.2.0 fixture migrates correctly, preserves user comments, splits `lgpm_flake`, moves modules to top level
- [ ] Reviewer should validate the back-compat path: hand-craft a v0.2.0 `scaffold.toml` with absolute `[repos.lez].path` pointing at an existing clone, confirm `setup` honors the literal path

## Future Work

Once lez sequencer is available as a module, it may make sense to move spel and scaffold to nix, so that nix can be used for pinning instead of hand rolling a pinning system.

`lgs basecamp modules` should then probably become a root command `lgs modules`.


src/config.rs
```diff
@@ -1,377 +1,442 @@
-use anyhow::bail;
+//! Parser, serializer, and migration helpers for `scaffold.toml`.
```
fryorcraken: Can we move migration helper to its own module?

src/migrate.rs
```diff
@@ -0,0 +1,326 @@
...
+                current_version.as_str()
+            },
+            SCAFFOLD_TOML_SCHEMA_VERSION,
+        ));
+    }
+
+    // [repos.lssa] -> [repos.lez] alias rename. Only triggered if the
+    // explicit lssa section exists; harmless otherwise.
+    if let Some(repos) = doc.get_mut("repos").and_then(Item::as_table_mut) {
+        if repos.contains_key("lssa") && !repos.contains_key("lez") {
```
danisharora099: we should split this `if` condition to ensure that `lssa` doesn't exist even if `lez` exists.

`[repos.lssa]` is only removed when `repos.lez` is absent. it is uncommon but would be a nice safe-guard

danisharora099 (approved):
LGTM, one tiny comment

src/migrate.rs
fryorcraken: Good catch — split as suggested in b5330b4. When both sections exist, `lez` wins and the stale `lssa` is dropped (otherwise the post-migration file would still fail `detect_old_schema`). Added a test for the both-present case.

PR #96, opened 2026-05-05
feat: add `logos-scaffold run` command (MVP)

fryorcraken:
Introduces the `logos-scaffold run` (alias `lgs run`) command that drives the standard project lifecycle in one shot: build → build IDL → ensure localnet → wallet topup → deploy → optional post-deploy hook. Configuration surface for this slice is the minimal inline `[run].post_deploy` (string or array). CLI flags: `--post-deploy <CMD>` (repeatable, replaces config) and `--no-post-deploy` (skip even if configured). Hook env contract is the single-program shortcut (`SEQUENCER_URL`, `NSSA_WALLET_HOME_DIR`, `SCAFFOLD_PROJECT_ROOT`, `SCAFFOLD_IDL_DIR`, `SCAFFOLD_PROGRAM_ID`, `SCAFFOLD_GUEST_BIN`).

## Stack

This is part 1 of a 5-PR stack that splits PR #66:

1. **feat/run-mvp ← this PR** — pipeline core, post-deploy hook, single-program env
2. feat/run-deploy-cache (depends on 1) — skip deploy when guest .bin + IDL hashes are unchanged
3. feat/run-multiprogram-env (depends on 2) — `SCAFFOLD_PROGRAMS`, `SCAFFOLD_PROGRAM_ID_<n>`, …
4. feat/run-reset-and-profiles (depends on 3) — `--reset` wipe-and-reseed and named `[run.profiles.*]`
5. feat/run-watch (depends on 4) — `--watch` re-runs pipeline on file changes

Each PR's base is the previous PR's branch; merge in order.

## Replaces part of

#66 — PR #66 will be retired by editing its description once this stack lands.

## Test plan

- [x] `cargo fmt --check`
- [x] `cargo check --all-targets`
- [x] `cargo test --all-targets` (passes locally; new tests: `run_help_lists_command_summary`, `run_outside_project_fails_with_project_scoped_message`, `run_rejects_both_post_deploy_flags`, `run_with_post_deploy_hook_shows_6_steps_in_output`, `run_with_multiple_post_deploy_hooks_uses_array_form`, `run_fails_at_build_step_in_mock_project`, plus 10 unit tests in `commands::run`)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Co-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>

PR #66, opened 2026-04-14
feat: add `logos-scaffold run` command

fryorcraken:
> **Superseded by stacked split — please review #96–#100 instead.**

This PR was a single 2535-LOC squashed commit bundling 7 distinct features. Per review feedback, it has been split into 5 stacked draft PRs that land each feature as a coherent, independently reviewable slice. The branches in the stack each compile and test cleanly on their own.

## Replacement stack (review in order)

| # | PR | Branch | Base | Adds |
|---|---|---|---|---|
| 1 | #96 | `feat/run-mvp` | `feat/portable-scaffold-toml` | Pipeline core (build → idl → localnet → topup → deploy → hook), `TopupOutcome` enum, single-program hook env contract, `--post-deploy` / `--no-post-deploy` overrides, docs (ADR, FURPS, DOGFOODING, README) |
| 2 | #97 | `feat/run-deploy-cache` | `feat/run-mvp` | Deploy idempotence cache (`run_deploy.json`, SHA-256 of guest bin + IDL), `sha2` dep, `SCAFFOLD_DEPLOY_SKIPPED` env var |
| 3 | #98 | `feat/run-multiprogram-env` | `feat/run-deploy-cache` | Multi-program hook env fan-out (`SCAFFOLD_PROGRAMS`, `SCAFFOLD_PROGRAM_ID_<n>`, `SCAFFOLD_GUEST_BIN_<n>`, `SCAFFOLD_DEPLOY_SKIPPED_<n>`), env-var name sanitization + warning |
| 4 | #99 | `feat/run-reset-and-profiles` | `feat/run-multiprogram-env` | `--reset` / `--no-reset` + `[run].reset` (full localnet wipe + wallet reseed), named profiles `[run.profiles.<name>]`, `--profile` selector, `default_profile` fallback |
| 5 | #100 | `feat/run-watch` | `feat/run-reset-and-profiles` | `--watch` flag, debounced re-run on filesystem changes, `notify` dep |

## Why split

Each follow-up is independently reviewable: cache logic in isolation (#97), multi-program env contract separate from single-program MVP (#98), reset semantics + config-shape change together (#99), and watch loop as an opt-in dev-loop addition (#100). The MVP (#96) ships a usable `lgs run` on its own.

Closing this PR once #96 lands. Original branch `feat/run-command` is preserved at `backup/feat-run-command-pre-rebase` and `backup/feat-run-command-pre-split`.

PR #97, opened 2026-05-05
feat(run): cache deploy by guest binary + IDL hash

fryorcraken:
Adds an idempotence layer to the run pipeline's deploy step. Before invoking `cmd_deploy`, hash every guest `.bin` together with its matching IDL JSON and a small deploy-config digest (sequencer port, wallet home, IDL path), then compare against `.scaffold/state/run_deploy.json` from the prior run. Skip when **all** of (a) program hashes match, and (b) the cached sequencer PID matches the currently-running sequencer. Deploy and update the cache otherwise.

The PID gate is what closes the stale-deploy trap: `lgs localnet stop && lgs localnet start` wipes on-chain state but leaves the cache file intact, so without PID equality the next `lgs run` would skip deploy and silently leave the new chain empty. The config digest closes the analogous trap for `localnet.port` / `wallet_home_dir` / `framework.idl.path` edits without `--reset`. Folding IDL JSON into the program hash means an ABI-only edit forces a re-deploy even when the compiled binary is byte-identical; for `lez-framework` projects, a missing IDL file at hash time bails loudly rather than caching a partial digest.

The cache file is written via tempfile + rename in the same directory, so a SIGINT mid-write leaves the prior cache intact rather than truncating it. Malformed cache files (e.g. a truncated `run_deploy.json` from a crash) are recovered as cache-miss with a `warning:` line naming the path and the parse error, instead of silently re-deploying every run forever. Legacy state files without `localnet_pid` (written by an earlier iteration of this branch) parse via `#[serde(default)]` and are treated as unknown — forcing a re-deploy rather than trusting them.

Post-deploy hooks now also receive `SCAFFOLD_DEPLOY_SKIPPED=0|1` in the single-program shortcut so consumers can short-circuit redundant on-chain work.

## Env vars exposed to post-deploy hooks at this point of the stack

Always set:

- `SEQUENCER_URL`
- `NSSA_WALLET_HOME_DIR`
- `SCAFFOLD_PROJECT_ROOT`
- `SCAFFOLD_IDL_DIR`

Single-program shortcut (only when the project has exactly one deployable program):

- `SCAFFOLD_PROGRAM_ID` (omitted when `spel inspect` fails)
- `SCAFFOLD_GUEST_BIN`
- `SCAFFOLD_DEPLOY_SKIPPED` (new in this PR)

The multi-program env fan-out (`SCAFFOLD_PROGRAMS`, `SCAFFOLD_PROGRAM_ID_<name>`, `SCAFFOLD_PROGRAM_NAME`, etc.) lands in part 3 of the stack.

## Stack

This PR is part 2 of a 4-PR stack that splits PR #66. Part 1 of the original split (run-mvp) merged separately as #96, so the live stack is:

1. ~~feat/run-mvp~~ — landed as #96 ✅
2. **feat/run-deploy-cache ← this PR** — skip deploy when guest .bin + IDL + config hashes are unchanged and sequencer is the same instance
3. feat/run-multiprogram-env (#98, depends on this) — `SCAFFOLD_PROGRAMS`, `SCAFFOLD_PROGRAM_ID_<name>`, …
4. feat/run-reset-and-profiles (#99, depends on #98) — `--reset` wipe-and-reseed and named `[run.profiles.*]`
5. feat/run-watch (#100, depends on #99) — `--watch` re-runs pipeline on file changes

Base for this PR is `master`. Each downstream PR's base is the previous PR's branch; merge in order.

## Replaces part of

#66 — PR #66 will be retired by editing its description once this stack lands.

## Test plan

- [x] `cargo fmt --check`
- [x] `cargo check --all-targets`
- [x] `cargo test --all-targets` — 11 unit tests in `commands::run_state` covering `deploy_can_be_skipped` invariants (PID-match, PID-differs, PID-unknown-either-side, hash-differs, new-program, either-empty), `hex_encode` formatting, save→load round-trip, atomic-write tempfile cleanup, malformed-cache fallback, and legacy-cache forward-compat.
- [x] Dogfooded against a real Phase-1 project: cold run, no-change re-run (deploy skipped), source edit (re-deployed), deleted cache (re-deployed), corrupted cache (warning + re-deployed), `lgs localnet stop && start` between runs (PID mismatch → re-deployed), `SCAFFOLD_DEPLOY_SKIPPED` reaching the hook. See `docs/dogfood-pr97-2026-05-11.md` in the cryptarchs-scaffold-develop project for the writeup.

Co-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>

### DIFF
--- a/src/commands/run.rs
+++ b/src/commands/run.rs
@@ -9,6 +9,9 @@ use crate::commands::deploy::{
 };
 use crate::commands::idl::build_idl_for_current_project;
 use crate::commands::localnet::{build_localnet_status_for_project, cmd_localnet, LocalnetAction};
+use crate::commands::run_state::{
+    compute_program_hashes, deploy_can_be_skipped, load_state, save_state, RunDeployState,
+};
 use crate::commands::wallet::{cmd_wallet_topup_inner, TopupOutcome};
 use crate::constants::SPEL_BIN_REL_PATH;
 use crate::model::{LocalnetOwnership, Project};
@@ -57,9 +60,31 @@ fn run_pipeline_once(project: &Project, hooks: &[String]) -> DynResult<()> {
         bail!("wallet topup confirmation timed out; aborting run to avoid deploying with uncertain funding.\nHint: retry `logos-scaffold run` or run `logos-scaffold wallet topup` manually.");
     }
 
-    // Step 5: Deploy
-    println!("[5/{total_steps}] Deploying programs...");
-    cmd_deploy(None, None, false)?;
+    // Step 5: Deploy (idempotent: skip when guest .bin + IDL hashes both
+    // match the prior deploy. IDL is folded in so an ABI-only edit forces
+    // a re-deploy even if the binary is byte-identical). To force a
+    // re-deploy, delete `.scaffold/state/run_deploy.json` manually
+    // (a `--reset` switch arrives in a later branch of this stack).
+    let current_hashes = compute_program_hashes(project)?;
+    let prior = load_state(project);
+    let skipped = if deploy_can_be_skipped(&current_hashes, &prior.program_hashes) {
+        println!(
+            "[5/{total_steps}] Deploy skipped (guest binaries + IDL unchanged; delete `.scaffold/state/run_deploy.json` to force a re-deploy)"
+        );
+        true
+    } else {
+        println!("[5/{total_steps}] Deploying programs...");
+        cmd_deploy(None, None, false)?;
+        if !current_hashes.is_empty() {
+            save_state(
+                project,
+                &RunDeployState {
+                    program_hashes: current_hashes,
+                },
+            )?;
+        }
+        false
+    };
 
     // Step 6: Post-deploy hooks (or summary)
     if has_hooks {
@@ -67,7 +92,7 @@ fn run_pipeline_once(project: &Project, hooks: &[String]) -> DynResult<()> {
         println!("[6/{total_steps}] Running {n} post-deploy hook(s)...");
         for (i, hook) in hooks.iter().enumerate() {
             println!("===> post_deploy[{}/{n}]: {hook}", i + 1);
-            run_post_deploy_hook(project, hook)?;
+            run_post_deploy_hook(project, hook, skipped)?;
             println!("<=== post_deploy[{}/{n}] OK", i + 1);
         }
     } else {
@@ -133,7 +158,7 @@ fn print_deploy_summary(project: &Project) -> DynResult<()> {
     Ok(())
 }
 
-fn build_hook_command(project: &Project, hook_command: &str) -> Command {
+fn build_hook_command(project: &Project, hook_command: &str, deploy_skipped: bool) -> Command {
     let port = project.config.localnet.port;
     let sequencer_url = format!("http://127.0.0.1:{port}");
     let wallet_home = project
@@ -172,6 +197,10 @@ fn build_hook_command(project: &Project, hook_command: &str) -> Command {
             }
         }
         cmd.env("SCAFFOLD_GUEST_BIN", &binary_path);
+        cmd.env(
+            "SCAFFOLD_DEPLOY_SKIPPED",
+            if deploy_skipped { "1" } else { "0" },
+        );
     }
     cmd
 }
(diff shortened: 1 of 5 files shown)


src/commands/run_state.rs
```diff
@@ -0,0 +1,355 @@
...
+    let programs = discover_deployable_programs(&project.root)?;
+    if programs.is_empty() {
+        return Ok(out);
+    }
+    let binaries = discover_program_binaries(&project.root, &programs);
+    let idl_dir = project.root.join(&project.config.framework.idl.path);
+    let cfg_digest = config_digest(project);
+    let is_lez_framework = project.config.framework.kind == FRAMEWORK_KIND_LEZ_FRAMEWORK;
+
+    for (stem, bin_path) in binaries {
```
weboko, line 92: This only iterates binaries that were successfully discovered. If a new deployable `methods/guest/src/bin/foo.rs` exists but `foo.bin` is missing, `current_hashes` can still match the previous cache for the old programs and `lgs run` will skip deploy instead of surfacing the missing binary error that `cmd_deploy` would have reported. Please invalidate the cache when `binaries.len() != programs.len()` or otherwise include missing programs in the hash/comparison state.

src/commands/run_state.rs
```diff
@@ -0,0 +1,355 @@
...
+    Ok(out)
+}
+
+/// Canonical, stable digest of the deploy-affecting bits of scaffold.toml.
+/// Pinned to a small explicit set rather than serializing the whole config
+/// so unrelated config edits don't bust the cache.
+fn config_digest(project: &Project) -> String {
+    let cfg = &project.config;
+    format!(
+        "port={}|wallet={}|idl={}",
```
weboko: This digest includes `localnet.port`, but the actual deploy target comes from the wallet runtime's `sequencer_addr` when present. If `.scaffold/wallet/wallet_config.json` changes to a different sequencer while `scaffold.toml` is unchanged, the cache can still skip deploy. Please fold the resolved deploy sequencer address into this state, or make `run`/`deploy` use one canonical target source.

weboko (commented):
added some comments, otherwise looks good

src/commands/run.rs
```diff
@@ -210,6 +240,10 @@ fn build_hook_command(
             cmd.env("SCAFFOLD_PROGRAM_ID", id);
         }
         cmd.env("SCAFFOLD_GUEST_BIN", &sp.binary_path);
+        cmd.env(
```
weboko: `SCAFFOLD_DEPLOY_SKIPPED` is run-level state, but it is only exported inside the single-program shortcut block. Multi-program hooks cannot observe whether deploy was skipped. Please set this with the always-on hook env vars above so every post-deploy hook gets the same skip signal.

weboko (approved):

src/commands/run_state.rs
franckjitsu: Fixed in d06bc16. `compute_program_hashes` now bails with `guest source(s) without a compiled binary: <names>. Run `lgs build` to produce the missing `.bin` file(s) before re-running.` when `binaries.len() != programs.len()`. The user sees the error here instead of `cmd_deploy` failing later, and a cache hit can no longer skip a deploy that would have failed. Test: `compute_hashes_bails_when_a_program_is_missing_its_bin`.

src/commands/run_state.rs
franckjitsu: Fixed in d06bc16 — went with the "fold it into the digest" option. `config_digest` now resolves the sequencer address via `load_wallet_runtime` (same code path `cmd_deploy` uses), with the same `http://127.0.0.1:3040` fallback when the wallet runtime isn't loadable yet. So the cache key tracks what `cmd_deploy` actually targets — repointing `.scaffold/wallet/wallet_config.json` at a different sequencer now invalidates the cache even when scaffold.toml is unchanged. Test: `compute_hashes_includes_sequencer_addr_in_digest`. (The "single canonical target source" alternative would have been a bigger refactor of `cmd_deploy` and `cmd_localnet`; deferred as out of scope for this PR.)

src/commands/run.rs
franckjitsu: Fixed in d06bc16. Moved `SCAFFOLD_DEPLOY_SKIPPED` out of the `if let Some(single_program)` block to alongside `SEQUENCER_URL`/`NSSA_WALLET_HOME_DIR`/etc. so every post-deploy hook sees it regardless of program count. Tests: `hook_receives_deploy_skipped_env_without_single_program` and `hook_receives_deploy_skipped_zero_when_deploy_ran`. The downstream multi-program PR (#98) carries this through correctly — the per-program `SCAFFOLD_DEPLOY_SKIPPED_<name>` vars stay program-scoped, the un-suffixed one is now run-level.

PR #102, opened 2026-05-06
fix: address dogfooding-report findings (circuits/nix preflight, cache + project-dir safety, polish)

weboko:
## Summary

A dogfooding pass through DOGFOODING.md scenarios E1, E2, D1, D2, D5, L1 surfaced one **Blocker**, two **High**-severity bugs, and a handful of doc/DX polish items. This PR addresses all of them.

The intended outcome: a first-time user following only README + \`--help\` either succeeds at \`lgs new\` → \`lgs setup\` → \`lgs localnet start\`, or fails with a single, scaffold-owned, actionable error. No raw OS-level / cargo-build-script panics, no unannounced deletion of populated caches, no orphan project directories from a failed first command.

### Blocker / High

- **F1 (Blocker)** — `setup` failed deep inside cargo with a raw \`logos-blockchain-pol\` build-script panic when \`LOGOS_BLOCKCHAIN_CIRCUITS\` was unset and \`~/.logos-blockchain-circuits/\` was missing. New scaffold-side precheck in [src/commands/setup.rs](src/commands/setup.rs) bails before any cargo work; matching row in \`doctor\` ([src/doctor_checks.rs](src/doctor_checks.rs), [src/commands/doctor.rs](src/commands/doctor.rs)). README, DOGFOODING.md, and the generated project README list the prereq.
- **F2 (High)** — \`basecamp setup\` exposed a raw \`os error 2\` when \`nix\` was missing; the user had no signal the missing tool was \`nix\`. New \`ensure_nix_present()\` preflight in [src/commands/basecamp.rs](src/commands/basecamp.rs) hints at the Nix install URL. \`doctor\` shows a warn-level \`tool nix\` row. Preflight runs *after* the more-specific "not set up yet" / "no modules captured" guards so the more useful error wins.
- **F3a (High)** — A typo on \`--lez-path /nonexistent\` used to wipe the populated LEZ cache because \`reconcile_repo_source\` ran \`fs::remove_dir_all\` *before* \`git clone <new_source>\`. Now we probe the new source via \`git ls-remote\` (with prompt-suppression env vars to avoid hanging at a credential prompt) or local-path inspection first and bail with the cache intact ([src/repo.rs](src/repo.rs)). Covered by \`source_mismatch_reclone_refuses_when_new_source_unreachable\`.
- **F3b (High)** — A failure during \`new\` left a half-built project dir for the user to clean up. \`cmd_new\` now wraps its body in an inner function and best-effort removes the target on error ([src/commands/new.rs](src/commands/new.rs)). The pre-existing \`target.exists()\` guard ensures we never delete a pre-existing user directory.

### Polish / DX

- **F4** — Stripped clap's leading \`error: \` prefix before re-wrapping in anyhow, so the user sees one prefix instead of \`error: error: ...\` ([src/cli.rs](src/cli.rs)).
- **F5** — \`basecamp doctor\` empty-state row was \`PASS\` with a \`remediation\` — but \`print_rows\` only prints remediation for Warn/Fail, so the hint was hidden. Changed to \`Warn\` ([src/commands/basecamp.rs](src/commands/basecamp.rs)).
- **F6** — \`docs/basecamp-module-requirements.md\` still used the pre-v0.2.0 \`[basecamp.modules.<name>]\` table form everywhere. Synced to the canonical \`[modules.<name>]\` form to match README + FURPS + actual \`serialize_config\` output.
- **F7** — Added a hidden-dispatch \`Spel\` clap variant with \`long_about\` so \`--help\` lists \`spel\` and \`lgs spel --help\` documents the passthrough form. Added \`long_about\` to \`WalletArgs\` documenting \`wallet -- <args>\`. Special-cased \`spel --help\` in the early intercept so clap can render help. The runtime \`spel -- <args>\` passthrough is unchanged ([src/cli.rs](src/cli.rs)).
- **F8** — Suppressed scaffold's \`$ git ...\` echo in \`cmd_new\` (3 lines per first run) and replaced with one human-friendly progress line. \`git\`'s own progress output on stderr is unchanged ([src/commands/new.rs](src/commands/new.rs)).

## Out of scope

The dogfooding TL;DR also flagged "over-redaction of public flags in \`report\`." That is left for a follow-up — without a concrete false-positive list, tightening the keyword regex risks regressing real secret detection.

## Test plan

- [x] \`cargo build\` clean
- [x] \`cargo test --lib\` — 223 pass; one pre-existing flaky \`wallet_support\` test fails under parallelism but passes in isolation
- [x] \`cargo test --test cli\` — 119 pass; 6 failures, all pre-existing or flaky (5 fail on master pre-changes; 1 flaky under parallelism)
- [x] New unit test \`source_mismatch_reclone_refuses_when_new_source_unreachable\` passes
- [x] Smoke-test on the built binary:
  - \`logos-scaffold --help\` lists \`spel\` and shows the wallet passthrough form in \`wallet --help\`
  - \`logos-scaffold spel --help\` reaches clap (was previously short-circuited by the early intercept)
  - \`logos-scaffold bogus-subcommand\` → single \`error:\` prefix
  - \`logos-scaffold new bad --lez-path /nonexistent\` → bails with a clear error referencing the cache; cache directory is preserved; the partially-created project directory is cleaned up
  - First run of \`new\` shows \`Cloning lez at pin ...\` instead of \`$ git clone ...\` / \`$ git fetch ...\` / \`$ git checkout ...\`


src/commands/setup.rs
```diff
@@ -14,7 +17,12 @@ use super::wallet_support::{
 };
 
 pub(crate) fn cmd_setup() -> DynResult<()> {
+    // Load project first so an outdated scaffold.toml gets the canonical
+    // "run `lgs init`" hint before we surface unrelated environment
+    // gripes (circuits artifact, etc.). Tests assert the migration hint
+    // wins on pre-v0.2.0 configs.
     let project = load_project()?;
```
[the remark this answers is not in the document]
weboko: The reorder Copilot suggests is already in place — `load_project()` runs at [src/commands/setup.rs:24](https://github.com/logos-co/scaffold/blob/0a3c3d4/src/commands/setup.rs#L24) and `ensure_logos_blockchain_circuits_present()` at line 25. Out-of-project `setup` invocations get the project-context error first; this is also covered by `setup_hard_fails_on_pre_v0_2_0_scaffold_toml` and `setup_fails_outside_project_with_project_scoped_message` in tests/cli.rs. Resolving as already-addressed.

src/doctor_checks.rs
```diff
@@ -146,6 +146,64 @@ pub(crate) fn check_port_warn(name: &str, addr: &str, remediation: &str) -> Chec
...
+        status: CheckStatus::Fail,
+        name: "logos-blockchain-circuits".to_string(),
+        detail: format!(
+            "not found: $LOGOS_BLOCKCHAIN_CIRCUITS unset and {} is missing",
+            home_hint
+        ),
+        remediation: Some(format!(
+            "Obtain the logos-blockchain-circuits release and either set LOGOS_BLOCKCHAIN_CIRCUITS=<path> or place it at {}",
+            home_hint
+        )),
```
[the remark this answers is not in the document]
weboko: Good catch — fixed in 0a3c3d4. The FAIL detail now distinguishes three env-var states: `unset`, `set to nonexistent path `<path>``, and `set to non-directory `<path>``. Empty-string env values fall through to the "unset" branch (since they can't point anywhere meaningful). Refactored the body into a pure `check_logos_blockchain_circuits_with(env_path, home_dir)` helper so the row shapes are testable without process-global env mutation; five new unit tests cover all four shapes.

src/commands/basecamp.rs
```diff
@@ -994,6 +1023,9 @@ fn cmd_basecamp_modules(
         print_modules_table("captured modules", &existing_modules);
         return Ok(());
     }
+    // Note: nix preflight for production callers lives in `cmd_basecamp`'s
```
fryorcraken: **Important — invariant is fragile.** This note says nix preflight for production callers lives in `cmd_basecamp`'s `Modules` dispatch arm, but `cmd_basecamp_install` (which *is* a production caller) auto-invokes `cmd_basecamp_modules` further down when `total_captured_modules == 0`. That path bypasses the dispatch-arm preflight.

Today it's covered because `cmd_basecamp_install` runs its own `ensure_nix_present()` earlier, so this is latent — not a live bug. But a future refactor moving the install preflight could silently regress the B1 UX (`os error 2` with no signal that `nix` is the missing executable).

**Fix options:**
- (a) Strengthen this comment to "callers MUST preflight nix; see `cmd_basecamp` dispatch + `cmd_basecamp_install`" — zero code change.
- (b) Lift `ensure_nix_present()` into the top of `cmd_basecamp_modules` itself (after the `--show` early-return) and drop the two dispatch-site calls. More robust; one extra preflight in the `install` → `modules` auto-discover path, but `which` is cheap.

src/repo.rs
```diff
@@ -231,6 +243,34 @@ fn looks_like_url(source: &str) -> bool {
...
+/// private/typo'd URL doesn't hang at a `Username:` prompt.
+/// For local paths, check the directory looks like a git repo (worktree
+/// `.git`, bare repo `HEAD` + `objects`).
+fn source_is_reachable(source: &str) -> bool {
+    if looks_like_url(source) {
+        std::process::Command::new("git")
+            .env("GIT_TERMINAL_PROMPT", "0")
+            .env("GIT_ASKPASS", "/bin/false")
+            .env("GCM_INTERACTIVE", "Never")
+            .arg("ls-remote")
```
fryorcraken, line 285: **Suggestion.** Every other git invocation in this file goes through `run_checked` / the `crate::process` helpers so command echo and labelling stay consistent. This one deliberately bypasses them (silent probe, swallowed stderr, bool return) — which is the right call for a reachability check. Consider a one-line `//` noting *why* this probe sidesteps the `run_*` helpers; readers grepping for git invocations will otherwise wonder why this one is special.

src/commands/new.rs
```diff
@@ -38,6 +38,24 @@ pub(crate) fn cmd_new(cmd: NewCommand) -> DynResult<()> {
...
+    }
+
+    // Run the rest in an inner function so we can clean up `target` on
+    // failure. Without this, a sync/template error (e.g. a typo on
+    // `--lez-path`) leaves a half-built project directory behind. The
+    // `target.exists()` guard above guarantees we only delete a directory
+    // we created ourselves in this run.
+    let result = cmd_new_inner(&cmd, &target, &template_variant);
+    if result.is_err() {
+        let _ = fs::remove_dir_all(&target);
```
fryorcraken: **Suggestion (merged S2 + S6).** Two minor things on this cleanup line:

1. If `cmd_new_inner` fails before the first `fs::create_dir_all(target.join(...))` call (currently line 68), `target` doesn't exist and this is a no-op — `let _` correctly swallows `NotFound`. Worth a one-line comment confirming intent: `// ignore NotFound — inner may have failed before creating target`.
2. If `remove_dir_all` itself fails (perms, etc.), the user gets the original error with no signal that a half-built project dir remains. Consider an `eprintln!` warn on cleanup failure so they know to clean it up manually.

Both optional.

src/cli.rs
```diff
@@ -546,6 +590,13 @@ fn spel_passthrough_args(args: &[String]) -> DynResult<Option<Vec<String>>> {
     if args.len() < 2 || args[1] != "spel" {
         return Ok(None);
     }
+    // Let `lgs spel --help` / `-h` reach clap so the user sees the
+    // variant's `long_about` (which documents the passthrough form).
+    // Without this, the intercept below would trigger the "Did you mean"
+    // error path and shadow clap's help.
+    if args.len() >= 3 && (args[2] == "--help" || args[2] == "-h") {
```
fryorcraken: **Suggestion.** This catches `--help` / `-h` but not `help` or `-?`, which clap also recognizes. Not a regression — pre-PR, all `spel <foo>` shapes hit the "Did you mean" hint, so `lgs spel help` also fell through. If you want full parity with clap's help conventions, the check is one extra arm. Optional.

src/commands/new.rs
```diff
@@ -62,9 +76,24 @@ pub(crate) fn cmd_new(cmd: NewCommand) -> DynResult<()> {
...
+    // shell-prefixed lines before any human-friendly status. Suppress
+    // the echo for the LEZ sync only — `setup`, `localnet`, etc. keep
+    // their existing echo behavior. `git clone --no-hardlinks` still
+    // prints its own progress to stderr, which is reassuring during a
+    // slow first clone.
+    println!(
+        "Cloning lez at pin {} from {} (this may take a minute the first time)...",
+        DEFAULT_LEZ.sha, lez_source
+    );
+    let _echo_guard = crate::process::EchoGuard::suppress();
```
fryorcraken: **Suggestion.** `EchoGuard` lives until function exit, so the *post-clone* steps (template copy, overlay, `cargo init` if ever added) also run under suppression. Today that's benign — none of them shell out — but if someone later adds e.g. a `cargo init` here it would silently inherit the suppression. Consider scoping the guard tightly to the `lez_repo_path = if cmd.vendor_deps { ... } else { ... }` block so the intent ("suppress echo for the LEZ sync only") is enforced by lexical scope, not by what happens to live below. Optional.

src/doctor_checks.rs
```diff
@@ -146,6 +146,79 @@ pub(crate) fn check_port_warn(name: &str, addr: &str, remediation: &str) -> Chec
...
+
+/// Pure helper: takes the resolved env-var path (or `None` when unset/empty)
+/// and the resolved `~/.logos-blockchain-circuits` path (or `None` when
+/// `$HOME` is missing), returns the `CheckRow`. Split out so the row
+/// shaping is testable without process-global env mutation.
+fn check_logos_blockchain_circuits_with(
+    env_path: Option<&Path>,
+    home_dir: Option<&Path>,
+) -> CheckRow {
+    if let Some(path) = env_path.filter(|p| p.is_dir()) {
```
fryorcraken, line 182: **Nit.** `is_dir()` follows symlinks, so a symlink-to-directory passes (good). A *broken* symlink would currently report as "set to non-directory" — technically accurate but might confuse a user whose symlink target moved. Current behavior is defensible; flagging only in case you want to distinguish that case in the failure detail.

fryorcraken (commented):
Reviewed across the five axes (correctness, readability, architecture, security, performance). Overall this clearly improves health — every fix is scoped, surfaces a scaffold-styled error in place of a raw OS/cargo trace, and carries focused unit tests. One **Important** invariant worth tightening (I1) and a handful of **Suggestions** below as inline comments.

**Verification cross-check:** the diff matches the PR body's test claims (new `source_mismatch_reclone_refuses_when_new_source_unreachable` is present in `src/repo.rs`; the five new circuits-check tests are in `src/doctor_checks.rs`; the early-intercept `--help` short-circuit and `unreachable!` arm are both in `src/cli.rs`).

Not blocking on my end — leaving the final call to a maintainer.

src/repo.rs
```diff
@@ -238,6 +250,36 @@ fn looks_like_url(source: &str) -> bool {
...
+/// For local paths, check the directory looks like a git repo (worktree
+/// `.git`, bare repo `HEAD` + `objects`).
+fn source_is_reachable(source: &str) -> bool {
+    if looks_like_url(source) {
+        // Intentionally bypass run_* helpers: this silent probe should collapse
+        // auth, DNS, and missing-remote failures into `false`.
+        std::process::Command::new("git")
+            .env("GIT_TERMINAL_PROMPT", "0")
+            .env("GIT_ASKPASS", "/bin/false")
+            .env("GCM_INTERACTIVE", "Never")
```
fryorcraken, line 270: **SSH URL probe can still hang.** Prompt-suppression handles HTTPS (`GIT_TERMINAL_PROMPT=0` + `GIT_ASKPASS=/bin/false` + `GCM_INTERACTIVE=Never`) but not SSH. For `git@github.com:foo/bar.git` style sources with no agent + key, OpenSSH can fall back to `$SSH_ASKPASS` (gnome-ssh-askpass etc.) when `$DISPLAY` is set, defeating the F3a guarantee on a developer laptop with a desktop session.

Suggest also setting:
```rust
.env("SSH_ASKPASS", "/bin/false")
.env("SSH_ASKPASS_REQUIRE", "never")  // OpenSSH 8.4+
.env("GIT_SSH_COMMAND", "ssh -o BatchMode=yes -o ConnectTimeout=10")
```
The `BatchMode=yes` is the belt-and-suspenders — it tells ssh to never prompt, period. `ConnectTimeout=10` also addresses the no-timeout concern (see separate comment).

src/repo.rs
```diff
@@ -238,6 +250,36 @@ fn looks_like_url(source: &str) -> bool {
...
+            .arg("ls-remote")
+            .arg("--exit-code")
+            .arg(source)
+            .arg("HEAD")
+            .stdout(std::process::Stdio::null())
+            .stderr(std::process::Stdio::null())
+            .status()
```
fryorcraken, line 291: **No timeout on the probe.** `ls-remote` against an unreachable host can hang on DNS for 30s+ even with all prompts disabled, which would turn a `--lez-path` typo from "instant scaffold-styled error" into "the CLI froze." The `GIT_SSH_COMMAND="ssh -o ConnectTimeout=10"` suggestion handles the SSH case; for HTTPS, consider wrapping in a `Command::spawn` + thread-poll with a deadline, or setting `GIT_HTTP_LOW_SPEED_LIMIT=1` + `GIT_HTTP_LOW_SPEED_TIME=10`.

src/repo.rs
```diff
@@ -238,6 +250,36 @@ fn looks_like_url(source: &str) -> bool {
...
+            .map(|s| s.success())
+            .unwrap_or(false)
+    } else {
+        let p = Path::new(source);
+        p.exists()
+            && (p.join(".git").exists() || p.join("HEAD").exists() || p.join("objects").exists())
+    }
```
fryorcraken, line 298: **Naming nit:** `source_is_reachable` returns `false` for "path exists but isn't a git repo" (e.g. `--lez-path /tmp`). The error message at the bail site says "path does not exist or remote refused HEAD probe" — a user pointing at the wrong directory will be confused. Consider broadening the bail wording to "path does not exist, is not a git repo, or remote refused HEAD probe."

src/commands/new.rs
```diff
@@ -63,31 +86,47 @@ pub(crate) fn cmd_new(cmd: NewCommand) -> DynResult<()> {
...
+    // the echo for the LEZ sync only — `setup`, `localnet`, etc. keep
+    // their existing echo behavior. `git clone --no-hardlinks` still
+    // prints its own progress to stderr, which is reassuring during a
+    // slow first clone.
+    println!(
+        "Cloning lez at pin {} from {} (this may take a minute the first time)...",
+        DEFAULT_LEZ.sha, lez_source
+    );
+    let lez_repo_path = {
+        let _echo_guard = crate::process::EchoGuard::suppress();
```
fryorcraken, line 106: **`EchoGuard` scope is wider than the comment implies.** The guard is bound to the rest of `cmd_new_inner`, not just the LEZ sync. Currently harmless because the body has no further subprocess calls, but if a future change adds e.g. `git init` in the new project, echo will silently stay suppressed and the user will lose a useful signal. Either narrow with a block (`{ let _g = EchoGuard::suppress(); sync_repo_to_pin_at_path_with_opts(...) }`) or update the comment to say "for the rest of `cmd_new_inner`."

tests/cli.rs
```diff
@@ -2143,6 +2143,21 @@ fn spel_without_dash_dash_suggests_passthrough_form() {
         );
 }
 
+#[test]
+fn spel_help_accepts_clap_help_spellings() {
```
fryorcraken, line 2224: Nice catch covering the `--help` / `-h` clap-help-spelling cases. Two coverage gaps worth considering as follow-up (not blocking):
- **F1 (Blocker):** an integration test that runs `setup` with `LOGOS_BLOCKCHAIN_CIRCUITS` unset asserts the scaffold-styled error fires before any cargo invocation. The `check_logos_blockchain_circuits_with` unit tests cover the doctor row but not the `setup` bail path end-to-end.
- **F4:** `bogus-subcommand` produces a single `error:` prefix, not `error: error:`. Currently only verified manually.

fryorcraken (commented):
Re-reviewed against the true PR head (1f51638). The earlier draft I sketched offline was against `0a3c3d4` (an older head) and several findings have since been addressed — apologies for the noise. After re-checking the current head, the substantive surviving concerns are around the `git ls-remote` probe's robustness for SSH URLs and unreachable hosts. The doctor refactor and F3a regression test are particularly well-done.

**Verdict:** comment-only — none of these block merge, but the SSH-probe hardening is worth doing before this lands since the whole point of F3a is to avoid surprising the user on a typo.

fryorcraken (approved):

src/repo.rs
weboko: Addressed in 7882197. The probe now also sets:

```rust
.env("SSH_ASKPASS", "/bin/false")
.env("SSH_ASKPASS_REQUIRE", "never")
.env("GIT_SSH_COMMAND", "ssh -o BatchMode=yes -o ConnectTimeout=10")
```

`BatchMode=yes` makes ssh refuse to prompt, period (covers any askpass fallback OpenSSH might still try), and `ConnectTimeout=10` doubles as the SSH-side bound for the no-timeout concern in the sibling thread.

src/repo.rs
weboko: Addressed in 7882197. HTTPS now gets:

```rust
.env("GIT_HTTP_LOW_SPEED_LIMIT", "1")
.env("GIT_HTTP_LOW_SPEED_TIME", "10")
```

So a transfer below 1 B/s for more than 10 s aborts. SSH is bounded by the `ConnectTimeout=10` in `GIT_SSH_COMMAND` from the sibling thread. Together they cover the DNS-stall / dropped-connection cases without needing a Rust-side deadline thread.

src/repo.rs
weboko: Addressed in 7882197. The bail message is now:

> `<label>` source ``<source>`` is not reachable (path does not exist, **is not a git repo**, or remote refused HEAD probe). Refusing to discard the existing cache at `<path>` — fix the path/URL and retry.

A user pointing `--lez-path` at e.g. `/tmp` now gets the middle clause as the proximate diagnosis.

src/commands/new.rs
weboko: The guard is already block-scoped — `cmd_new_inner` reads:

```rust
let lez_repo_path = {
    let _echo_guard = crate::process::EchoGuard::suppress();
    if cmd.vendor_deps {
        ...
        sync_repo_to_pin_at_path_with_opts(...)?;
        lez_vendor
    } else {
        ...
        sync_repo_to_pin_at_path_with_opts(...)?;
        lez_cached
    }
};
```

The guard binds to the outer `{ ... }`, not to `cmd_new_inner`, so it drops at the `};` and the post-clone template/overlay steps run with echo restored — the comment matches the code. (This was the form merged from the previous round of review on 05-12; happy to re-verify if I am misreading.)

tests/cli.rs
weboko: Both coverage gaps closed in b5a8f16:

- **F1**: `setup_bails_with_scaffold_styled_error_when_circuits_missing` runs `lgs setup` with `LOGOS_BLOCKCHAIN_CIRCUITS` removed and `HOME` repointed at an empty dir, asserts the scaffold-styled error text (`logos-blockchain-circuits` / `$LOGOS_BLOCKCHAIN_CIRCUITS unset` / `logos-scaffold doctor`) **and** that stderr contains neither `logos-blockchain-pol` nor `build script` — so a regression that lets cargo run before the precheck would fail loudly.
- **F4**: `unrecognized_subcommand_produces_single_error_prefix` runs `lgs bogus-subcommand` and asserts the stderr surfaces the unrecognized-subcommand message **and** does not contain `error: error:`.

PR #140, opened 2026-05-11
fix(doctor): tighten localnet-connectivity heuristic to require transport-error + address

### LINKED ISSUE #113: doctor's connectivity heuristic mis-classifies real wallet failures as warnings
## Problem

`is_localnet_connectivity_failure` (`src/commands/doctor.rs:298-308`, used at `:457-463`) decides whether to downgrade a wallet failure from `Fail` to `Warn` based on a substring match against the wallet's combined stdout+stderr — including a literal needle for `"127.0.0.1:3040"` and `"localhost:3040"`.

But `127.0.0.1:3040` shows up in *any* wallet output that mentions the configured sequencer address, including completely genuine non-connectivity failures (e.g. RPC rejected the call, signature mismatch, malformed payload — all of which echo the target URL in error context). Those real `Fail`s get reported as `Warn: cannot reach sequencer`.

This couples the connectivity classifier to the same hardcoded port string already filed in #40 — but the bug here is the misclassification logic, not just the port literal.

## Why it matters

`doctor`'s job is to surface failures. Misclassifying a real wallet bug as a transient connectivity warning hides regressions and makes the "next steps" advice useless (it tells the user to start the sequencer — which is already running).

## Proposed solution

Tighten the heuristic. Require at least one explicit connection-error token (`connection refused`, `tcp connect error`, `econnrefused`, `dns`, `connecterror`, `network is unreachable`) and *also* the address. Lines that only echo the address in unrelated context should not trigger the downgrade.

Better: take the structured wallet exit code into account. A wallet binary that fails on RPC connectivity has a distinguishable exit shape vs. one that fails on protocol-level errors.

Related but separate: this code path also substring-matches `127.0.0.1:3040` instead of using `project.config.localnet.port`. See #40 for the port hardcoding (that issue covers a different code path; this one survives that fix's narrow scope).

## Source

`code` lens — review of `src/commands/doctor.rs`.

## Traceability

- Run ID: `20260508-181437`

wozos:
## Description
`doctor`'s wallet-usability check currently downgrades a real wallet failure to a `Warn: cannot reach sequencer` whenever the wallet's output mentions `127.0.0.1:3040` or `localhost:3040`. Wallet binaries routinely echo the configured sequencer URL in unrelated error contexts (RPC rejection, signature mismatch, malformed payload), so genuine functional failures are silently reclassified as transient connectivity warnings — exactly the regressions `doctor` exists to surface. The misclassification also makes the "Run `logos-scaffold localnet start`" remediation actively misleading, since the sequencer is already running.

## Solution
- Rewrote `is_localnet_connectivity_failure` in `src/commands/doctor.rs` so the downgrade fires only when the wallet output contains **both** the sequencer address (`127.0.0.1:3040` / `localhost:3040`) **and** an explicit transport-error token. The token list now covers the common shapes: `connection refused`, `econnrefused`, `tcp connect error`, `network is unreachable`, `no route to host`, `failed to connect`, `couldn't connect`, `connecterror`, `dns error`.
- A short block comment captures the design rule (`mentions URL alone is not enough`) so a future reader doesn't reintroduce the bare-address shortcut.
- Added seven unit tests in `src/commands/doctor.rs` covering: refused-connection with the address triggers; `ECONNREFUSED` over `localhost` triggers; signature-mismatch / `400 Bad Request` failures that merely echo the URL stay `Fail`; bare URL with no transport token stays `Fail`; transport-error token without the localnet address stays `Fail` (so an external RPC failure is not misclassified as a localnet outage); and the match is case-insensitive.

All 244 lib unit tests and 135 integration tests pass on the worktree (`cargo build`, `cargo test`).

## Notes
- This change is intentionally narrow to the heuristic and does not touch the hardcoded `3040` port literal — that is tracked separately in #40 (the issue calls this out).
- Trade-off considered: leaving any one transport-error token alone trigger the downgrade (without requiring the address). Rejected because non-localnet transport errors (e.g., proxy/external RPC) could then be misclassified as a localnet outage and produce the wrong remediation. The "address + token" rule is the conservative call.
- A wallet that fails to connect but does not echo the URL in its output will now stay `Fail` rather than `Warn`. That is the safer default (fail loud, don't hide regressions); the prior false-positive direction was the actual bug.

Closes #113

---
Run ID: 20260512-003012

### DIFF
--- a/src/commands/doctor.rs
+++ b/src/commands/doctor.rs
@@ -454,12 +454,34 @@ fn check_spel_lez_alignment(spel_path: &std::path::Path) -> CheckRow {
     }
 }
 
+// Decide whether a wallet-check failure should be downgraded from Fail to
+// Warn because the local sequencer is simply unreachable. Conservative on
+// purpose: doctor's job is to surface failures, so a misclassification in
+// either direction has to be a real connectivity error before the user
+// sees the friendly "start localnet" hint instead of the raw failure.
+//
+// The hard rule is that mentioning the sequencer URL is **not enough** —
+// wallet binaries routinely echo the configured address in unrelated
+// error contexts (RPC rejection, signature mismatch, malformed payload).
+// We require *both* an explicit transport-error token *and* the address,
+// so an unrelated failure that happens to print the URL is left as Fail.
 fn is_localnet_connectivity_failure(stdout: &str, stderr: &str) -> bool {
     let text = format!("{stdout}\n{stderr}").to_lowercase();
-    text.contains("connection refused")
+
+    let mentions_localnet_address =
+        text.contains("127.0.0.1:3040") || text.contains("localhost:3040");
+
+    let has_transport_error_token = text.contains("connection refused")
+        || text.contains("econnrefused")
+        || text.contains("tcp connect error")
+        || text.contains("network is unreachable")
+        || text.contains("no route to host")
+        || text.contains("failed to connect")
+        || text.contains("couldn't connect")
         || text.contains("connecterror")
-        || text.contains("127.0.0.1:3040")
-        || text.contains("localhost:3040")
+        || text.contains("dns error");
+
+    mentions_localnet_address && has_transport_error_token
 }
 
 fn derive_next_steps(rows: &[CheckRow]) -> Vec<String> {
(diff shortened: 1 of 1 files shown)


danisharora099 (approved):

danisharora099:
@weboko you need to enable signed commits for @wozos

PR #149, opened 2026-05-12
feat(cli): add --cache-root flag to new/create

### LINKED ISSUE #122: DOGFOODING/E2 and README claim --cache-root flag on new/create that does not exist
## Problem

`DOGFOODING.md` E2 (`Project Creation with Advanced Flags`, lines ~129 and ~769-790) instructs:

```bash
"$SCAFFOLD_BIN" new dogfood-cache --cache-root "$SCRATCH_ROOT/custom-cache"
```

with success criteria `--cache-root is honored and scaffold uses the specified directory for cache operations`. README's "Command Semantics" section (line ~129) also lists `--cache-root` among the flags `create`/`new` accept.

But `src/cli.rs` `NewArgs` defines only:

```rust
struct NewArgs {
    name: String,
    #[arg(long)] vendor_deps: bool,
    #[arg(long, alias = "lssa-path")] lez_path: Option<PathBuf>,
    #[arg(long, default_value = "default")] template: String,
}
```

There is no `--cache-root` flag. clap will reject the documented invocation with `unexpected argument`. Tests in `tests/cli.rs` set `cache_root = "cache"` *inside scaffold.toml* — that's a TOML field, not a CLI flag; the CLI flag has never existed.

## Why it matters

Dogfooders following the runbook fail at E2's precondition step. Worse, the related `--lez-path` flag is documented honestly, so the omission looks like a typo rather than a missing feature — making it harder to spot.

`cache_root` *is* honored when set as `[scaffold].cache_root` inside `scaffold.toml` (or via `LOGOS_SCAFFOLD_CACHE_ROOT` env), so the underlying capability exists.

## Proposed solution

Pick one:

1. Implement `--cache-root <PATH>` on `new`/`create` that writes the resolved path to `scaffold.toml [scaffold].cache_root` at generation time. This is the user-friendliest option — matches `cargo new --target-dir` ergonomics.
2. Drop `--cache-root` from README's Command Semantics and DOGFOODING E2; replace with a note that documents `LOGOS_SCAFFOLD_CACHE_ROOT="$SCRATCH_ROOT/custom-cache"` as the supported override.

Option 1 is small (one struct field + one config write). Option 2 is even smaller and avoids growing the CLI surface.

## Source

`conceptual` lens — doc-vs-code drift on CLI flag.

## Traceability

- Run ID: `20260508-181437`

wozos:
## Description

`DOGFOODING.md` E2 (`Project Creation with Advanced Flags`) and `README.md` Command Semantics document a `--cache-root <PATH>` flag on `new`/`create`, but `src/cli.rs`'s `NewArgs` only carries `--vendor-deps`, `--lez-path`, and `--template`. Following the runbook verbatim (`"$SCAFFOLD_BIN" new dogfood-cache --cache-root "$SCRATCH_ROOT/custom-cache"`) fails clap parsing with `unexpected argument`, even though the capability already exists via `[scaffold].cache_root` in `scaffold.toml` and `LOGOS_SCAFFOLD_CACHE_ROOT`. This PR wires the CLI flag through so the documented command works.

## Solution

- `src/cli.rs`: add `cache_root: Option<PathBuf>` to `NewArgs` (`#[arg(long, value_name = "PATH")]`) and forward it to `NewCommand` in the `Create`/`New` match arm.
- `src/commands/new.rs`:
  - Add `cache_root: Option<PathBuf>` to `NewCommand`.
  - Resolve the supplied value via a small pure helper `resolve_cli_cache_root(cwd, path)`: absolute paths stay as-is; relative paths join to `cwd` at `new` time (not the future project root, which doesn't exist yet); the resolved absolute is persisted to `[scaffold].cache_root` so `resolve_cache_root` produces the same result from any CWD on subsequent commands.
  - Use the resolved path as the bootstrap LEZ cache root for non-vendored projects, matching the DOGFOODING expected signal that non-vendored clones land under `<cache-root>/repos/lez/<pin>/`. Vendored projects continue to clone into `.scaffold/repos/lez`; the `[scaffold].cache_root` is still written so it applies to future runs.
  - When the flag is omitted, behavior is unchanged (`default_cache_root()` chain + empty persisted value).
- `src/commands/new.rs` tests: three pure unit tests covering absolute, plain-relative, and `./`-prefixed relative path resolution.
- `tests/cli.rs`: `new_help_lists_documented_flags_including_cache_root` verifies `lgs new --help` advertises all four documented flags (`--vendor-deps`, `--lez-path`, `--template`, `--cache-root`) so this regression cannot recur silently.

Build: `cargo build --offline` finishes cleanly.
Tests: `cargo test --offline` — 240 lib + 136 integration passed (was 237 + 135 before; the four new tests included).

## Notes

- Relative-path semantics: anchored to cwd at `new` time, persisted absolute. This is the only consistent option — `resolve_cache_root` joins relative `cache_root` against `project.root`, but during `new` the project root does not yet exist, so a relative literal would resolve to different directories at `new` time vs. later commands. Persisting absolute eliminates the inconsistency; users who want machine-portable scaffold.toml are still served by the `LOGOS_SCAFFOLD_CACHE_ROOT` env layer.
- `--cache-root` paired with `--vendor-deps`: persists to `[scaffold].cache_root` but does not redirect the LEZ clone (vendored LEZ goes under `.scaffold/repos/lez` per the existing branch). The bootstrap cache directories are still created under the supplied path so `cache_root` is a valid resolved location on first use.
- No doc changes needed: `DOGFOODING.md` E2 and `README.md` Command Semantics already describe the flag's intended behavior; this PR makes the implementation match.

Closes #122

---
Run ID: 20260512-043014

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -169,6 +169,14 @@ struct NewArgs {
     lez_path: Option<PathBuf>,
     #[arg(long, default_value = "default", help = TEMPLATE_HELP.as_str())]
     template: String,
+    /// Override the cache root for this project. Persisted to
+    /// `[scaffold].cache_root` in the generated scaffold.toml so all
+    /// subsequent commands honor it; also used during `new` itself for the
+    /// initial LEZ clone of non-vendored projects (`<cache-root>/repos/lez/<pin>/`).
+    /// Relative paths are resolved against the current working directory at
+    /// `new` time and persisted as absolute.
+    #[arg(long, value_name = "PATH")]
+    cache_root: Option<PathBuf>,
 }
 
 #[derive(Debug, clap::Args)]
@@ -457,6 +465,7 @@ pub(crate) fn run(args: Vec<String>) -> DynResult<()> {
             vendor_deps: args.vendor_deps,
             lez_path: args.lez_path,
             template: args.template,
+            cache_root: args.cache_root,
         }),
         Some(Commands::Setup(_)) => cmd_setup(),
         Some(Commands::Build(args)) => match args.subcommand {
--- a/src/commands/new.rs
+++ b/src/commands/new.rs
@@ -26,6 +26,7 @@ pub(crate) struct NewCommand {
     pub(crate) template: String,
     pub(crate) vendor_deps: bool,
     pub(crate) lez_path: Option<PathBuf>,
+    pub(crate) cache_root: Option<PathBuf>,
 }
 
 pub(crate) fn cmd_new(cmd: NewCommand) -> DynResult<()> {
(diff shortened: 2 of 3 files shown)


weboko:
some stupid addition, not needed

PR #150, opened 2026-05-12
feat(new): warn when --lez-path HEAD differs from the scaffold lez pin

### LINKED ISSUE #123: --lez-path silently resets the user-supplied LEZ tree to the scaffold pin, dropping local changes
## Problem

`--lez-path <PATH>` on `lgs new` is documented in README, CONTRIBUTING.md ("Local Development" `cargo run --bin logos-scaffold -- new dogfood-app --lez-path /absolute/path/to/logos-execution-zone`), and the CLI help. The flag's natural reading — used widely in CONTRIBUTING and the dogfooding flow — is "use that LEZ working tree as-is for this scaffold."

In practice (`src/commands/new.rs`), the path is fed into `git clone <lez-path> <dest>` and the destination is then `git checkout <DEFAULT_LEZ.sha>`. Two cases:

- **Cache-managed mode** (no `--vendor-deps`): the user-supplied tree is cloned, then forced to scaffold's pin SHA. Any local commits, branch tip, or uncommitted work in the user's working tree is silently discarded in the clone.
- **Vendored mode** (`--vendor-deps`): same clone-and-reset, just into `.scaffold/repos/lez` instead of cache.

A LEZ co-developer who runs `lgs new my-app --lez-path ~/src/lez` and is on a feature branch sees their feature branch silently dropped in the resulting scaffolded project, and `lgs setup` later resets again to scaffold's pinned SHA every time it touches the tree.

## Why it matters

CONTRIBUTING.md actively recommends this flag for dogfooding LEZ changes against the scaffold. The whole point of the flag, per CONTRIBUTING, is "use my local LEZ work" — but the flag silently overrides the user's checkout to the scaffold pin.

This is also one of the few failure modes that's *invisible*: clone+checkout succeed, the new project builds successfully, the user's LEZ work just isn't there.

## Proposed solution

Pick one:

1. Detect when the supplied path is a worktree at a *different* SHA from `DEFAULT_LEZ.sha`, and either (a) use it in place without cloning + checkout, with a warning, or (b) fail with a targeted error pointing at how to override the pin (e.g., `--lez-pin <sha>` or hand-editing `[repos.lez].pin` post-creation).
2. Document the pin-override semantics loudly in `--lez-path`'s `--help` text and in CONTRIBUTING.md: "the supplied path is cloned and then forced to the scaffold pin; any working-tree state that isn't on that commit is dropped." Surprising-but-documented is better than surprising-and-undocumented.

## Source

`conceptual` lens — feature-vs-docs intent drift.

## Traceability

- Run ID: `20260508-181437`

wozos:
## Description
`lgs new --lez-path <PATH>` is documented as the way to seed a scaffolded project from a local lez checkout (CONTRIBUTING.md actively recommends it for dogfooding). In practice the supplied path is cloned and then `git checkout`-ed to scaffold's pinned lez SHA, so any commits, branch tip, or uncommitted work above that pin silently fail to appear in the new project. The clone + checkout-to-pin succeed, the new project builds cleanly, the user's local lez work just isn't there — exactly the *intent drift* failure mode the issue describes.

## Solution
- Add `lez_path_pin_warning(path, pin)` in `src/repo.rs`: reads the supplied tree's HEAD via `git rev-parse HEAD`, returns `Some(<multi-line warning>)` when HEAD differs from the pin and `None` otherwise (missing path, non-git path, HEAD == pin). The warning names the user's HEAD, the scaffold pin, and points at `[repos.lez].pin` as the override site.
- Wire it into `src/commands/new.rs`: when `cmd.lez_path` is `Some`, emit the warning to stderr before the existing `sync_repo_to_pin_at_path_with_opts` call. Pre-existing behavior (clone + reset to `DEFAULT_LEZ.sha`) is unchanged, so existing dogfooding scripts keep working.
- Tighten the `--lez-path` clap `help` text in `src/cli.rs` to document the pin-reset semantics explicitly.
- Update README.md (`Command Semantics` bullet for `create`/`new`), CONTRIBUTING.md (Dogfooding section), and DOGFOODING.md (B-1 failure signals) with the same note.
- Add four unit tests in `src/repo.rs`: missing path, non-git path, HEAD-matches-pin (no warning), and HEAD-mismatch (asserts the warning names both SHAs and the override site).

## Notes
- Conservative fix: option (2) from the issue's "Proposed solution" section (loudly document + warn at runtime). Option (1) — using the supplied tree in-place or hard-failing on mismatch — would change semantics for dogfooders who *want* the pin reset (which is most current callers, since most local lez checkouts already sit at the pin or close to it). Leaving the override path intact and just making it loud is the lower-risk shape.
- The warning is emitted unconditionally on HEAD mismatch (both cache-managed and `--vendor-deps` modes); the pin reset happens in both modes, and so does the silent override.
- `cargo fmt --check`, `cargo check`, `cargo test` (376 total: 241 lib + 135 integration + 0 doc) all pass on this branch — same gate the CI runs.

Closes #123

---
Run ID: 20260512-044013


weboko:
is not needed

PR #153, opened 2026-05-12
fix(run): drop duplicate post-deploy summary, keep sequencer pointer

### LINKED ISSUE #126: lgs run prints two deploy summaries when no post-deploy hooks are configured
## Problem

`run.rs:91-93` calls `print_deploy_summary` (defined ~143-171) at step `[5/N]` when no post-deploy hooks are configured. But `cmd_deploy` invoked at step `[5/N]` *already* prints its full deploy summary — including program IDs, tx, success/failed counts — to stdout. `print_deploy_summary` then re-walks the project to print a second, less informative summary.

A user reading the output sees two consecutive "summary"-shaped blocks for what is one deploy. The shapes diverge slightly (the second has `discovered_programs` framing the first didn't), which compounds the confusion.

## Why it matters

User-perceived quality: same data twice is noise that hides the actual deploy result. `lgs run`'s value proposition is single-command clarity; the duplicated summary undermines that.

## Proposed solution

Drop `print_deploy_summary` entirely and let `cmd_deploy`'s output stand. If a one-line "deploy complete" footer is desired before post-deploy hooks fire, add a single-line marker (`===> deploy complete (N programs)`) and gate hook execution on it.

If `print_deploy_summary` is the right shape and `cmd_deploy`'s output is the duplicate, suppress `cmd_deploy`'s summary inside `lgs run` (an `EchoGuard`-style suppressor + a `--summary=false` knob threaded through `cmd_deploy`).

Either way, the output should print one summary, not two.

## Source

`conceptual` lens — UX review of `lgs run`'s output.

## Traceability

- Run ID: `20260508-181437`
- Related: #105 (lgs run shipped)

wozos:
## Description
`lgs run` printed two consecutive "summary"-shaped output blocks for a single deploy when no post-deploy hooks were configured. The user reading the terminal saw what looked like two deploy results, and the duplicate hid the actual one. The shapes diverged slightly (the second carried `Deployed programs:` framing the first did not), which compounded the confusion. The `run` command's value proposition is single-command clarity for the build → deploy → run pipeline, so duplicated output here is a direct hit to the primary user flow.

## Solution
- Remove the no-hooks branch's call to `print_deploy_summary`, which re-walked the project at step `[5/N]` and re-printed every deployed program plus a sequencer URL even though `cmd_deploy` had already emitted the canonical per-program summary (succeeded/failed counts, tx, program_id) one step earlier.
- Replace it with `write_run_footer` in `src/commands/run.rs`, a writer-typed helper that prints only the one piece of information `cmd_deploy`'s output does not already carry: a single-line `Sequencer: http://127.0.0.1:<port>` pointer for the localnet endpoint. The footer never walks the filesystem and never enumerates programs, so it cannot drift back into a second summary shape.
- Drop the three `print_deploy_summary_*` unit tests, which only verified that the old function did not panic, and replace them with three tests that pin the new contract on `write_run_footer`: it contains the sequencer pointer, it does NOT contain a `Deployed programs:` or `Binary:` listing, it respects a non-default `localnet.port` (does not leak the default `3040`), and it succeeds with no `methods/guest/src/bin` directory present. Tests capture output into a `Vec<u8>` rather than touching stdout, so they are deterministic.

Files changed:
- `src/commands/run.rs` (+79 / -49)

Build: `cargo build --lib` succeeded.
Tests: `cargo test --lib` — 237 passed / 0 failed. `cargo test --tests` — 135 integration tests passed / 0 failed. The 14 tests in `commands::run::tests` (11 pre-existing + 3 new) all passed.

## Notes
- The hooks branch is unchanged — it never invoked `print_deploy_summary` and already ends with `<=== post_deploy[i/n] OK` markers, so it does not need a footer.
- The single-line sequencer pointer is kept (rather than dropping the function entirely) because `cmd_deploy`'s output does not include the localnet endpoint, and the no-hooks user typically needs that pointer to wire up a client immediately after `lgs run` returns.
- `discover_deployable_programs`, `discover_program_binaries`, and `extract_program_id` remain imported because `resolve_single_program_metadata` (hooks path) still uses them.

Closes #126

---
Run ID: 20260512-051012

### DIFF
--- a/src/commands/run.rs
+++ b/src/commands/run.rs
@@ -1,3 +1,4 @@
+use std::io::Write;
 use std::path::PathBuf;
 use std::process::Command;
 
@@ -75,7 +76,7 @@ fn run_pipeline_once(
     println!("[5/{total_steps}] Deploying programs...");
     cmd_deploy(None, None, false)?;
 
-    // Step 6: Post-deploy hooks (or summary)
+    // Step 6: Post-deploy hooks (or footer)
     if has_hooks {
         let n = hooks.len();
         println!("[6/{total_steps}] Running {n} post-deploy hook(s)...");
@@ -89,7 +90,13 @@ fn run_pipeline_once(
             println!("<=== post_deploy[{}/{n}] OK", i + 1);
         }
     } else {
-        print_deploy_summary(project)?;
+        // `cmd_deploy` already printed the canonical deploy summary above
+        // (succeeded/failed counts, per-program tx + program_id). Re-walking
+        // the project here to print a second summary-shaped block produced
+        // two consecutive "summary"-shaped outputs for a single deploy and
+        // hid the real result (#126). Print only a one-line sequencer
+        // pointer so the user knows where to point a client.
+        let _ = write_run_footer(project, &mut std::io::stdout());
     }
 
     Ok(())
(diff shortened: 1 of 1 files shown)


weboko:
looks not very useful for now

PR #154, opened 2026-05-12
fix(templates): patch logos-blockchain transitive deps to a known-good rev

### LINKED ISSUE #133: lgs build fails: logos-blockchain-storage-service references rocksdb module without rocksdb-backend feature
## Summary

Running `lgs build` on a freshly scaffolded project (template: `lez-framework`) fails to compile a transitive dependency, `logos-blockchain-storage-service` (pulled in via `logos-execution-zone`). The crate's public API references types from a module that's gated behind the `rocksdb-backend` feature, but that feature is not enabled in the build, so compilation fails with `E0433`.

## Steps to reproduce

1. Scaffold a new project: `lgs new <name> --template lez-framework`
2. Run `lgs build` in the project directory.

## Expected

The default scaffold builds cleanly.

## Actual

Build fails with the following errors:

```
error[E0433]: failed to resolve: could not find `rocksdb` in `backend`
  --> ~/.cargo/git/checkouts/logos-blockchain-2d93fcb89448de89/17d5843/services/storage/src/api/chain/mod.rs:17:26
   |
17 | use crate::api::backend::rocksdb::chain::HeaderIdStream;
   |                          ^^^^^^^ could not find `rocksdb` in `backend`
   |
note: found an item that was configured out
  --> services/storage/src/api/backend/mod.rs:2:9
   |
 1 | #[cfg(feature = "rocksdb-backend")]
   |       --------------------------- the item is gated behind the `rocksdb-backend` feature
 2 | pub mod rocksdb;

error[E0433]: failed to resolve: could not find `rocksdb` in `backend`
  --> ~/.cargo/git/checkouts/logos-blockchain-2d93fcb89448de89/17d5843/services/storage/src/api/chain/requests.rs:18:18
   |
18 |         backend::rocksdb::chain::{
   |                  ^^^^^^^ could not find `rocksdb` in `backend`

error: could not compile `logos-blockchain-storage-service` (lib) due to 2 previous errors
error: cargo build --workspace (project) failed with exit status: 101
```

## Environment

- Rust: `rustc 1.92.0 (ded5c06cf 2025-12-08)`
- `lgs` (logos-scaffold) version: `0.1.1`
- Scaffold template: `lez-framework`

## Workaround

None known. Currently blocked on `lgs build`.

wozos:
## Description
`lgs build` on a freshly scaffolded `lez-framework` (or `default`) project fails to compile `logos-blockchain-storage-service`, blocking the primary user flow described in `FURPS.md` Functional 1. The root cause is upstream: LEZ v0.2.0-rc1 (current `DEFAULT_LEZ`) declares its `logos-blockchain-*` workspace deps without a `rev`, so a fresh `cargo build` resolves every transitive `logos-blockchain-*` crate against whatever `main` of `logos-blockchain` happens to be. A regression on `main` (`services/storage/src/api/chain/mod.rs:17` adds an unconditional `use crate::api::backend::rocksdb::chain::HeaderIdStream;` even though `api::backend::rocksdb` is gated behind the `rocksdb-backend` feature) breaks the build with `E0433: could not find rocksdb in backend`.

## Solution
- Add a new `DEFAULT_LB_PIN` constant in `src/constants.rs` set to `81dbb4517aa466358ed425d92fad7d45a0c419fd` — the exact `logos-blockchain` rev LEZ rc1's own `Cargo.lock` was tested against. The accompanying doc-comment ties this rev to `DEFAULT_LEZ` and notes that the patch table can be dropped once `DEFAULT_LEZ` advances to LEZ rc2 or newer (rc2+ pins these deps itself).
- Plumb the rev through a new `{{lb_pin}}` overlay placeholder: extend `OverlayRenderContext` in `src/template/project.rs`, wire the substitution in `render_template_text`, and pass `DEFAULT_LB_PIN` from `src/commands/new.rs`.
- Ship a `[patch."https://github.com/logos-blockchain/logos-blockchain.git"]` table in both `templates/default/Cargo.toml.template` and `templates/lez-framework/Cargo.toml.template`. The table pins all 30 transitive `logos-blockchain-*` crates (the set enumerated by LEZ rc1's `Cargo.lock`) to `{{lb_pin}}`. Every crate must be patched to the same rev because Cargo otherwise leaves un-patched siblings to resolve against `main` HEAD, which would mix incompatible workspace revs.
- Add `overlay_emits_logos_blockchain_patch_table_pinned_to_lb_pin` in `src/template/project.rs`. The test renders both templates, asserts the `[patch."https://github.com/logos-blockchain/logos-blockchain.git"]` header is present, asserts the `logos-blockchain-storage-service` patch entry (the crate at the centre of the reported failure) is rendered, asserts the rendered `rev` matches the supplied `lb_pin`, asserts no `{{lb_pin}}` placeholder leaked, and `toml_edit::DocumentMut::from_str`-parses the result to guard against malformed patch tables surfacing only at user-side `cargo build`.

Files changed:
- `src/constants.rs` (+19 / -0)
- `src/template/project.rs` (+62 / -1)
- `src/commands/new.rs` (+3 / -2)
- `templates/default/Cargo.toml.template` (+41 / -0)
- `templates/lez-framework/Cargo.toml.template` (+41 / -0)

Build: `cargo build --lib` succeeded.
Tests: `cargo test --lib` (238 passed, 0 failed) and `cargo test --test cli` (135 passed, 0 failed). The new regression test passes.

## Notes
- This is a downstream workaround for an upstream bug. The proper long-term fix is one of: (a) `logos-blockchain` re-gates `services/storage/src/api/chain/mod.rs:17` behind `#[cfg(feature = "rocksdb-backend")]`, (b) `logos-execution-zone` v0.2.0-rc1 backports the rev-pinning that rc2+ already ships, or (c) scaffold advances `DEFAULT_LEZ` to LEZ rc2 or newer (this requires the matching `DEFAULT_SPEL` bump per the cross-framework invariant documented above `DEFAULT_LEZ` in `src/constants.rs`). Once any of those land the entire patch table can be dropped; the inline comment in the templates spells this out.
- 30 `logos-blockchain-*` crates appear in the patch table — every one that LEZ rc1's `Cargo.lock` resolved. Cargo emits a warning for unused `[patch]` entries; if the resolver does not need a particular crate in practice the warning is benign. Trimming the list to only the strictly-needed subset was considered and rejected because the `logos-blockchain-*` crates share a workspace upstream and must stay co-versioned — leaving any one to resolve against `main` HEAD would risk ABI skew against the pinned siblings.
- ARM64-specific caveats: none. The patch table is platform-independent.
- Did not bump `DEFAULT_LEZ` / `DEFAULT_SPEL` together because that would couple this bug fix to a release-management decision; the patch is the smaller blast radius.

Closes #133

---
Run ID: 20260512-052012

### DIFF
--- a/src/commands/new.rs
+++ b/src/commands/new.rs
@@ -9,8 +9,8 @@ use crate::config::{
 };
 use crate::constants::{
     DEFAULT_BASECAMP_PIN, DEFAULT_FRAMEWORK_IDL_PATH, DEFAULT_FRAMEWORK_IDL_SPEC,
-    DEFAULT_FRAMEWORK_VERSION, DEFAULT_LEZ, DEFAULT_LGPM_PIN, DEFAULT_SPEL, FRAMEWORK_KIND_DEFAULT,
-    FRAMEWORK_KIND_LEZ_FRAMEWORK, LEZ_SOURCE, SCAFFOLD_TOML_SCHEMA_VERSION,
+    DEFAULT_FRAMEWORK_VERSION, DEFAULT_LB_PIN, DEFAULT_LEZ, DEFAULT_LGPM_PIN, DEFAULT_SPEL,
+    FRAMEWORK_KIND_DEFAULT, FRAMEWORK_KIND_LEZ_FRAMEWORK, LEZ_SOURCE, SCAFFOLD_TOML_SCHEMA_VERSION,
 };
 use crate::model::{Config, FrameworkConfig, FrameworkIdlConfig, LocalnetConfig, RunConfig};
 use crate::project::default_cache_root;
@@ -148,6 +148,7 @@ pub(crate) fn cmd_new(cmd: NewCommand) -> DynResult<()> {
         crate_name: &crate_name,
         lez_pin: &cfg.lez.pin,
         spel_tag: DEFAULT_SPEL.tag,
+        lb_pin: DEFAULT_LB_PIN,
     };
     apply_overlay(&target, &template_variant, &overlay_ctx)?;
     if template_variant == FRAMEWORK_KIND_LEZ_FRAMEWORK {
--- a/src/constants.rs
+++ b/src/constants.rs
@@ -37,6 +37,25 @@ pub(crate) const DEFAULT_SPEL: GitRef = GitRef {
     tag: "v0.2.0-rc.5",
 };
 
+/// `logos-blockchain` rev that scaffold-generated projects use to pin the
+/// `logos-blockchain-*` transitive crates pulled in via LEZ.
+///
+/// `DEFAULT_LEZ` (v0.2.0-rc1) declares its `logos-blockchain-*` deps in
+/// `Cargo.toml` *without* a `rev`, so a downstream `cargo build` resolves
+/// each transitive crate against whatever `main` of `logos-blockchain` is
+/// at the moment of resolution. A regression on `main` (a `services/storage`
+/// file unconditionally `use`s a module gated behind the `rocksdb-backend`
+/// feature) currently breaks `lgs build` on a freshly scaffolded project.
+///
+/// To insulate generated projects from `main` drift, the templates ship a
+/// `[patch."https://github.com/logos-blockchain/logos-blockchain.git"]`
+/// table that pins every transitive `logos-blockchain-*` crate to this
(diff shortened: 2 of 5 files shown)


weboko:
not needed now

weboko (approved):

PR #156, opened 2026-05-12
fix(repo): compare HEAD against resolved pin SHA, not raw pin string

### LINKED ISSUE #111: [repos.*].pin as a tag silently fails the post-checkout SHA-equality assertion
## Problem

`ensure_pin_exists` resolves `<pin>^{commit}` via `git rev-parse --verify` (which accepts tags, branches, and SHAs). The post-checkout assertion in `sync_repo_to_pin_at_path_with_opts` then compares the resolved HEAD SHA against the *raw* `pin` string:

```rust
// src/repo.rs:60-73
run_checked(
    Command::new("git").current_dir(path).arg("checkout").arg(pin),
    &format!("git checkout pin ({label})"),
)?;
let head = git_head_sha(path)?;
if head != pin {
    bail!("{label} pin mismatch after checkout (expected {pin}, got {head})");
}
```

If `[repos.<name>].pin` is a tag or branch ref (e.g. `pin = "v0.2.0"`), `rev-parse` succeeds, `checkout` succeeds, but the equality check on line 67 always fails — `head` is a 40-hex SHA and `pin` is a tag name. The user sees a confusing "pin mismatch after checkout (expected v0.2.0, got 35d8df…)" error even though the checkout worked.

## Why it matters

`pin` semantics imply immutability. README's first-success path and FURPS+ both lean on the SHA-pin invariant. A user who follows the natural intuition and writes a tag instead of a SHA is silently blocked from `lgs setup` with an opaque error that doesn't mention "your pin must be a SHA."

This is also a real near-term hazard because `DEFAULT_LEZ` and `DEFAULT_SPEL` constants are now structs carrying both `.sha` and `.tag`, and several docs reference the tags.

## Proposed solution

Pick one:

1. Validate at config-parse time that `[repos.<name>].pin` is a 40-hex SHA, with a targeted error pointing at "use the resolved commit SHA, not a tag."
2. Resolve `pin` to a SHA once via `git rev-parse <pin>^{commit}` and compare the *resolved* SHA against `head`. This is the friendlier choice because it lets users write `pin = "v0.2.0-rc1"` and have it work.

Add a test exercising tag-as-pin (most natural failure to write).

## Source

`code` lens — review of `src/repo.rs:38-76`.

## Traceability

- Run ID: `20260508-181437`

wozos:
## Description
`[repos.<name>].pin` semantics imply immutability, but the post-checkout assertion in `sync_repo_to_pin_at_path_with_opts` compared the raw `pin` string against the 40-hex HEAD SHA. When `pin` was a tag or branch ref (e.g. `pin = "v0.2.0"`), `ensure_pin_exists` and `git checkout` both succeeded, then the SHA-equality check on `src/repo.rs:67` always failed with a confusing `"<label> pin mismatch after checkout (expected v0.2.0, got 35d8df…)"` even though the checkout had worked. README/FURPS+ lean on the SHA-pin invariant, and users who follow the natural intuition of writing a tag are silently blocked from `lgs setup`.

## Solution
- `src/repo.rs`: `ensure_pin_exists` already runs `git rev-parse --verify <pin>^{commit}` to verify the pin is reachable; capture and return that resolved 40-hex SHA instead of `()`. `sync_repo_to_pin_at_path_with_opts` now compares HEAD against the resolved SHA, so tag, branch, short-SHA, and full-SHA pins all work. Full-SHA pins continue to round-trip identically.
- The pin-mismatch error message now includes both the resolved SHA and the original `pin` for traceability when the assertion does trip (e.g. a concurrent checkout race).
- New regression test `pin_as_tag_succeeds_with_resolved_sha_check` in `src/repo.rs`'s `mod tests`: initializes a source repo, tags HEAD as `v0.2.0`, clones a cache copy, calls `sync_repo_to_pin_at_path_with_opts` with `pin = "v0.2.0"`, and asserts both the call succeeds and `HEAD` equals the tagged commit SHA.

## Notes
- Chose the friendlier of the two solutions named in the issue ("resolve `pin` to a SHA once and compare the resolved SHA against `head`") rather than rejecting non-SHA pins at config-parse time. This lets users write `pin = "v0.2.0-rc1"` and have it work, which matches the intuition surfaced in the issue. If a stricter "pin must be a SHA" invariant is desirable later, it can be layered on top in `parse_repo_ref`.
- `ensure_pin_exists` is `pub(crate)` and only has one in-tree caller (`sync_repo_to_pin_at_path_with_opts`); the signature change from `DynResult<()>` to `DynResult<String>` is contained.
- Full test suite (238 unit + 135 integration + new test) passes on this branch.
- Re-submission of PR #138, which was auto-closed by the `github-actions` rate-limit bot on 2026-05-11 before the `wozos` allowlist entry landed in master (commit 4c1f756). The code change is identical to the original.

Closes #111

---
Run ID: 20260512-055621

### DIFF
--- a/src/repo.rs
+++ b/src/repo.rs
@@ -53,7 +53,7 @@ pub(crate) fn sync_repo_to_pin_at_path_with_opts(
         &format!("git fetch ({label})"),
     );
 
-    ensure_pin_exists(path, source, pin, label)?;
+    let resolved_pin = ensure_pin_exists(path, source, pin, label)?;
 
     run_checked(
         Command::new("git")
@@ -64,9 +64,13 @@ pub(crate) fn sync_repo_to_pin_at_path_with_opts(
     )?;
 
     let head = git_head_sha(path)?;
-    if head != pin {
+    if head != resolved_pin {
+        // Comparison is against the resolved SHA, not the raw `pin` string —
+        // otherwise tag/branch pins (e.g. `pin = "v0.2.0"`) always trip this
+        // assertion even though the checkout succeeded.
         bail!(
-            "{label} pin mismatch after checkout (expected {}, got {})",
+            "{label} pin mismatch after checkout (expected {} resolved from {}, got {})",
+            resolved_pin,
             pin,
             head
         );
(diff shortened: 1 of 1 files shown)


weboko (approved):
seems legit problem

PR #158, opened 2026-05-13
fix(setup): make `lgs build` and `lgs localnet start` work end-to-end

weboko:
`lgs build` and `lgs localnet start` panic in `logos-blockchain-pol`'s build script with "Could not find logos-blockchain-circuits directory": LEZ-pinned `logos-blockchain-*` crates need a `LOGOS_BLOCKCHAIN_CIRCUITS` release on disk (or in the env), and scaffold never provided one. The prior workaround (`#154`) shipped a `[patch."<lb-url>"]` table in the generated `Cargo.toml`, but cargo refuses self-source patches and `cargo build --workspace` would die with `patches must point to different sources` before reaching the build script anyway. Result: *no* freshly scaffolded project could compile.

Fix the chain:

* `circuits::ensure_circuits_for_subprocess` downloads the matching `logos-blockchain-circuits` GitHub release tarball into `<cache_root>/circuits/v<ver>-<triple>` and exports `LOGOS_BLOCKCHAIN_CIRCUITS` for the rest of the process. Idempotent; honours a pre-set env var; uses `curl` so the system trust store picks up corporate CAs (ureq's bundled rustls/webpki-roots doesn't). Wired into `setup`, `idl`, and `localnet start|reset` so every subprocess that calls into the zksign runtime inherits it.
* Drop the broken self-source `[patch]` table from both templates and delete the related `lb_pin` plumbing (constants, `OverlayRenderContext` field, `cmd_new` wiring). Replace the patch-table contract test with a regression guard that asserts the patch table is never re-emitted.
* Patch the `default` template bins to compile against the actual LEZ rc1 API (`SequencerClient::send_transaction(NSSATransaction)` from `sequencer_service_rpc::RpcClient`, returning `HashType` which is `pub struct HashType(pub [u8; 32])`). Add `common` and `sequencer_service_rpc` (client-only) as workspace deps.

## Checklist

- [ ] CI is green
- [ ] Relevant DOGFOODING scenarios rerun (listed above)
- [ ] Docs updated (`README.md`, `DOGFOODING.md`) if user-facing behavior changed
- [ ] I have read [CONTRIBUTING.md](/CONTRIBUTING.md) and this PR fits within the weekly contribution cap


weboko:
This PR fixes pretty much broken scaffold. Bypassing rules for merge.

PR #159, opened 2026-05-14
fix(localnet): widen sequencer max_block_size so deploys don't crash the loop

weboko:
## Summary

The pinned LEZ standalone sequencer caps `max_block_size` at 1 MiB in its debug config, which fits roughly two risc0 guest ELFs (~360 KiB each). The default template ships five sample programs, so the very first post-deploy block overflows and the deferred-tx path takes the sequencer's main loop down:

```
WARN  sequencer_core] Transaction with hash … deferred to next block:
                      block size 1094408 bytes would exceed limit of 1048576 bytes
ERROR sequencer_service] Sequencer failed unexpectedly: Main loop exited unexpectedly
```

The crash leaves the localnet unreachable and breaks the documented first-success path before `lgs wallet topup` / examples can run.

## Fix

Until LEZ stops aborting on block-size deferral, scaffold widens the limit at `localnet start` time by patching `max_block_size` to `"8 MiB"` alongside the existing port patch in the pinned `sequencer_config.json`.

- `patch_sequencer_port` → `patch_sequencer_config` (it now writes more than one field).
- The existing `localnet_start_patches_config_and_uses_configured_port` integration test asserts both `port` and `max_block_size` land in `sequencer_config.json`.

## Test plan

- [x] `cargo test --release --test cli localnet_start_patches_config_and_uses_configured_port` — passes.
- [x] `cargo test --release --test cli localnet` — all 11 localnet integration tests pass.
- [x] `cargo test --release --lib commands::localnet` — all 10 localnet unit tests pass.
- [x] Manual e2e against the default `lgs new` template (D1 in DOGFOODING.md):
  - `lgs new lez-demo` (default template), `lgs build`, `lgs localnet start`
  - `lgs wallet topup` → preflight + auth-transfer init + pinata claim all confirm in successive blocks
  - `lgs deploy` deploys all 5 sample programs; sequencer keeps producing blocks instead of crashing
  - `cargo run --bin run_hello_world -- Public/…` submits successfully, sequencer validates and includes in the next block
  - `lgs wallet -- check-health` returns `All looks good!`
  - `lgs doctor` reports 21 PASS (only the pre-existing dirty-cache WARN remains, expected because we patch the sequencer config file in-place)

### DIFF
--- a/src/commands/localnet.rs
+++ b/src/commands/localnet.rs
@@ -214,7 +214,7 @@ fn cmd_localnet_start(
         bail!("{message}");
     }
 
-    patch_sequencer_port(lez, localnet_port)?;
+    patch_sequencer_config(lez, localnet_port)?;
 
     // Use a path relative to lez (the child's cwd), not relative to the
     // parent's cwd.  `current_dir(lez)` applies before exec, so a parent-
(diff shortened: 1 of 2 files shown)


weboko:
bypassing and merging as it is critical fix and not new logic

PR #162, opened 2026-05-14
fix(run): skip IDL step for non-lez-framework projects

weboko:
### Description

While running the full DOGFOODING.md scenarios end-to-end, three first-success-path issues surfaced. This PR addresses them.

### What changed

[src/commands/run.rs](https://github.com/logos-co/scaffold/blob/claude/dogfooding-testing-PNh6z/src/commands/run.rs) — Gate the Building IDL… step on framework.kind == "lez-framework". build_idl_for_current_project bails on default-template projects rather than no-op'ing, so lgs run on the default template previously failed at step 2/5 with build idl is only supported for lez-framework projects. Step labels ([i/n]) now renumber dynamically so the user sees consecutive markers in both the default (4-step) and lez-framework (5-step) pipelines.
[src/doctor_checks.rs](https://github.com/logos-co/scaffold/blob/claude/dogfooding-testing-PNh6z/src/doctor_checks.rs) + [src/commands/doctor.rs](https://github.com/logos-co/scaffold/blob/claude/dogfooding-testing-PNh6z/src/commands/doctor.rs) — New risc0 r0vm and logos-blockchain-circuits doctor rows. Both are sequencer runtime prereqs that live outside the project tree; without them the sequencer panics on its first block-settlement (zk signature) or first risc0 program execution with opaque ProgramExecutionFailed / Main loop exited unexpectedly errors. Surfacing them at lgs doctor time with actionable install hints replaces a mid-run crash with a clean preflight failure.
[src/cli_help.rs](https://github.com/logos-co/scaffold/blob/claude/dogfooding-testing-PNh6z/src/cli_help.rs) — Drop the logos-scaffold create my-app --cache-root … example from create/new help. The flag was never wired up; the non-vendored cache root is XDG-resolved.
DOGFOODING.md — Add the runtime prereqs (rzup, risc0 rust, r0vm, logos-blockchain-circuits) to "Shared Preconditions". Capture the lez-framework-vs-LEZ-pin compile mismatch as a "Known Gap" under L1 so L1–L4 evidence is recorded the same way across runs. Rewrite E2 against the real new/create flag surface (no --cache-root; document XDG_CACHE_HOME instead).

### Test plan

D1–D7 pass end-to-end against a fresh lgs new dogfood-default project (including lgs run with and without [run].post_deploy hooks).
lgs doctor reports the two new rows with PASS when r0vm/circuits are installed and FAIL with the remediation strings when they're absent.
Existing unit tests in commands::run (42) and doctor_checks (3) pass; no new tests added since these are display/wiring changes.

### DIFF
--- a/DOGFOODING.md
+++ b/DOGFOODING.md
@@ -61,6 +61,25 @@ Do not run project-scoped commands from the repository root unless the scenario 
 - Network access available for setup/build flows that fetch dependencies.
 - No preinstalled `wallet` binary is required. If one exists on `PATH`, do not treat it as the runtime under test for scaffold wallet scenarios.
 - Optional but supported: `LOGOS_SCAFFOLD_WALLET_PASSWORD` when validating password override behavior.
+- Sequencer runtime prerequisites (live outside the project tree but required from the first block onward):
+    - **risc0 toolchain** installed via `rzup` (https://risczero.com/install). The
+      pinned LEZ guest crates require a `risc0-rust` toolchain that supports
+      ruint ≥ 1.18 (verified working: `rzup install rust 1.91.1`). A too-old
+      toolchain (e.g., `1.88.0`) produces `ruint@... requires rustc 1.90` from
+      the guest build.
+    - **r0vm** discoverable by `risc0-zkvm`. Either install via
+      `rzup install r0vm` (lands under
+      `~/.risc0/extensions/v<ver>-cargo-risczero-<platform>/r0vm`) or set
+      `RISC0_SERVER_PATH` to a r0vm binary. Without it the sequencer
+      crashes on its first risc0 program execution with `ProgramExecutionFailed`
+      and `Main loop exited unexpectedly`.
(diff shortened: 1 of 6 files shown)


DOGFOODING.md
```diff
@@ -547,6 +566,20 @@ The `ls` step verifies that LEZ-specific directories were scaffolded before proc
 - If LEZ bootstrap behavior diverges from the default template in setup/localnet/doctor flows, capture the difference explicitly.
 - If `build` does not automatically trigger IDL + client generation for the LEZ template, record that as a regression.
 
+### Known Gap (current pins)
+
+The `lez-framework` pin (`jimmy-claw/lez-framework` rev `1e146970…`) referenced
```
weboko, line 571: @vpavlin is it time to fix it?

fryorcraken (changes requested):
I would say no, that scaffold is only here to support project that use spel.

weboko:
> I would say no

no to what exactly?

weboko:
I need to re-run dogfooding to see if new additions address this PR

PR #164, opened 2026-05-18
fix(basecamp): drop --yes/--no-clean/--dry-run from launch

fryorcraken:
## Summary

- `basecamp launch <profile>` only ever touches the project-local profile dir at `.scaffold/basecamp/profiles/<profile>/`. It does not touch the user's `~/.local/share/Logos/` or any home XDG dir. The clean-slate scrub is the whole point of the command, so the recently-added "destructive by default" framing was wrong: it treated a project-scoped reset like it could harm real user data.
- Remove `--yes`, `--no-clean`, and `--dry-run`, plus the gate that refused a bare `launch`. The command now just runs.
- The empty-`[modules]` bail stays — without it, `launch` would scrub the profile and bring it up with zero modules.

## What changed

- `src/cli.rs`, `src/commands/basecamp.rs` — drop the three flags, the gate, and the `cmd_basecamp_launch_dry_run` helper. Simplify the `Launch` action variant.
- `tests/cli.rs` — remove the three flag-specific tests; keep the empty-modules bail test, dropping the now-unused `--yes`.
- `README.md`, `skills/basecamp/SKILL.md`, `skills/lgs-cli/SKILL.md`, `DOGFOODING.md` (B4) — update synopses to match.

## Test plan

- [x] `cargo test --all-targets` — 152 pass
- [x] `cargo run -- basecamp launch --help` shows the simplified usage with no `--yes`/`--dry-run`/`--no-clean`


weboko (approved):
less is more

PR #165, opened 2026-05-19
fix(errors): stop lying when a stale-schema project hits load_project

fryorcraken:
## Summary

Running `lgs basecamp setup` (or any command going through `load_project`) inside a project with a pre-0.2.0 `scaffold.toml` printed:

```
error: This command must be run inside a logos-scaffold project.
Next step: cd into your scaffolded project directory and retry.: scaffold.toml uses an old schema ([scaffold].version is pre-0.2.0; [repos.lez|spel].url is removed in 0.2.0 (use `source` only)). Run `logos-scaffold init` to migrate to v0.2.0; existing settings are preserved.
```

Three things wrong with that output:

1. **"This command must be run inside a logos-scaffold project" is a lie** — the user *is* in one.
2. **"cd into your scaffolded project directory and retry" is the wrong fix** — the actual fix is `lgs init`.
3. **The marker detail (`[scaffold].version is pre-0.2.0; [repos.lez\|spel].url is removed…`) is noise** — `init`'s migrator handles every variant we detect.

Root cause: every command going through `load_project()` (basecamp, wallet, deploy, report) wrapped any failure with the same generic `.context("This command must be run inside a logos-scaffold project. Next step: cd into…")`. That made sense only for the "no scaffold.toml found" branch and stomped on the stale-schema branch.

## Fix

- `load_project()` itself emits the "cd into a project (or run `lgs create <name>`)" hint for the missing-project case, and bubbles the stale-schema error verbatim. Call sites drop the redundant `.context(...)` wrap.
- Stale-schema error is now one action-only line:
  ```
  error: scaffold.toml uses an old schema. Run `logos-scaffold init` to migrate to v0.2.0; existing settings are preserved.
  ```
- Marker detection is factored into `detect_old_schema_markers` returning a struct of bools, so tests can assert detection without coupling to user-facing message wording.

## Test plan

- [x] `cargo test --all-targets` — 156 pass (was 152: +1 regression test, 3 schema tests refactored to assert markers, 1 outside-project test updated for unified wording)
- [x] Manual repro of the reported scenario: stale `scaffold.toml`, `lgs basecamp setup` now prints exactly:
  ```
  error: scaffold.toml uses an old schema. Run `logos-scaffold init` to migrate to v0.2.0; existing settings are preserved.
  ```
- [x] Outside-project case still emits "This command must be run inside a logos-scaffold project. Next step: cd into your scaffolded project directory and retry, or run `lgs create <name>` …"

### DIFF
--- a/src/commands/basecamp.rs
+++ b/src/commands/basecamp.rs
@@ -70,9 +70,7 @@ pub(crate) fn cmd_basecamp(action: BasecampAction) -> DynResult<()> {
         return cmd_basecamp_docs();
     }
 
-    let project = load_project().context(
-        "This command must be run inside a logos-scaffold project.\nNext step: cd into your scaffolded project directory and retry.",
-    )?;
+    let project = load_project()?;
 
     match action {
         BasecampAction::Setup => cmd_basecamp_setup(project),
--- a/src/commands/deploy.rs
+++ b/src/commands/deploy.rs
@@ -36,9 +36,7 @@ pub(crate) fn cmd_deploy(
     program_path: Option<PathBuf>,
     json: bool,
 ) -> DynResult<()> {
-    let project = load_project().context(
-        "This command must be run inside a logos-scaffold project.\nNext step: cd into your scaffolded project directory and retry.",
-    )?;
+    let project = load_project()?;
     let wallet = load_wallet_runtime(&project)?;
     let spel_bin =
         resolve_repo_path(&project, &project.config.spel, "spel")?.join(SPEL_BIN_REL_PATH);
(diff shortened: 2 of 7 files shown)


src/commands/basecamp.rs
```diff
@@ -70,9 +70,7 @@ pub(crate) fn cmd_basecamp(action: BasecampAction) -> DynResult<()> {
         return cmd_basecamp_docs();
     }
 
-    let project = load_project().context(
-        "This command must be run inside a logos-scaffold project.\nNext step: cd into your scaffolded project directory and retry.",
-    )?;
+    let project = load_project()?;
```
weboko, line 73: noticed the same today (also fixed in another PR)

weboko (approved):

PR #166, opened 2026-05-19
ci: publish release candidates from master

danisharora099:
## Summary
- Publish release candidates automatically on pushes to `master`, covering merged PRs and direct commits.
- Preserve the existing tag/manual stable publish flow.
- Generate unique RC crate versions in CI and update the manifest/lockfile before `cargo publish --locked`.

## Validation
- Parsed `.github/workflows/publish-crate.yml` as YAML locally.
- Ran `git diff --check` for the workflow change.
- Smoke-tested the RC version calculation and manifest version substitution locally.

### DIFF
--- a/.github/workflows/publish-crate.yml
+++ b/.github/workflows/publish-crate.yml
@@ -66,7 +66,7 @@ jobs:
             candidate_base="$manifest_version"
 
             url="https://crates.io/api/v1/crates/${CRATE_NAME}/${manifest_version}"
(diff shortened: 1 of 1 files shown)


fryorcraken (changes requested):
Not convinced we need this because one can just checkout and do `cargo install`.

I would focus instead of doing releases (tagging) regularly enough so users can have a stable experience.

danisharora099:
Superseded by https://github.com/logos-co/scaffold/issues/170 — per discussion the agreed direction is bi-weekly stable releases rather than RCs. Closing in favour of that tracking issue.

PR #168, opened 2026-05-20
docs(basecamp): drop dangling basecamp-profiles spec citations

fryorcraken:
## Summary

`src/commands/basecamp.rs` had five citations to a "basecamp-profiles spec" with numbered sections (§2.2, §2.3, §3.4, §6). I tried to follow them and couldn't — there's no such doc in this repo, no such doc in `logos-co/logos-basecamp`, and no URL anywhere in the codebase pointing at one. The basecamp app's own `docs/spec.md` covers a different scope (app architecture, not profiles) and uses named, unnumbered sections, so the section numbers don't match any reachable artifact.

This drops the `§X.Y` qualifiers from all five citation sites. The rationale stays inline — every affected comment still says *what* the rule is and *why* it exists; readers just don't get pointed at a doc they can't find.

Surfaced by PR #164 review (the §2.3 cite was the first one I tried to verify after that PR's comment edits).

## What changed

- `src/commands/basecamp.rs:232` — macOS bundle hint: drop "track basecamp-profiles spec §6".
- `src/commands/basecamp.rs:242` — concurrent-launch comment: drop "per spec §2.3".
- `src/commands/basecamp.rs:343–346` — port-override env comment: drop both "spec §3.4" and "basecamp-profiles §3.4".
- `src/commands/basecamp.rs:358` — `launch_env` docstring: drop "(spec §3.4)".
- `src/commands/basecamp.rs:2336` — `resolve_install_sources` precedence list: drop "(matches §2.2 of the basecamp-profiles spec)".

Net: 9 lines changed, 0 behavior change. Same number of lines deleted as added — comments shrink slightly because the dangling qualifier goes away.

## Test plan

- [x] `cargo check` clean
- [x] `cargo test --all-targets` — 153 pass
- [x] `grep "§\\|basecamp-profiles spec" src/commands/basecamp.rs` — no remaining hits

### DIFF
--- a/src/commands/basecamp.rs
+++ b/src/commands/basecamp.rs
@@ -229,7 +229,7 @@ fn resolve_basecamp_binary(app_link: &Path) -> DynResult<PathBuf> {
     }
     let platform_hint = if cfg!(target_os = "macos") {
         "\nNote: on macOS, `nix build .#app` produces an app bundle under Applications/. \
-         v0.1.x does not yet expose a CLI-invocable binary on macOS; track basecamp-profiles spec §6."
+         v0.1.x does not yet expose a CLI-invocable binary on macOS."
     } else {
         ""
     };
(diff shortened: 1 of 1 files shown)


weboko (approved):
AI 🤷‍♂️

PR #169, opened 2026-05-20
Fix DEFAULT_SPEL pin to public rc5 commit

danisharora099:
## Summary

Fix the broken `DEFAULT_SPEL` pin and tighten the way scaffold carries it.

The previous default SHA, `ed3bbedb4b684645da05455d30a4a0be7cc4dfe0`, is a GitHub PR merge commit for `logos-co/spel#164`. GitHub can resolve it, but it is not advertised by the normal public branch/tag refs that `scaffold setup` fetches from `logos-co/spel`, so cold setups against this default fail to resolve the pin.

This PR replaces it with `1db7c5f8af3165318f9046c13ab185e24773fd59`, the public commit that the existing `v0.2.0-rc.5` tag peels to. The effective SPel version scaffold defaults to is unchanged.

## What changed

- Replaced `DEFAULT_SPEL: GitRef { sha, tag }` in `src/constants.rs` with `DEFAULT_SPEL_PIN: &str = "1db7c5f8af3165318f9046c13ab185e24773fd59"`.
- Dropped the unused `{{spel_tag}}` template plumbing (`OverlayRenderContext.spel_tag`, the `.replace("{{spel_tag}}", …)` call, test fixtures, and the substitution-only test) — no template file ever consumed it.
- Updated all `DEFAULT_SPEL.sha` callsites in `commands/init.rs`, `commands/new.rs`, `commands/doctor.rs`, `migrate.rs`, and `config.rs` to `DEFAULT_SPEL_PIN`.
- Updated the regression test to assert the commit pin only.
- Updated `FURPS.md` wording from tag-based to commit-based.

`DEFAULT_LEZ` keeps its `GitRef { sha, tag }` shape — the LEZ tag is still consumed by `check_spel_lez_alignment` in `commands/doctor.rs` to read spel's vendored LEZ ref.

## Why commit-only

Carrying the mutable `v0.2.0-rc.5` tag in scaffold's defaults made the SPel default appear bump-able by retagging upstream — which would silently change scaffold's behaviour at every `lgs setup`. With a commit-only pin, the default cannot drift.

Per @fryorcraken's review on the previous revision of this PR.

## Scope

Narrow:

- Keeps the same effective SPel default (the commit `v0.2.0-rc.5` peels to).
- Does **not** bump SPel to a stable release.
- Does **not** bump LEZ or change `DEFAULT_LEZ`.
- Does **not** add CI guardrails.

Migrating off RC-derived defaults (matching SPEL + LEZ stable releases, the LEZ release prerequisite, the `lez-template` / `spel-template` decision, and CI guardrails for SPEL/LEZ alignment + pin reachability) is tracked separately in #170.

## Evidence

- `refs/tags/v0.2.0-rc.5` points to tag object `67a589f3ae5299f126a4e8cbf192d65c4189bb3f`.
- `refs/tags/v0.2.0-rc.5^{}` peels to commit `1db7c5f8af3165318f9046c13ab185e24773fd59`.
- `ed3bbedb4b684645da05455d30a4a0be7cc4dfe0` is the merge commit for `logos-co/spel#164`, not reachable from a public branch/tag head.

## Verification

- `cargo fmt --check`
- `cargo test default_spel_pin_is_public_rc5_peeled_commit`
- `cargo test --lib` — 334 passed
- `cargo check`
- Built `lgs` and confirmed `lgs init` writes `repos.spel.pin = "1db7c5f8af3165318f9046c13ab185e24773fd59"` into a generated `scaffold.toml`.

### DIFF
--- a/FURPS.md
+++ b/FURPS.md
@@ -53,7 +53,7 @@
 - Logos Core DevEx for overall developer journey alignment and terminology.
 - Logos Blockchain and Logos Execution Environment for functionality.
 - Wallet Module for interactions with Logos Execution Environment.
-- `logos-co/spel` CLI — vendored per project at a pinned commit (`DEFAULT_SPEL_PIN`, currently tag `v0.2.0`); supplies the `spel inspect` output that `deploy` parses for the program ID.
+- `logos-co/spel` CLI — vendored per project at a pinned commit (`DEFAULT_SPEL_PIN`, currently the public commit peeled from `v0.2.0-rc.5`); supplies the `spel inspect` output that `deploy` parses for the program ID.
 
 #### Runtime Dependencies
 
--- a/src/commands/doctor.rs
+++ b/src/commands/doctor.rs
@@ -6,7 +6,7 @@ use anyhow::bail;
 use super::wallet_support::wallet_password;
 use crate::commands::wallet_support::WALLET_CONFIG_PRIMARY;
 use crate::constants::{
-    DEFAULT_LEZ, DEFAULT_SPEL, SEQUENCER_BIN_REL_PATH, SPEL_BIN_REL_PATH, WALLET_BIN_REL_PATH,
+    DEFAULT_LEZ, DEFAULT_SPEL_PIN, SEQUENCER_BIN_REL_PATH, SPEL_BIN_REL_PATH, WALLET_BIN_REL_PATH,
 };
 use crate::doctor_checks::{
     check_binary, check_container_runtime, check_path, check_port_warn, check_repo,
@@ -117,7 +117,7 @@ pub(crate) fn build_doctor_report() -> DynResult<DoctorReport> {
     rows.push(check_repo("spel", &spel, &project.config.spel.pin));
 
     rows.push(CheckRow {
-        status: if project.config.spel.pin == DEFAULT_SPEL.sha {
+        status: if project.config.spel.pin == DEFAULT_SPEL_PIN {
             CheckStatus::Pass
         } else {
             CheckStatus::Warn
@@ -125,14 +125,14 @@ pub(crate) fn build_doctor_report() -> DynResult<DoctorReport> {
         name: "spel pin matches scaffold default".to_string(),
         detail: format!(
             "configured pin={} expected={}",
-            project.config.spel.pin, DEFAULT_SPEL.sha
+            project.config.spel.pin, DEFAULT_SPEL_PIN
         ),
-        remediation: if project.config.spel.pin == DEFAULT_SPEL.sha {
+        remediation: if project.config.spel.pin == DEFAULT_SPEL_PIN {
             None
         } else {
             Some(format!(
                 "Set repos.spel.pin in scaffold.toml to {} and run `{}`",
-                DEFAULT_SPEL.sha, STEP_SETUP
+                DEFAULT_SPEL_PIN, STEP_SETUP
             ))
         },
     });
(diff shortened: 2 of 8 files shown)


weboko:
Why pining to RC instead of a release? I think we want, and at least for Logos modules, agreed on using only releases instead of RC.

wdyt @vpavlin @fryorcraken ?

danisharora099:
> Why pining to RC instead of a release? I think we want, and at least for Logos modules, agreed on using only releases instead of RC.
> 
> wdyt @vpavlin @fryorcraken ?

Good question. What I am fixing here is the SHA. The old `DEFAULT_SPEL.sha` pointed at a GitHub PR merge commit, which is not advertised by the public `logos-co/spel` branch or tag refs. This updates it to the public commit that the existing `v0.2.0-rc.5` tag resolves to.

I agree with the releases over RC policy, but I would keep that as a separate coordinated `spel` + LEZ bump. The stable `spel` tags currently vendor different LEZ refs from scaffold's `DEFAULT_LEZ`. For example, the latest `spel` release is `v0.3.0` with LEZ `v0.2.0-rc3`, while scaffold is still on `v0.2.0-rc1`, so bumping `spel` by itself probably needs a bit more testing.

This PR is meant to be the narrower fix: keep the current `spel` version, but point at the public tag commit instead of the PR merge commit

weboko:
Appreciated that this PR is solving a real and concrete issue by replacing the non-public PR merge SHA with the actual public tag commit reference. That said, I still strongly believe we should avoid RC versions entirely and move toward proper stable releases only, even if that broader coordination between spel and LEZ happens separately from this narrower fix.

danisharora099:
> Appreciated that this PR is solving a real and concrete issue by replacing the non-public PR merge SHA with the actual public tag commit reference. That said, I still strongly believe we should avoid RC versions entirely and move toward proper stable releases only, even if that broader coordination between spel and LEZ happens separately from this narrower fix.

Agreed! We should work towards a v0.2.0, and also a plan of future releases

Opened an issue here: https://github.com/logos-co/scaffold/issues/170

fryorcraken:
Yes, I agree we should use release moving forward.
However, the PR that bumps the release should:
1. Bump SPEL and LEZ to matching releases - note LEZ doesn't have a release (yet)
2. It should fix up the `lez-template` -> and review the situation on whether it should be called `spel-tempalte` or if the concept should simply be scratched
3. It should add a CI job to ensure that (1) and (2) do not get broken in the future

So I would say this current PR is fine to be limited scope.

Finally, as tag and commit hash may change in the future (a repo maintainer can always change a tag, and that's why nix uses tag), a better change would be to remove the tag all together, and just use commit hash.

danisharora099:
Updated this to keep the PR scoped to the broken SPel public pin, but now scaffold stores the SPel default as a commit-only pin rather than carrying the mutable `v0.2.0-rc.5` tag alongside it.

I also updated #170's body to track the broader release migration: matching SPEL + LEZ releases, the LEZ release prerequisite, the `lez-template`/`spel-template` decision, and CI guardrails.

weboko:
@danisharora099 need to sign commit

FURPS.md
```diff
@@ -53,7 +53,7 @@
 - Logos Core DevEx for overall developer journey alignment and terminology.
 - Logos Blockchain and Logos Execution Environment for functionality.
 - Wallet Module for interactions with Logos Execution Environment.
-- `logos-co/spel` CLI — vendored per project at a pinned commit (`DEFAULT_SPEL_PIN`, currently tag `v0.2.0`); supplies the `spel inspect` output that `deploy` parses for the program ID.
+- `logos-co/spel` CLI — vendored per project at a pinned commit (`DEFAULT_SPEL_PIN`, currently the public commit peeled from `v0.2.0-rc.5`); supplies the `spel inspect` output that `deploy` parses for the program ID.
```
weboko: why can't we https://github.com/logos-co/spel/releases/tag/v0.4.0 ? are you depending on something in 0.2?

fryorcraken (approved):

PR #180, opened 2026-05-26
chore: address follow ups to pr#169

weboko:
This is follow up to https://github.com/logos-co/scaffold/pull/169

### DIFF
--- a/src/constants.rs
+++ b/src/constants.rs
@@ -18,24 +18,20 @@ pub(crate) struct GitRef {
     pub(crate) tag: &'static str,
 }
 
-// Cross-framework invariant: DEFAULT_SPEL_PIN must point at a spel commit
+// Cross-framework invariant: DEFAULT_SPEL must point at a spel commit
 // whose `spel-cli/Cargo.toml` vendors LEZ at the same ref as DEFAULT_LEZ.
 // Otherwise spel's sequencer-RPC client speaks a different protocol than
 // scaffold's own wallet/sequencer build. `check_spel_lez_alignment` in
 // `commands/doctor.rs` enforces this at runtime — re-run `doctor` after
 // bumping either pin.
-//
-// Special note on DEFAULT_SPEL_PIN: this is the public commit that the
-// `v0.2.0-rc.5` tag peels to on `logos-co/spel`. rc.5 is the spel release
-// whose vendored LEZ matches DEFAULT_LEZ.tag. We deliberately store the
-// commit only (not the tag) so the default cannot drift if the upstream
-// tag is ever moved or replaced. Migrating off RC-derived defaults is
-// tracked in <https://github.com/logos-co/scaffold/issues/170>.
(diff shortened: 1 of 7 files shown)


src/constants.rs
```diff
@@ -18,24 +18,20 @@ pub(crate) struct GitRef {
...
-// tracked in <https://github.com/logos-co/scaffold/issues/170>.
 pub(crate) const DEFAULT_LEZ: GitRef = GitRef {
     sha: "35d8df0d031315219f94d1546ceb862b0e5b208f",
     tag: "v0.2.0-rc1",
 };
-pub(crate) const DEFAULT_SPEL_PIN: &str = "1db7c5f8af3165318f9046c13ab185e24773fd59";
+pub(crate) const DEFAULT_SPEL: GitRef = GitRef {
+    sha: "1db7c5f8af3165318f9046c13ab185e24773fd59",
+    tag: "v0.2.0-rc.5",
+};
```
fryorcraken, line 40: This is only valid at a given point in time. Upstream maintainer may change teh tag to point to a different commit at any time. Waht is the value of trackign the tag here?

fryorcraken (commented):
I don't agree with this PR.

src/constants.rs
weboko: human readability as it goes to the logs in CLI later and much easier to notice when debugging than a random hash

and I agree with the tag can change, that is why we have hash

src/constants.rs
```diff
@@ -18,24 +18,20 @@ pub(crate) struct GitRef {
...
 // `commands/doctor.rs` enforces this at runtime — re-run `doctor` after
 // bumping either pin.
-//
-// Special note on DEFAULT_SPEL_PIN: this is the public commit that the
-// `v0.2.0-rc.5` tag peels to on `logos-co/spel`. rc.5 is the spel release
-// whose vendored LEZ matches DEFAULT_LEZ.tag. We deliberately store the
-// commit only (not the tag) so the default cannot drift if the upstream
-// tag is ever moved or replaced. Migrating off RC-derived defaults is
-// tracked in <https://github.com/logos-co/scaffold/issues/170>.
 pub(crate) const DEFAULT_LEZ: GitRef = GitRef {
```
weboko, line 32:
```suggestion
//
// Special note on DEFAULT_SPEL: this is the public commit that the
// `v0.2.0-rc.5` tag peels to on `logos-co/spel`. We deliberately use the
// commit so the default cannot drift if the upstream
// tag is ever moved or replaced.
pub(crate) const DEFAULT_LEZ: GitRef = GitRef {
```

src/constants.rs
fryorcraken: Ok so purely a log thing... sure

weboko: yes, so are you against improving it?

weboko:
lmk if there is strong opposition, we can roll back this PR anytime

@fryorcraken @danisharora099

PR #183, opened 2026-06-01
fix(basecamp): honor `attr` across setup/install/launch

danisharora099:
## Problem

The Logos toolchain ships in `dev` and `portable`. Both of them work differently: dev is for a local-development build and `portable` is for a distrubution-ready build.

Scaffold defaults everything to dev. So even when a project specifically tries to build for portable, the output is a dev-build, thus overriding project defaults.

## Solution

Treat `[repos.basecamp].attr` as the single source of truth. When it names a portable output (`bin-macos-app`, `bin-appimage`, `bin-bundle-dir`), scaffold auto-picks the matching `lgpm`, module attr, variant key, and XDG subpath.

Default dev behaviour is unchanged. Explicit per-layer overrides still win.

## Notes

- `cargo test --all-targets` passes — 348 lib tests (+8 new for the portable helpers) + 159 integration tests.
- Unblocks deleting a ~60-line manual workaround in [eth-lez-atomic-swaps](https://github.com/logos-co/eth-lez-atomic-swaps/blob/master/scripts/basecamp-instance.sh#L113-L178).

### DIFF
--- a/src/commands/basecamp.rs
+++ b/src/commands/basecamp.rs
@@ -12,9 +12,10 @@ use anyhow::{bail, Context};
 use crate::config::{default_basecamp_repo, default_lgpm_repo};
 use crate::constants::{
     BASECAMP_ATTR, BASECAMP_AUTODISCOVER_SKIP_SUBDIRS, BASECAMP_DEPENDENCIES,
-    BASECAMP_PREINSTALLED_MODULES, BASECAMP_PROFILES_REL, BASECAMP_PROFILE_ALICE,
-    BASECAMP_PROFILE_BOB, BASECAMP_SOURCE, BASECAMP_XDG_APP_SUBPATH, DEFAULT_BASECAMP_PIN,
-    DEFAULT_LGPM_PIN, LGPM_ATTR, LGPM_SOURCE,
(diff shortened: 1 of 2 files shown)


src/commands/basecamp.rs
```diff
@@ -213,6 +220,37 @@ fn format_flake_ref(repo: &RepoRef) -> String {
...
+        ("aarch64-linux", true) => Some("linux-arm64"),
+        ("aarch64-darwin", false) => Some("darwin-arm64-dev"),
+        ("aarch64-darwin", true) => Some("darwin-arm64"),
+        ("x86_64-darwin", false) => Some("darwin-amd64-dev"),
+        ("x86_64-darwin", true) => Some("darwin-amd64"),
+        _ => None,
+    }
+}
+
 fn build_basecamp_app(project_root: &Path, repo: &Path, out_dir: &Path) -> DynResult<PathBuf> {
```
weboko: I think co-pilot left you a comment about it already:

`build_basecamp_app` ignores `attr` while`setup builds` a `dev` binary while seeding portable dirs

danisharora099:
yeah, looks like the PR missed fixing the basecamp binary build derivation, but added for others
should be fixed now - added

weboko (approved):

PR #185, opened 2026-06-01
fix(run): use wallet resolved sequencer URL in run summary

### LINKED ISSUE #161: report: Sequencer line in run summary uses raw localnet.port rather than wallet's resolved sequencer_addr
## Summary

The post-deploy summary line `Sequencer: http://127.0.0.1:<port>` in `cmd_run`'s no-hooks branch (and the analogous output in `print_deploy_summary`) renders `project.config.localnet.port` directly. The address the wallet (and therefore `cmd_deploy`) actually targets comes from the wallet runtime config (`.scaffold/wallet/wallet_config.json` → `sequencer_addr`), with the localnet port only as a fallback.

When the two diverge — wallet repointed at a remote sequencer, port overridden by env, etc. — the summary line is misleading.

## Repro

Edit `.scaffold/wallet/wallet_config.json` to point `sequencer_addr` at a non-default host. Run `lgs run` with no post-deploy hooks. Summary still prints `http://127.0.0.1:<localnet.port>`.

## Suggested fix

Use the resolution path that `run_state::resolved_sequencer_addr` already implements: load the wallet config, fall back to `DEFAULT_SEQUENCER_ADDR`. Same call site as the deploy-cache config digest, so this also keeps the user-visible message in sync with the cache key.

## Carried from

Flagged in PR #97's dogfooding, again in PR #98's (https://github.com/fryorcraken/cryptarchs-scaffold-develop/blob/scaffold-develop/docs/dogfood-pr98-2026-05-13.md, Issue 4). Cosmetic — out of scope for the run-command stack but worth tracking.

Sertug17:
## Problem

`lgs run` prints a summary line:

    Sequencer: http://127.0.0.1:<port>

The port is read directly from `project.config.localnet.port` (scaffold.toml).
When `wallet_config.json` sets a different `sequencer_addr`, the displayed URL
is wrong the wallet is talking to a different endpoint than what the summary
shows.

Fixes #161.

## Fix

Replace the raw port construction in `print_run_summary` with a call to
`default_sequencer_http_url_for_project()` the same helper already used
by `build_hook_command` and `wallet_support`. This keeps the summary
consistent with the address the wallet actually targets.

## Test

All 56 existing run tests pass. No new test added the function under
test is a single-line display helper with no branching logic.

### DIFF
--- a/src/commands/run.rs
+++ b/src/commands/run.rs
@@ -540,9 +540,13 @@ fn print_deploy_summary(project: &Project) -> DynResult<()> {
         }
     }
 
-    let port = project.config.localnet.port;
+    // Use the same URL construction as build_hook_command and wallet_support
+    // so the summary always reflects the address the wallet actually targets,
+    // rather than the raw localnet.port value which may differ when
+    // wallet_config.json overrides sequencer_addr (issue #161).
+    let sequencer_url = crate::commands::wallet_support::default_sequencer_http_url_for_project(project);
     println!();
-    println!("Sequencer: http://127.0.0.1:{port}");
+    println!("Sequencer: {sequencer_url}");
 
     Ok(())
 }
@@ -552,8 +556,7 @@ fn build_hook_command(
     hook_command: &str,
     deployed: &DeployedPrograms,
 ) -> Command {
-    let port = project.config.localnet.port;
-    let sequencer_url = format!("http://127.0.0.1:{port}");
+    let sequencer_url = crate::commands::wallet_support::default_sequencer_http_url_for_project(project);
     let wallet_home = project
         .root
         .join(&project.config.wallet_home_dir)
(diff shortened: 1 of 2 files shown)


src/commands/run.rs
weboko: looks like here we need to update it too + on the return of `lgs localnet status`

weboko (changes requested):
thanks for the PR
added small comment

Sertug17:
Good catch! Fixed in the latest commit also updated `build_hook_command` 
to use `default_sequencer_http_url_for_project()` consistently.

Regarding `lgs localnet status`: `localnet_addr` is constructed directly 
from `localnet.port` in `cmd_localnet_in_project` and passed as a parameter 
 since localnet always binds to localhost this should be fine, but happy 
to update it too if you'd prefer consistency.

Sertug17:
Also updated `lgs localnet status` listener lines now show the full 
`http://` URL for consistency.

Sertug17:
@weboko addressed your feedback ready for re review!

weboko (approved):
LGTM

PR #186, opened 2026-06-01
fix(run): warn when vendored spel binary is absent before extracting program IDs

### LINKED ISSUE #160: post-deploy: SCAFFOLD_PROGRAM_ID absent when vendored spel not built
## Summary

`extract_program_id` shells out to `spel inspect` to fill `SCAFFOLD_PROGRAM_ID` / `SCAFFOLD_PROGRAM_ID_<name>` for post-deploy hooks. When the vendored spel hasn't been built (fresh project, or `lgs setup` hasn't completed), `spel inspect` returns `program_id: unavailable (run logos-scaffold setup to build the vendored spel)` and the env var is correctly left unset.

This is defensible behavior: hooks using `set -u` will fail loudly, which is what we want. But it's a confusing first-run experience for users who already ran `lgs setup` and then hit a transient case where `spel inspect` can't read the binary (different machine pin, stale cache, etc.).

## Repro

Fresh-clone a project, run `lgs run` with a post-deploy hook that uses `$SCAFFOLD_PROGRAM_ID`. Hook fails with `SCAFFOLD_PROGRAM_ID: unbound variable` (under `set -u`) or silently runs against an empty id (without). The deploy itself succeeded.

## Context

- Run.rs (handles the None gracefully): https://github.com/logos-co/scaffold/blob/master/src/commands/run.rs
- Deploy.rs `extract_program_id`: returns `Option<String>`; None means \"couldn't get id\".
- Unit test `hook_program_id_unset_when_extraction_failed` codifies the current contract.

## Possible directions

1. Detect the \"unavailable\" condition during step 3 (ensure_localnet) or step 5 (deploy) and bail with a clear setup hint, rather than silently propagating None to post-deploy.
2. Make hooks emit a diagnostic line when `extract_program_id` returns None (currently silent).
3. Document the precondition in the run command --help.

## Carried from

Flagged in PR #66's dogfooding pass, again in PR #97's, again in PR #98's (https://github.com/fryorcraken/cryptarchs-scaffold-develop/blob/scaffold-develop/docs/dogfood-pr98-2026-05-13.md, Issue 3). Out of scope for the run-command stack but worth tracking.

Sertug17:
## Problem

When `lgs setup` has not been run (or spel failed to build), the vendored
spel binary does not exist. `extract_program_id()` calls
`Command::new(spel_bin).spawn().ok()?` which silently returns `None` on
any spawn failure including a missing binary.

The result: `SCAFFOLD_PROGRAM_ID` is silently unset for every program,
and post-deploy hooks fail with no actionable error message.

Fixes #160.

## Fix

Before the per-program loop in `collect_deployed_programs`, check
`spel_bin.is_file()`. If the binary is missing, emit a warning on stderr
that names the missing path and tells the user to run `lgs setup`.

The existing behaviour for a present-but-failing spel binary is unchanged.

## Test

All 56 existing run tests pass. The warning path is not exercised by
existing tests a missing spel binary in the test environment would
require a setup fixture outside the scope of this fix.

### DIFF
--- a/src/commands/run.rs
+++ b/src/commands/run.rs
@@ -383,6 +383,17 @@ fn collect_deployed_programs(project: &Project, skipped: bool) -> DynResult<Depl
         .context("failed to resolve spel repo path for run post-deploy env")?;
     let spel_bin = spel_repo.join(SPEL_BIN_REL_PATH);
 
+    // Warn early when the vendored spel binary has not been built yet.
+    // extract_program_id() spawns spel_bin and returns None on any failure,
+    // so a missing binary silently leaves SCAFFOLD_PROGRAM_ID unset — which
+    // causes post-deploy hooks to fail in confusing ways (issue #160).
+    if !spel_bin.is_file() {
+        eprintln!(
+            "warning: vendored spel binary not found at {}.              SCAFFOLD_PROGRAM_ID will be unset for all programs.              Run `lgs setup` to build spel.",
+            spel_bin.display()
+        );
+    }
+
     let mut out = Vec::new();
     for stem in programs {
         let Some(bin_path) = binaries.get(&stem).cloned() else {


weboko (approved):
lgtm! thanks!

PR #188, opened 2026-06-01
feat(wallet): add --json output to `wallet list` and `wallet topup`

### LINKED ISSUE #23: feat: add --json output to wallet commands
## Problem

`wallet topup` and `wallet list` produce human-readable text output only. There is no `--json` flag on either command, unlike `localnet status --json` and `doctor --json` which already support structured output.

This makes programmatic consumption difficult. Consumers that call scaffold from application code must parse stdout strings and infer error types from stderr text (e.g. checking for "already initialized" or "connection refused" substrings).

There is also no first-class `wallet balance` command. Checking balances requires the passthrough: `logos-scaffold wallet -- account ls -l`.

**Source references:**
- `src/commands/wallet.rs:17-31` -- `WalletAction` enum has no `--json` on List or Topup
- `src/commands/wallet.rs:46-66` -- `cmd_wallet_list` only accepts `long: bool`
- `src/commands/wallet.rs:90-226` -- `cmd_wallet_topup` prints unstructured text
- No `Balance` variant in `WalletAction`

## Suggestion

1. Add `--json` to `wallet list` and `wallet topup` for machine-readable output.
2. Consider a `wallet balance [address]` subcommand that returns the balance in a structured format, removing the need for the passthrough workaround.

weboko:
### Problem / Description
`wallet list` and `wallet topup` print human-readable text only. Application code that drives scaffold has to scrape stdout and infer error types from stderr substrings (`"already initialized"`, `"connection refused"`, …). `localnet status --json` and `doctor --json` already set the structured-output precedent; wallet didn't follow it.

### Solution
Add `--json` to both commands:

- **`wallet list --json`** — emits an envelope `{ command, exit_code, accounts, stdout, stderr }`, so consumers read `exit_code` instead of guessing success from text while still keeping the raw wallet output.
- **`wallet topup --json`** — emits a structured outcome with a stable `status` (`success` | `dry_run` | `pending` | `error`) and, on failure, a categorized `reason` (`connectivity` | `preflight` | `init` | `failed`), plus `address`, `method`, `network`, and `tx`. Command echoes are suppressed in JSON mode so stdout is a single clean object; the process still exits non-zero on failure/timeout.

Plain-text output is unchanged when `--json` is omitted. Adds 4 integration tests and `--help` examples.

The optional `wallet balance` subcommand floated in the issue is intentionally left as a follow-up — it needs a parse of the inner wallet's balance format that can't be exercised against a stub here.

Closes #23.

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -410,6 +410,9 @@ enum WalletSubcommand {
 struct WalletListArgs {
     #[arg(long)]
     long: bool,
+    /// Emit accounts as a JSON object instead of forwarding the wallet's text.
+    #[arg(long)]
+    json: bool,
 }
 
 #[derive(Debug, clap::Args)]
@@ -420,6 +423,9 @@ struct WalletTopupArgs {
     address_flag: Option<String>,
     #[arg(long)]
     dry_run: bool,
+    /// Emit the topup outcome as a JSON object instead of human-readable text.
+    #[arg(long)]
+    json: bool,
 }
 
 #[derive(Debug, clap::Args)]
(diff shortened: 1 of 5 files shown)


danisharora099 (approved):
noice

PR #190, opened 2026-06-01
feat(idl): hash-based short-circuit for `build idl`

### LINKED ISSUE #134: feat(idl): hash-based short-circuit for `build idl` so cache-hit `lgs run` collapses to seconds
## Motivation

Surfaced while dogfooding PR #97 (`feat/run-deploy-cache`). With the deploy cache in place, a no-change `lgs run` against a real Phase-1 project takes ~35s on my machine. ~30s of that is step `[2/5] Building IDL`, which calls `cargo test --workspace __lssa_idl_print --show-output --quiet` unconditionally and re-prints an IDL whose inputs have not changed.

The deploy cache buys back the actual on-chain deploy (a few seconds against a local sequencer); the methods rebuild only triggers if guest source actually changed. Most of what is left in the re-run wall time is the IDL step re-emitting bytes that already exist on disk.

If `build_idl_for_current_project` gains a hash-based "unchanged inputs" short-circuit, the re-run cost of `lgs run` collapses from ~35s to seconds.

## Proposed approach

Mirror what `commands::run_state` already does for the deploy step:

1. On entry to `build_idl_for_current_project`, compute a SHA-256 over the same guest-source set that produces the IDL output (or, more conservatively: the guest source set + the version of any IDL-print macro that emits the blocks).
2. Compare against a cache file at `.scaffold/state/idl_build.json` (same `.scaffold/state/` neighbourhood as `run_deploy.json`).
3. If the hash matches and all expected `<stem>.json` files still exist in `framework.idl.path`, skip the `cargo test __lssa_idl_print` invocation and print a `[2/N] IDL up-to-date (skipped)` line.
4. Otherwise run the existing pipeline and write the new hash.

The plumbing in `run_state.rs` (`compute_program_hashes`, `save_state`/`load_state`, tempfile + rename atomic writes, malformed-cache fallback) is the right shape to reuse — extract a small generic over the state filename or just copy the pattern.

## Caveats / non-goals

- This is independent of PR #97 and should not block it. #97 makes the bottleneck visible; this issue fixes the bottleneck.
- The IDL step is already a no-op for non-`lez-framework` projects, so the cache only matters for the framework path.
- A `--reset` should clear the IDL cache alongside `run_deploy.json` (same pattern PR #99 already implements for the deploy cache).
- Cache-key must include whatever proc-macro/codegen version is responsible for producing the IDL blocks, otherwise a `cargo update` that changes the macro silently keeps the stale IDL. The simplest safe fingerprint is the macro crate's pin from `Cargo.lock`.

## Acceptance

- [ ] A no-change `lgs run` against a real Phase-1 project drops to < 5s wall on a warm machine.
- [ ] `lgs run --reset` clears both the deploy cache and the IDL cache.
- [ ] Editing a guest source file that affects the IDL re-runs the IDL step.
- [ ] Bumping the IDL-emitting macro's pin re-runs the IDL step even if guest sources are byte-identical.

## Background

- Originating dogfood writeup: `cryptarchs-scaffold-develop/docs/dogfood-pr97-2026-05-11.md`, "Observation 2 / Feedback item 1".
- Pattern to copy from: `src/commands/run_state.rs` on `feat/run-deploy-cache` (PR #97).

weboko:
### Problem / Description
With the deploy cache in place, a no-change `lgs run` against a real Phase-1 project spends ~30s of its ~35s in step `[2/5] Building IDL`, which calls `cargo test --workspace __lssa_idl_print` unconditionally and re-prints an IDL whose inputs haven't changed.

### Solution
Add a hash-based short-circuit to `build_idl_for_current_project`, mirroring the deploy cache:

- Fingerprint the IDL inputs into a SHA-256 stored at `.scaffold/state/idl_build.json`: every project `.rs` source (excluding `target/`, `.scaffold/`, and other build/vendor dirs), `Cargo.lock` (which pins the IDL-emitting proc-macro, so a `cargo update` busts the cache even when guest sources are byte-identical), and the `[framework.idl]` config. The state also records the output `.json` file names.
- On entry, when the fingerprint matches the prior build **and** every recorded output still exists on disk, skip the `cargo test` invocation and print `IDL up-to-date (skipped)`.
- Explicit `lgs build idl` always regenerates (force); the `build` / `run` pipelines use the cache.
- `lgs run --reset` clears the IDL cache (before the IDL step, so it re-runs that invocation) alongside the deploy cache.
- Atomic tempfile+rename writes and malformed-cache fallback, same as `run_deploy.json`.

### Acceptance mapping
- no-change `lgs run` collapses (IDL step skipped) ✔
- `lgs run --reset` clears both caches ✔
- editing a guest source re-runs IDL (covered by `idl_hash_changes_when_a_source_file_changes`) ✔
- bumping the IDL macro's pin re-runs even with identical sources, via Cargo.lock (`idl_hash_changes_when_cargo_lock_changes`) ✔

### Tests / verification
7 unit tests cover hash stability, change detection (source + Cargo.lock), `target/`+`.scaffold/` exclusion, state round-trip, output-presence, and malformed-cache fallback. The full `cargo test __lssa_idl_print` path needs a real lez-framework workspace + circuits release (not reproducible in CI here), so the cache decision logic is unit-tested directly. The framework-gate behavior is unchanged (`build_idl_fails_loudly_on_default_framework` still green).

Closes #134.

### DIFF
--- a/src/commands/idl.rs
+++ b/src/commands/idl.rs
@@ -1,11 +1,15 @@
 use std::fs;
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
 use std::process::Command;
 
-use anyhow::{anyhow, bail};
+use anyhow::{anyhow, bail, Context};
+use serde::{Deserialize, Serialize};
+use sha2::{Digest, Sha256};
+use walkdir::WalkDir;
 
 use crate::circuits::ensure_circuits_for_subprocess;
 use crate::constants::FRAMEWORK_KIND_LEZ_FRAMEWORK;
+use crate::model::Project;
 use crate::process::run_capture;
 use crate::project::{load_project, resolve_cache_root, run_in_project_dir};
 use crate::state::write_text;
@@ -15,6 +19,21 @@ const IDL_BEGIN_PREFIX: &str = "--- LSSA IDL BEGIN ";
 const IDL_END_PREFIX: &str = "--- LSSA IDL END ";
 const IDL_MARKER_SUFFIX: &str = " ---";
 
+/// Cache file recording the fingerprint of the inputs that produced the IDL,
+/// alongside the list of output `.json` files written. Lives in the same
+/// `.scaffold/state/` neighbourhood as `run_deploy.json`.
+pub(crate) const IDL_STATE_REL: &str = ".scaffold/state/idl_build.json";
+
+/// Persisted IDL-build fingerprint. `input_hash` covers every project `.rs`
+/// source, `Cargo.lock` (which pins the IDL-emitting proc-macro), and the
+/// `[framework.idl]` config; `outputs` are the file names written so a cache
+/// hit can verify they still exist before skipping the rebuild.
+#[derive(Debug, Default, Serialize, Deserialize)]
+struct IdlBuildState {
+    input_hash: String,
+    outputs: Vec<String>,
+}
+
 pub(crate) fn cmd_idl(args: &[String]) -> DynResult<()> {
     if args.is_empty() {
         bail!("usage: logos-scaffold build idl [project-path]");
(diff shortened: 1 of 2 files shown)


danisharora099 (approved):
LGTM!

One thing I couldn’t verify from here: there’s no E2E in CI that actually runs `lgs run` twice on a real lez project and checks that step 2 skips. The unit tests cover the cache logic, but the expensive `cargo test __lssa_idl_print` path still needs a manual pass.

Did/could you dogfood on a lez project - back-to-back `lgs run` with no changes (second should skip IDL and be much faster), plus `run --reset`, a guest .rs edit, and `build idl` always forcing a rebuild? A few log lines is enough; doesn’t need to block merge.

weboko:
@danisharora099 yes, seemed to be working, didn't do twice on a project though

danisharora099:
> @danisharora099 yes, seemed to be working, didn't do twice on a project though

were you able to validate that caching works as expected?

PR #191, opened 2026-06-01
feat(run): `[run.watch]` include/exclude globs + `--watch-debounce-ms`

### LINKED ISSUE #175: lgs run --watch: [run.watch] include/exclude globs + --watch-debounce-ms CLI flag
## TL;DR

`lgs run --watch` re-runs the full pipeline on any filesystem change with a fixed 500ms debounce. Two small additions would make it usable on real projects: glob-based include/exclude filtering so non-input edits don't trigger re-deploys, and a CLI knob for the debounce window.

Surfaced while planning `eth-lez-atomic-swaps`'s adoption of `lgs run`. Originating tracker entry: [`docs/scaffold-upstream-tracker.md#tr-09`](https://github.com/logos-co/eth-lez-atomic-swaps/blob/master/docs/scaffold-upstream-tracker.md#tr-09).

## Background

The current default — coalesce a flurry of editor saves into one re-run — is right. The missing piece is *what to watch*. On a typical Phase-1 project with `risc0` guest binaries:

- Editing `src/` orchestrator Rust shouldn't trigger guest re-deploys (the guest ELF / image ID hash is invariant under those edits).
- Editing `programs/<chain>/methods/guest/` *should* trigger a full re-run (guest source changes → new ELF → new image ID → must re-deploy).
- Editing `contracts/`, `docs/`, `README.md`, `*.lock`, etc. shouldn't trigger anything.

Without filtering, a hot orchestrator-side edit triggers a full pipeline re-run on every save, which (with the deploy cache from #97 / IDL cache from #134 still merging) can still be tens of seconds of work.

This is **indirect** pain for `eth-lez-atomic-swaps` today (we haven't adopted `lgs run` yet — see umbrella [#172](https://github.com/logos-co/scaffold/issues/172) for the blocker on the `lgs run` adoption side). Filing now so the shape exists when we get there.

### `[run.watch]` glob filters

```toml
[run.watch]
include = ["programs/**/guest/**", "contracts/**/*.sol"]
exclude = ["**/*.md", "**/target/**", "**/.scaffold/**", "Cargo.lock"]
```

Resolution: a path triggers a re-run iff it matches at least one `include` glob (or `include` is unset, meaning "any path") AND matches zero `exclude` globs. `exclude` always wins over `include`. Patterns are project-relative `gitignore`-style globs.

If both are unset, current behaviour (watch everything) is preserved.

### `--watch-debounce-ms <N>` CLI flag

```
$ lgs run --watch --watch-debounce-ms 1500
```

Per-invocation override of the default 500ms. Also allow it as `[run.watch].debounce_ms = 1500` for project-level persistence.

## Acceptance criteria

- [ ] `[run.watch]` schema accepted with `include`, `exclude`, optional `debounce_ms`.
- [ ] `--watch-debounce-ms` CLI flag lands.
- [ ] Resolution rules documented (exclude wins, project-relative glob style).
- [ ] Default behaviour unchanged when `[run.watch]` is absent.

## Downstream impact

Indirect for `eth-lez-atomic-swaps` today — the project doesn't use `lgs run` yet (waits on [#172](https://github.com/logos-co/scaffold/issues/172)). Once adopted, edits to `src/` (orchestrator Rust) shouldn't trigger guest-program re-deploys; scoping watch to `programs/lez-htlc/**` plus `contracts/**` is the obvious config.

## See also

- [`logos-co/scaffold#170`](https://github.com/logos-co/scaffold/issues/170) — release-pipeline context.
- [`logos-co/scaffold#172`](https://github.com/logos-co/scaffold/issues/172) — `lgs run` pipeline extensions umbrella; `--watch` becomes more interesting once the full pipeline is `lgs run`-driven.
- [`logos-co/scaffold#134`](https://github.com/logos-co/scaffold/issues/134) — IDL hash-based short-circuit; the deploy/IDL caches plus watch-filtering together make hot re-runs feel instant.
- Originating tracker: [`eth-lez-atomic-swaps/docs/scaffold-upstream-tracker.md#tr-09`](https://github.com/logos-co/eth-lez-atomic-swaps/blob/master/docs/scaffold-upstream-tracker.md#tr-09).

weboko:
### Problem / Description
`lgs run --watch` re-runs the full pipeline on any filesystem change with a fixed 500ms debounce. On a real project a hot orchestrator-side edit (or a `README.md` save) triggers a full re-deploy, and there's no way to scope what's watched or tune the coalescing window.

### Solution
Two additions, both opt-in (default `--watch` behaviour is unchanged when `[run.watch]` is absent):

**`[run.watch]` glob filters**
```toml
[run.watch]
include = ["programs/**/guest/**", "contracts/**/*.sol"]
exclude = ["**/*.md", "Cargo.lock"]
debounce_ms = 1500
```
A changed path triggers a re-run **iff** it matches at least one `include` glob (or `include` is unset → "any path") **and** matches zero `exclude` globs — `exclude` always wins. Globs are project-relative, gitignore-style: `**` spans path segments, `*`/`?` match within a segment, and a slash-less pattern (`Cargo.lock`) matches at any depth. The built-in ignores (`.scaffold`, `target`, `.git`, the IDL output dir) still always apply regardless of these filters.

**`--watch-debounce-ms <MS>`** overrides the debounce per invocation; `[run.watch].debounce_ms` persists it project-wide. Resolution: CLI flag > config > 500ms default.

### Implementation notes
- The glob matcher is dependency-free (a small recursive segment matcher) to keep the published crate's dependency tree minimal.
- `[run.watch]` parses, validates (`debounce_ms` must be a non-negative integer), and round-trips through serialize so config rewrites preserve it.

### Acceptance mapping
- `[run.watch]` schema with `include` / `exclude` / `debounce_ms` ✔
- `--watch-debounce-ms` flag ✔
- resolution rules documented (README + `WatchConfig` doc comment + `--help`) ✔
- default behaviour unchanged when `[run.watch]` absent ✔

### Tests
10 unit tests: glob `**`/`*`/`?`/slash-less semantics, the include/exclude/exclude-wins resolution, built-in ignores layered with filters, `[run.watch]` parse + negative-debounce rejection + parse/serialize round-trip. The watch loop itself (filesystem events) isn't integration-tested — the testable filter/debounce decisions are unit-tested directly.

Closes #175.

### DIFF
--- a/README.md
+++ b/README.md
@@ -222,6 +222,29 @@ lgs run --no-post-deploy                                 # skip all hooks
 `--post-deploy` and `--no-post-deploy` conflict with each other and
 both override whatever `[run].post_deploy` defines.
 
+#### Watch mode
+
+`lgs run --watch` re-runs the pipeline on each filesystem change (localnet
+is reused; reset is skipped on re-runs). Scope what counts as a change with
+`[run.watch]` and tune the coalescing window:
+
+```toml
+[run.watch]
+include = ["programs/**/guest/**", "contracts/**/*.sol"]
+exclude = ["**/*.md", "Cargo.lock"]
+debounce_ms = 1500
+```
+
+A changed path triggers a re-run **iff** it matches at least one `include`
+glob (or `include` is unset, meaning "any path") **and** matches zero
+`exclude` globs — `exclude` always wins. Globs are project-relative,
+gitignore-style: `**` spans path segments, `*`/`?` match within a segment,
+and a slash-less pattern (`Cargo.lock`) matches at any depth. `.scaffold`,
+`target`, `.git`, and the IDL output dir are always ignored regardless of
+these filters. Override the debounce per invocation with
+`lgs run --watch --watch-debounce-ms 1500` (CLI wins over
+`[run.watch].debounce_ms`, which wins over the 500ms default).
+
 Checkpoint commands:
 
 ```bash
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -315,7 +315,8 @@ struct RunArgs {
     watch: bool,
     /// Override the `--watch` debounce window (milliseconds) for this
     /// invocation. Defaults to `[run.watch].debounce_ms`, else 500.
-    #[arg(long, value_name = "MS")]
+    /// Only meaningful with `--watch`, so it requires it.
+    #[arg(long, value_name = "MS", requires = "watch")]
     watch_debounce_ms: Option<u64>,
 }
 
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -313,6 +313,10 @@ struct RunArgs {
     /// Localnet is reused; reset is skipped on re-runs.
     #[arg(long)]
     watch: bool,
+    /// Override the `--watch` debounce window (milliseconds) for this
+    /// invocation. Defaults to `[run.watch].debounce_ms`, else 500.
+    #[arg(long, value_name = "MS")]
+    watch_debounce_ms: Option<u64>,
 }
 
 #[derive(Debug, clap::Args)]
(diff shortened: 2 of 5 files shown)


danisharora099 (approved):
Nice one. We should also update the README/doc to make the default path be `run` based, instead of separate steps and include the `--watch` flag as well

PR #209, opened 2026-06-06
refactor: remove AI-generated slop across modules

weboko:
## What

A module-by-module sweep of this Rust CLI to remove AI-generated slop. All edits are minimal and behavior-preserving, **except one real bug fix** (`setup.rs`). Verified with `cargo check` (clean, no warnings) and `cargo test --no-run` (all targets compile).

### Bug fix
- **`src/commands/setup.rs` — `sha256_hex` never computed SHA-256.** `sha2` is a plain, non-optional dependency and the crate has **no `[features]` table**, so `#[cfg(feature = "sha2")]` was *always false* and the FNV-1a fallback was what compiled. The function returned a 16-char non-SHA digest, so prebuilt `sequencer_service` verification (`setup.rs:348`) compared it against a published 64-char `.sha256` and **always mismatched → `bail!`**, silently breaking the `--prebuilt` path whenever a checksum file existed. Now uses `sha2` unconditionally, matching the project's existing hex-encode idiom (`run_state.rs::hex_encode`).

### Slop cleanup (no behavior change)
- **`src/config.rs`** — removed a dead `let path = format!("modules.{name}");` binding that was only consumed by `let _ = path;` (plus its "Defensive:" comment).
- **`src/commands/basecamp.rs`** — removed `#[allow(dead_code)] // Fields wired up in later phases` on `BasecampAction` (all variants are constructed; compiles clean without it); collapsed a verbose two-arm `cwd_override` match into `as_deref().unwrap_or(project_root)`, matching the idiom the file's own tests already use.
- **`src/commands/run.rs`** — stripped redundant `// Step N:` comment prefixes that just duplicate the user-facing `[N/total_steps]` progress prints; all explanatory text preserved.
- **`src/commands/wallet_support.rs`** — collapsed a stray run of spaces in the legacy wallet-config warning string.

## Notes on what was *not* changed
The codebase is otherwise well-maintained. Several flagged "redundant" defensive checks (e.g. `.exists()` after a non-empty-path guard in `basecamp.rs`) were verified to be **legitimate** — a non-empty path string doesn't guarantee the binary is on disk — and were intentionally left as-is. Idiomatic nested `if let` optional-chaining (no let-chains on rustc 1.81) and the well-documented non-cryptographic `fnv1a_64` helper were also left untouched.

https://claude.ai/code/session_01KutAyMX3auZUKhuHHp5xXq

### DIFF
--- a/src/commands/basecamp.rs
+++ b/src/commands/basecamp.rs
@@ -28,7 +28,6 @@ use crate::state::{read_basecamp_state, write_basecamp_state};
 use crate::DynResult;
 
 #[derive(Debug, Clone)]
-#[allow(dead_code)] // Fields wired up in later phases
 pub(crate) enum BasecampAction {
     Setup,
     Modules {
@@ -1084,14 +1083,7 @@ fn run_build_portable_nix(
 ) -> DynResult<Vec<PathBuf>> {
     println!("building {flake_ref}");
     let mut cmd = Command::new("nix");
-    match &inv.cwd_override {
-        Some(cwd) => {
-            cmd.current_dir(cwd);
-        }
-        None => {
-            cmd.current_dir(project_root);
-        }
-    }
+    cmd.current_dir(inv.cwd_override.as_deref().unwrap_or(project_root));
     for a in &inv.args {
         cmd.arg(a);
     }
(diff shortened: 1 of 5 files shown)


weboko:
closing in favor of https://github.com/logos-co/scaffold/pull/210

PR #210, opened 2026-06-07
refactor: continue scaffold desloping

wozos:
## Summary

Continuation of the scaffold desloping pass from the existing branch/PR, moved to a `wozos`-owned branch so the Pi can keep iterating.

This includes:
- the existing cleanup commits from the current deslop branch;
- one additional mechanical cleanup pass trimming clippy-style slop and needless allocations/borrows.

## Verification

- `cargo test`


weboko (commented):
## Review — desloping refactor: behavior-preserving except one hidden functional change

Reviewed the full diff (verified the load-bearing extractions line by line), built + ran the suite, and dogfooded `new`/`create`/`doctor` against the built binary.

### Verified locally
- `cargo build` + `cargo clippy --all-targets` clean; **456 lib + 168 CLI tests pass**.
- `new` and `create` succeed and install skills; `doctor` renders all rows correctly.
- Extraction faithfulness spot-checked: `hash::hex_encode` is byte-identical to the deleted copies; `commands::sanitize_separated` reproduces both old `to_cargo_crate_name` (`-`/`program_deployment`) and `sanitize_file_stem` (`_`/`program`) exactly; `template::copy_dir_recursive`, `run_checked`→`run_forwarded` alias, and the `succeeded()`/`summarize_command_failure` helpers are all equivalent.

This is a clean, well-scoped cleanup. One item, though, is **not** behavior-preserving and should be surfaced rather than folded silently into a "mechanical" PR:

### ⚠️ Hidden functional change — `sha256_hex` (src/hash.rs ← src/commands/setup.rs)
The old `sha256_hex` was gated:
```rust
#[cfg(feature = "sha2")]      { real SHA-256 }
#[cfg(not(feature = "sha2"))] { FNV-1a → format!("{h:016x}") }
```
There is **no `[features]` table** in `Cargo.toml` and `sha2` is a plain (non-optional) dependency, so the feature `sha2` never existed — the `#[cfg(feature = "sha2")]` arm was **always compiled out** (this is exactly the `unexpected cfg condition value: sha2` warning on the base branch). The **live path was the 16-char FNV-1a fallback**. That means the `--prebuilt` checksum check at `setup.rs:233-236` compared a 16-char FNV digest against a published 64-char `.sha256`, so it could **never match** and would always bail `SHA256 mismatch`.

The new `hash::sha256_hex` computes a real SHA-256 unconditionally — which actually **fixes** prebuilt checksum verification. That's a good fix, but:
1. It's a functional change riding inside a refactor PR; please call it out in the PR description.
2. There's **no test** covering prebuilt checksum verification (match and mismatch). Since this path was silently broken before and no one noticed, a small unit test on `sha256_hex` against a known vector + a mismatch-path test would lock it down.

### Minor
- **deploy.rs:~286** — the sequencer-reachability warning moved `println!`→`eprintln!` (stdout→stderr). Reasonable for keeping stdout JSON-clean, but it's a stream change; confirm nothing captures that line on stdout.
- A few warning strings had their literal multi-space runs collapsed to single spaces (localnet r0vm warning, run.rs vendored-spel warning, wallet_support legacy/missing-config). Cosmetic; fine unless a test asserts exact text.
- **Not mergeable** (`dirty`) against current `master` — needs a rebase (master advanced to `#234`). 83 commits is also a lot; consider squashing on merge.

Not approving — leaving as comments. Solid cleanup overall; the sha256 item is the only thing I'd want addressed (or at least documented) before merge.

weboko (commented):
## Independent re-review — desloping refactor: behavior-preserving, with one beneficial hidden change to surface

Re-reviewed on a fresh environment (Rust + risc0 3.0.6 + Nix + Docker). Confirms the prior review's conclusions and adds a data-flow analysis of the one non-mechanical change.

### Verified locally
- **`cargo build` + `cargo clippy --all-targets` clean; 456 lib tests pass.**
- **CLI integration tests**: 164 pass; the only 4 failures (`localnet_start_*`, `localnet_reset_*`) are **environmental, not caused by this PR** — they exercise `localnet start`, which downloads the `logos-blockchain-circuits` release, and the download fails in my sandbox (the GitHub `302 → S3` redirect is blocked by the egress proxy). **The same 4 tests fail identically on `master`**, so this is not a regression.
- **Dogfooded `new` / `create` / `doctor` against the built binary**: `lgs new myapp` and `lgs create app2` both clone + pin LEZ, render the `default` template, and install AI skills (`.claude/skills/`, `.cursor/rules/`, `AGENTS.md`) correctly. `doctor` renders every row, categorizes PASS/WARN/FAIL, exits non-zero on FAIL, and `doctor --json` emits valid JSON with exactly `checks`, `next_steps`, `status`, `summary`. Behavior matches `master`.
- **Extraction faithfulness spot-checked**: `hash::hex_encode` (covered by its new unit test), and `commands::sanitize_separated(x, '-', "program_deployment")` is line-for-line equivalent to the old `to_cargo_crate_name` (same alnum-lowercasing, same consecutive-separator collapsing via `prev_sep`, same trim + fallback). `testnode/blocks.rs::sha256_hex` is **byte-identical** to `master` (still real SHA-256, unchanged).

### ⚠️ The one non-mechanical change — `sha256_hex` — is real, and it is a *latent-bug fix* (please surface it explicitly)
Agree this must not be folded silently into a "mechanical" PR. Full trace of the impact:
- On `master`, `sha256_hex` was gated `#[cfg(feature = "sha2")]` / `#[cfg(not(...))]`, but there is **no `[features]` table** and `sha2` is a plain dependency, so the feature never existed and the **FNV-1a fallback (16 hex chars) was always the compiled arm** — it never produced a real SHA-256.
- The only production caller is the prebuilt `sequencer_service` verifier in `setup.rs`, where `expected` is read from a **`{url}.sha256` checksum file published alongside the release binary** (i.e. a genuine 64-char SHA-256). So on `master`, whenever that checksum file exists, `actual` (16-char FNV) can **never** equal `expected` (64-char SHA-256) → the check would **always `bail!("SHA256 mismatch ...")`** and abort `--prebuilt` setup. The check only appeared to work because it's guarded by `if let Ok(sha_resp)` / `if !expected.is_empty()` and no checksum file is currently published, leaving it dormant.
- This PR's new `hash::sha256_hex` computes **real SHA-256**, which makes the verifier actually correct. So the change is a **net improvement**, not a hazard — but it does alter runtime behavior and belongs in the PR description / commit message rather than under "desloping".

### Recommendation
Clean, well-scoped, ~1k-line net reduction with tests and dogfooding green. The single behavioral change is beneficial (fixes a dormant, broken integrity check). I'd be comfortable approving once the `sha256_hex` FNV-1a → SHA-256 change is explicitly called out in the PR/commit description so it isn't lost as "mechanical."

weboko:
## CI is red for reasons that predate this PR — both legs are stale-base failures fixed by #224

Following up on my previous review. I did **not** re-review the code (no new commits since `1caba2d`), but the CI state changed meaning after my last pass, so recording the diagnosis: **both red `render-and-build` legs are inherited from this PR's stale merge base, not caused by the refactor.**

### Why it can't be this PR

This PR changes **38 files, +791/−1035, and touches zero files under `templates/`** — nothing under `templates/default/` or `templates/lez-framework/`. Yet both failures are *template compile* failures. Its merge base is `1ca6783` (#219), which predates #224 `ci(template-e2e): make rendered template builds pass end to end`, merged to master on **2026-07-20T16:23Z** — about 13 hours *after* my last review here, which is why neither of us saw this connection.

### Leg 1 — `render-and-build (default)`

```
error[E0277]: the trait bound `AccountId: From<(&[u32; 8], &PdaSeed)>` is not satisfied
error: could not compile `e2e-app` (bin "run_hello_world_with_authorization_through_tail_call_with_pda")
```

#224 changed **exactly that line**, in exactly that file:

```diff
-    let pda = AccountId::from((&program.id(), &PDA_SEED));
+    let pda = AccountId::for_public_pda(&program.id(), &PDA_SEED);
```

### Leg 2 — `render-and-build (lez-framework)`

```
note: there are multiple different versions of crate `nssa_core` in the dependency graph
  --> .../lssa-07dfb2b61ac10c3c/767b5af/nssa/core/src/account.rs:27:1     (expected)
  ::: .../logos-execution-zone-6bae42d7c9cadfe7/cf3639d/nssa/core/src/account.rs:125:1  (found)
```

Two *different* git sources supply `nssa_core`: the old `lssa` repo at `767b5af`, pulled in transitively by `jimmy-claw/lez-framework`, and `logos-execution-zone` at `cf3639d`, from this template's own `rev = "{{lez_pin}}"` workspace pin. Cargo cannot unify them, so the guest gets two incompatible `AccountWithMetadata` types.

#224 replaced that entire stack (`lez-framework` → `spel-framework`) and switched the workspace deps from `rev =` to `tag =`, with a comment naming this exact hazard:

```toml
# Spel pins LEZ by tag. Match that source form so Cargo resolves a single
# nssa_core instead of treating equivalent tag and rev sources as distinct.
nssa_core = { git = "https://github.com/logos-blockchain/logos-execution-zone.git", tag = "{{lez_tag}}" }
```

### Action items

1. **Rebase/merge onto current master.** GitHub also now reports this PR `mergeable_state: dirty`, so it needs this regardless. After that both legs should go green — the red CI should not be read as "this refactor breaks the templates."
2. **Still outstanding from my previous review:** the `sha256_hex` FNV-1a → real SHA-256 change needs to be called out explicitly in the PR/commit description. It is a genuine behavior change (and a latent-bug fix for the `--prebuilt` checksum verifier), not a mechanical desloping change, and it should not be lost under that heading.

Comment only, not an approval.

---

weboko:
## Status update — the rebase I recommended for CI is now also a merge-conflict rebase (12 source files)

Following up on my 2026-07-29 comment. Still **no new commits** since `1caba2d`, so again this is not a re-review of the diff — but the merge state has degraded since that comment and it changes the size of the "just rebase it" ask I made, so recording it.

On 2026-07-29 the story was: CI is red purely from a stale merge base, rebase past `#224` and both `render-and-build` legs go green. That is still true, but `master` has since moved to `9fcc376` and the branch no longer merges:

```
CONFLICT (content): Merge conflict in src/circuits.rs
CONFLICT (content): Merge conflict in src/cli.rs
CONFLICT (content): Merge conflict in src/commands/basecamp.rs
CONFLICT (content): Merge conflict in src/commands/build.rs
CONFLICT (content): Merge conflict in src/commands/client.rs
CONFLICT (content): Merge conflict in src/commands/doctor.rs
CONFLICT (content): Merge conflict in src/commands/run.rs
CONFLICT (content): Merge conflict in src/commands/setup.rs
CONFLICT (content): Merge conflict in src/config.rs
CONFLICT (content): Merge conflict in src/lib.rs
CONFLICT (content): Merge conflict in src/process.rs
CONFLICT (content): Merge conflict in src/template/project.rs
```

Verified locally with `git merge-tree 9fcc376 1caba2d` — 12 files, all content conflicts, no modify/delete.

### Why it got this bad

Eight commits landed on those same files since this branch's merge base (`1ca6783`):

| Commit | What it touched |
|---|---|
| `#221` | `circuits` as a project-level dependency — `src/circuits.rs`, `config.rs` |
| `#222`, `#223` | basecamp module build/run + multi-peer profiles — `src/commands/basecamp.rs` |
| `#229` | fast block-time test-node API |
| `#234` | nested LEZ repo layout |
| `#238` | absolute `LOGOS_DATA_DIR` on macOS portable launch |
| `#239` | per-profile `deploy = false` — `src/commands/run.rs` |
| `#241` | dogfooding critical path (cache reuse, deploy pacing, spel program-id, runner auth, guest-build CC) — broad |

This PR is `+791/−1035` across 38 files and is a *desloping* refactor — it moves and rewrites code rather than adding to it, which is the worst possible shape for sitting behind six weeks of active development on the same files. The conflict count is only going to grow.

### What I'd suggest

I don't think "rebase and resolve 12 files" is a reasonable ask in one go for a refactor of this shape, and a mechanical resolution risks silently reverting behaviour that landed in `#221`/`#239`/`#241` — the deleted-side of a desloping diff looks a lot like code you'd want to drop. Two options that seem better than a single big rebase:

1. **Split it.** The refactor is largely independent per-module. Landing it as a series of small PRs rebased on current `master` (one for `circuits.rs`/`config.rs`, one for `basecamp.rs`, one for `process.rs`/`lib.rs`, …) would each be reviewable and each merge in hours rather than weeks. Given the file list, most of these would not conflict with each other.
2. **Rebase now and re-verify against the dogfooding path**, not just the test suite. `#241` specifically repaired the dogfooding critical path; if the resolution drops any of it the tests may still pass while D1/D7 break. If you go this route I'll re-run D1 and D7 end-to-end on the rebased head before signing off.

Either way the rebase is now blocking, and it needs a human who knows the refactor's intent doing the resolution rather than a mechanical `--theirs`/`--ours`.

---

weboko:
## Follow-up review — the `sha256_hex` fix is right, but the verification around it is fail-open (new finding)

No new commits since `1caba2d`, so this is not a re-review of the diff. I picked the PR back up because `master` has moved again (now `7523ee2`), and while re-checking the one behavioural change in this PR — `sha256_hex` FNV-1a → real SHA-256 — I found something the previous passes (2026-07-03, 2026-07-20) did not raise. Flagging it here because it sits on a supply-chain path and this PR is the change that makes it matter.

### The finding

`try_prebuilt_download` verifies the downloaded `sequencer_service` only when *two* conditions hold, and **both fail open**:

```rust
// src/commands/setup.rs (this branch, L222-241)
let sha_url = format!("{url}.sha256");
if let Ok(sha_resp) = ureq::get(&sha_url).call() {   // (1) any fetch error -> skip, silently
    ...
    if !expected.is_empty() {                        // (2) empty sidecar   -> skip, silently
        let actual = sha256_hex(&bytes);
        if actual != expected { anyhow::bail!("SHA256 mismatch ..."); }
        println!("SHA256 verified: {actual}");
    }
}

std::fs::write(&dest, &bytes)?;                      // reached in ALL cases
#[cfg(unix)]
{ std::fs::set_permissions(&dest, Permissions::from_mode(0o755))?; }
```

If the `.sha256` sidecar request fails for **any** reason — 404, 5xx, timeout, DNS, an intercepting egress proxy — or if it returns empty/whitespace, control falls straight through to `fs::write` + `chmod 0755`. The unverified bytes are installed as an executable and later spawned as the project's sequencer. Nothing is printed on either skip path.

### Why this PR is the right place to raise it

On `master` the check was inert: `sha256_hex` was the dead-`cfg` FNV-1a arm (I re-confirmed this is still true on `7523ee2` — no `[features]` table, `sha2` is a plain dependency, so `#[cfg(feature = "sha2")]` never compiles), so a 16-char FNV digest could never equal a 64-char published SHA-256. Verification could only ever `bail!`, so nobody could depend on it.

This PR makes the comparison genuinely correct — that is a real improvement and I agree with the earlier assessment that it is a latent-bug fix. But it also promotes `SHA256 verified: <hex>` from noise to a signal people will read, **while leaving the skip-on-failure structure untouched**. Silent absence of a security check is worse once the check is believed to work.

Threat model, stated honestly so this isn't overweighted: the binary comes from GitHub releases over HTTPS, so TLS covers transit and this is not a trivially exploitable hole. The residual exposure is anyone positioned to fail the sidecar request while letting the binary request through (a corporate/CI egress proxy, a caching mirror) — and, far more mundanely, **a release where the `.sha256` simply was never uploaded**, which is today's state. In that case `--prebuilt` installs an unverified executable and says nothing.

### Suggested resolution

Make the policy explicit rather than incidental. Either:

**(a) Require the sidecar** — treat fetch failure or an empty digest as "cannot verify", and fall back to the source build instead of installing unverified bytes:

```rust
let sha_resp = ureq::get(&sha_url).call().map_err(|e| anyhow!(
    "prebuilt {tag}: checksum file unavailable ({e}); refusing to install unverified binary"
))?;
```

**(b) Keep it optional but make the skip loud** — on every skip path:

```rust
eprintln!("warning: no published checksum for {tag}; installing prebuilt sequencer_service WITHOUT integrity verification");
```

I'd argue for **(a)**, since the artifact is executed rather than merely read, and the source build is always available as a fallback. (b) is defensible if you want `--prebuilt` to keep working before checksums are published — but then the absence of verification has to be as visible as its presence.

### Test coverage — the 2026-07-03 ask is still open

- `src/hash.rs` in this branch tests `hex_encode` but **not** `sha256_hex`. The known-vector + mismatch test requested on 2026-07-03 is still absent.
- `grep -rn prebuilt tests/` returns **nothing** — no test exercises the match, mismatch, or either skip path.

Since this code path was silently broken on `master` for its entire life and nobody noticed, that is exactly the argument for locking it down with a test now.

### Merge state — degraded again since 2026-08-04

`master` has moved `9fcc376` → `7523ee2` and the conflict set has grown from 12 files to **16** (`git merge-tree 7523ee2 1caba2d`):

```
src/circuits.rs        src/commands/deploy.rs   src/commands/setup.rs      src/lib.rs
src/cli.rs             src/commands/doctor.rs   src/commands/wallet.rs     src/process.rs
src/commands/basecamp.rs  src/commands/run.rs   src/commands/wallet_support.rs  src/state.rs
src/commands/build.rs  src/commands/client.rs   src/config.rs              src/template/project.rs
```

New since the last count: `deploy.rs`, `wallet.rs`, `wallet_support.rs`, `state.rs`. A 1035-line-deletion refactor rebased across 16 conflicting files is where behaviour-preserving cleanups stop being behaviour-preserving — the earlier reviews' line-by-line extraction-faithfulness checks will need redoing after the rebase, not just re-running the tests.

### Where this leaves it

The refactor itself is good and I still think it should land. Before it does, I'd want: the fail-open verification addressed (or an explicit decision to keep it, in writing), the `sha256_hex` test added, the FNV→SHA-256 behavioural change called out in the PR description as previously asked, and the rebase.

---

weboko:
## Addendum to my earlier comment today — `--prebuilt` is unreachable in practice, and its URL test is tautological

Still no new commits since `1caba2d`, and `master` is still `7523ee2`, so nothing in my earlier comment changes. Two small things I turned up while verifying the fail-open finding on real infrastructure — both are new, both are cheap to fix, and the first one reframes how urgent the fail-open issue is.

### 1. The prebuilt path can't currently fire at all — and the URL is pre-rename

I went looking for the published `.sha256` sidecar to see which skip branch actually triggers. There is no sidecar because **there is no release**:

```
GET /repos/logos-co/scaffold/releases          -> 200, count: 0
GET /repos/logos-co/logos-scaffold             -> 301 Moved Permanently
```

The URL the code builds still names the **pre-rename** repo (`src/commands/setup.rs:451` on `master`, `:210` on this branch):

```rust
"https://github.com/logos-co/logos-scaffold/releases/download/{tag}/sequencer_service"
```

I fetched both spellings of the real download URL for the currently pinned LEZ commit:

```
.../logos-co/logos-scaffold/releases/download/lssa-prebuilt-cf3639d8-x86_64-linux/sequencer_service  -> 404
.../logos-co/scaffold/releases/download/lssa-prebuilt-cf3639d8-x86_64-linux/sequencer_service        -> 404  (following redirects)
```

So `--prebuilt` always takes the `Err(Status(404, _))` arm, prints *"no prebuilt published yet"*, and falls back to the source build. That's a useful correction to the risk framing in my earlier comment: the fail-open write is **latent, not live** — nobody is installing unverified binaries today, because nobody is installing prebuilt binaries at all.

That cuts both ways, though. It means (a) there's no urgency, so the fix can be done properly rather than hastily, and (b) the hole opens the moment someone publishes the first prebuilt release — which is exactly when attention will be on getting the artifact out, not on the verifier. Fixing it before the first release is a lot easier than remembering to at release time. It also means the whole feature has never been executed end to end, so the stale repo name has had no way to surface.

### 2. `prebuilt_url_contains_tag_and_binary` cannot catch either problem

```rust
// src/commands/setup.rs:376-383 (this branch)
#[test]
fn prebuilt_url_contains_tag_and_binary() {
    let tag = "lssa-prebuilt-35d8df0d-x86_64-linux";
    let url = format!(
        "https://github.com/logos-co/logos-scaffold/releases/download/{tag}/sequencer_service"
    );
    assert!(url.contains("lssa-prebuilt-35d8df0d"));
    assert!(url.contains("sequencer_service"));
}
```

The test **rebuilds the URL inside itself** and then asserts against its own literal. It never calls production code, so it is true by construction — it would still pass if the production URL pointed at `example.com`. `prebuilt_tag_format` (`:403`) has the same shape: it re-derives `&pin[..8]` in the test body rather than exercising the function that does it.

This is the same "no test exercises the prebuilt path" gap I raised earlier, but worse than absent — it reads like coverage in a `grep`, which is probably why the stale repo name survived the rename. Worth extracting a `fn prebuilt_url(pin: &str, arch: &str, os: &str) -> String` and pointing both tests at it; that gets you a real regression test for the rename and a natural seam for the sidecar-policy tests I asked for.

### Standalone verification of this branch

Since the branch can't be merged right now, I built and tested it on its own base to confirm the refactor is sound independent of the rebase. On a from-scratch environment (Rust 1.94.1, Nix 2.35.2, risc0 3.0.6 — r0vm/cargo-risczero 3.0.6, guest toolchain 1.97.0, C++ 2024.1.5):

- `cargo build --release` — clean.
- `cargo test --locked --release` — **629 green** (456 lib + 168 CLI + 5 doc), zero failures.
- Concrete confirmation the dead `cfg` is really gone: building `master` emits `warning: unexpected cfg condition value: sha2` twice (`setup.rs:507`, `setup.rs:517`). On this branch those warnings are **absent** — `sha256_hex` now lives in `src/hash.rs` with an unconditional `Sha256::digest`. That's the compiler agreeing that `master`'s verification arm never compiled.

### Where this leaves it — unchanged

Everything in my earlier comment still stands: good refactor, should land, but wants the fail-open decision written down, a real `sha256_hex` test, the FNV→SHA-256 change noted in the PR description, and the 16-file rebase. I'd add the stale repo name and the two tautological tests to that list — they're small, and the rebase is the natural time to do them.

---

weboko:
## Re-verified on the new base (`834b6b0`) — I can now name the exact commit that makes CI red, and I audited the two security-sensitive hunks

No new commits since `1caba2d`, so this is not a re-review of the diff; my two comments from 2026-08-17 (the fail-open `sha256_hex` verification, and `--prebuilt` being unreachable with a tautological URL test) stand unchanged and are still the blockers. I picked this back up because `master` moved `7523ee2` → `834b6b0` (`#251`, `#252`). Merge state is identical in shape — still 16 content conflicts, `src/error.rs` is **not** among them (this branch never touches it), so `#251`'s new `Next:` guidance on `ExitedBeforeReady` / `StartTimeout` survives the merge intact.

Two things are new below: a hard root-cause for the red CI (previously I asserted "stale base"; now I can point at the commit and the line), and an audit of the two hunks in this refactor that could plausibly leak data, which I had not done before.

### The red CI is stale-base — proven, not inferred

Both legs of `render-and-build` die in code `master` has since fixed, in the single commit `7c52211` (`#224`).

**`render-and-build (default)`** fails at:

```
error[E0277]: the trait bound `AccountId: From<(&[u32; 8], &PdaSeed)>` is not satisfied
  --> src/bin/run_hello_world_with_authorization_through_tail_call_with_pda.rs:35:15
```

That file is repo-owned (`templates/default/src/bin/`), not LEZ-owned, and `#224` is the commit that fixed exactly that line:

```diff
- let pda = AccountId::from((&program.id(), &PDA_SEED));
+ let pda = AccountId::for_public_pda(&program.id(), &PDA_SEED);
```

**`render-and-build (lez-framework)`** fails at `#[lez_program]` in `methods/guest/src/bin/lez_counter.rs` (`E0282`/`E0308`/`E0425`) — and `#224` migrated that same file, plus `src/lib.rs` and the four manifests, to the `spel_framework` API.

The LEZ pin is byte-identical on both sides (`cf3639d8` at your merge base `1ca6783b` and at `834b6b0`), which rules out a pin-drift explanation. So: **nothing in this diff causes the red, and a rebase is sufficient to clear it** — no code changes needed for CI. That confirms what I said on 2026-07-29, now with the receipt.

### One rebase trap in the workflow, worth calling out before you resolve it

This branch adds an `Install RISC Zero Rust toolchain` step (`.github/workflows/template-e2e.yml`, before the cargo cache) that runs `rzup install rust`. `#224` later added its own step doing the same job **plus `rzup install cpp`**, with a comment explaining why:

```
# Guest deps that compile C sources (e.g. `ring` in the lez-framework
# guest) additionally need risc0's C++ cross-toolchain. When it is
# missing, risc0-build points CC at a sentinel path
# (/no_risc0_cpp_toolchain_installed_run_rzup_install_cpp) and the
# guest build dies in cc-rs compiler detection.
```

Because `git diff master..HEAD` renders master's newer step as a deletion, a mechanical resolution can easily keep *this* branch's step and drop master's — which silently removes `rzup install cpp` and re-breaks the `lez-framework` leg in a way that looks unrelated to the resolution. **Take master's step verbatim and drop the one this branch adds** (it is strictly a subset), along with master's `HOST_CC`/`HOST_CXX` comment on the build step.

### Security / privacy audit of the refactor — both hunks are behaviour-preserving

This is a 791/1035 refactor across 38 files, and two of them handle data that must not leak. I diffed them line by line against the merge base rather than trusting the "no functional change" framing. Both check out, and I want that on the record since it is the part of this PR that would be expensive to get wrong:

**`src/commands/report.rs` (diagnostics sanitisation).** The redaction pipeline is untouched in substance:

- `collect_sanitized_json_artifact` dropping its `DynResult<()>` return is safe — every path in the old body returned `Ok`, so the `?` at both call sites was a no-op. Skip-on-high-risk-markers, skip-on-sanitize-error, and skip-on-write-error all still `return` before `collected.push(...)`, so a file that fails sanitisation still never enters the archive.
- `collect_tool_versions`: `!result.succeeded()` (`model.rs:265` — `error.is_none() && status == Some(0)`) is exactly the old `error.is_some() || status.unwrap_or(1) != 0`, including the "no status ⇒ not success" case. The docker/podman loop is a faithful merge of the two duplicated blocks.
- `tail_file_lines_lossy`: dropping `buf.clear()` in favour of `std::mem::take(&mut buf)` is correct — `take` leaves `Vec::new()`, so each iteration still starts empty. No line bleeding into the next, which is the failure mode that would have mattered here (it would splice log content across the tail boundary).
- `scrub_manifest_text` → `scrub_path_string` is a pure alias removal; every manifest field (`path`, `source`, `notes`, `reason`, warnings) is still scrubbed.

**`src/commands/wallet_support.rs` (wallet home / tx parsing).** Also clean:

- The new `WalletRuntimeContext::command()` reproduces the old call sites byte for byte, including the `as_os_str().to_string_lossy().to_string()` conversion (I checked `deploy.rs:139-143` and `wallet.rs:157-161`/`183-187` at the merge base — same expression). All four converted call sites get `NSSA_WALLET_HOME_DIR`; no site lost it.
- `summarize_command_failure`: `stderr.lines().chain(stdout.lines()).find(non-empty)` is equivalent to the old stderr-then-stdout cascade, because the first non-empty line of the concatenation *is* stderr's when stderr has one.
- `extract_tx_hash_from_hash_type_bytes`: `collect::<Option<Vec<u8>>>()` short-circuits on the first unparseable chunk, matching the old `parse_failed`/`break`, and `.filter(|b| !b.is_empty())` preserves the empty-input rejection.
- `src/state.rs` (`push_str(&format!(..))` → `writeln!`) and the new `src/hash.rs` are likewise value-identical; `check_state_value` still guards CR/LF injection into the line-oriented state files before any write.

So the desloping itself is not where the risk is. The two open findings from 2026-08-17 are.

### Cross-PR note for whoever sequences these

This branch moves `mod logged_tests` from `src/process.rs:403` to `src/process.rs:586` without changing its contents. `#34` adds a new test (`daemonized_process_is_own_session_leader`) *inside* that module at the old location. They will conflict textually whichever lands second, though both changes survive a correct resolution — just don't let the moved module swallow the added test.

### Where this leaves it

Unchanged from 2026-08-17: not mergeable as-is. In order —

1. The fail-open verification around `sha256_hex` (my 2026-08-17 comment) — the only correctness blocker in the diff itself.
2. `--prebuilt` unreachable in practice + its tautological URL test (same day, addendum).
3. Rebase onto `834b6b0` (16 conflicts), taking `#224`'s risc0 step and dropping this branch's, then a fresh `render-and-build` — which should go green on the rebase alone.

---

weboko:
just don't need it for now

PR #220, opened 2026-06-16
docs(dogfooding): real-sequencer provisioning, T4 real-tx scenario, agent directives

weboko:
## Summary

The test-node command group and the public Rust API (#219) are merged. This is a **docs-only** follow-up that brings `DOGFOODING.md` in sync with what a *real* end-to-end dogfood run actually requires, and adds the one capstone scenario that was missing.

It is grounded in an actual real-sequencer run: I installed the risc0 toolchain, built the real LEZ `sequencer_service` + `wallet`, started real test-nodes, and submitted a real wallet transaction — then wrote down exactly how, so the next run (human or agent) doesn't stall at "this environment can't run a real sequencer."

## What changed

- **Agent Execution Directives** (new, top of file): a binding preamble for automated runs — do the *real* thing (a real sequencer executing real transactions, not a stub), install every dependency yourself, push through installer/proxy failures with the `curl` + `GH_TOKEN` fallbacks, background long builds so they survive turns (not `nohup &`), and reuse the provisioned toolchain.
- **Provisioning the Real LEZ Sequencer Toolchain** (new section): step-by-step install of
  - `r0vm` matched to the LEZ `risc0-zkvm` pin (read from `Cargo.lock`), via `rzup` or — when rzup's rustls rejects a TLS-intercepting proxy CA (`InvalidCertificate(UnknownIssuer)`) — a `curl` fallback that fetches the risc0 release tarball and drops `r0vm` at scaffold's expected `~/.risc0/extensions/v<ver>-cargo-risczero-<triple>/` path;
  - the circuits release (auto-fetched by `test-node prepare`);
  - the real `sequencer_service` (`test-node prepare`, ~6 min);
  - the `wallet` (`setup`) — including the two `setup` gotchas: its circuits precheck doesn't consult the cache (so `LOGOS_BLOCKCHAIN_CIRCUITS` must be exported first), and it uses cwd discovery (must run inside the project).
- **T1–T3 refreshed**: now require a real node and document the concrete real-output signals — the per-account clock cadence (`/0000001` tracks head, `/0000010`/`/0000050` lag), the exact seeded balance read back from a real node, and real borsh block parsing.
- **T4 (new): Real Committed User Transaction (test-node + wallet)** — `wallet topup` executed by a real test-node sequencer (auth-transfer init + pinata claim via r0vm in dev mode), observed as a real user-tx block (`transaction_count: 2`, a `public` user tx distinct from the clock tx) through `blocks wait` / `tx wait`. The capstone a stub cannot prove.
- **Scenario Index** + **Minimum Rerun Guidance** updated for T4 and the provisioning prerequisite.

## Why

During the original feature dogfood, the run stopped at "no risc0 toolchain here" and fell back to a faithful JSON-RPC stub. The toolchain *could* be installed (it now is, and persists in the container cache); the real run subsequently passed end-to-end and confirmed the hand-rolled borsh parsers match genuine sequencer output. This change encodes that so it isn't re-learned next time.

Docs-only — no code changes.

Follow-up to #219.

### DIFF
--- a/DOGFOODING.md
+++ b/DOGFOODING.md
@@ -117,7 +117,7 @@ mkdir -p "$EXT" && cp /tmp/cr/r0vm "$EXT/r0vm" && chmod +x "$EXT/r0vm"
 "$SCAFFOLD_BIN" test-node doctor  --project "$P" --json | jq .ok   # → true (all checks pass)
 ```
 
-**4. (Only for real transactions — T4) build the wallet via `setup`.** Two gotchas distinguish `setup` from `test-node prepare`: its circuits precheck does **not** consult the scaffold cache, so `LOGOS_BLOCKCHAIN_CIRCUITS` must be exported first; and it uses cwd discovery, so it must run **inside** the project (no `--project` flag):
+**4. (Only for real transactions — T4) build the wallet via `setup`.** Two gotchas distinguish `setup` from `test-node prepare`: its circuits precheck does **not** consult the scaffold cache (it only accepts `LOGOS_BLOCKCHAIN_CIRCUITS` or `$HOME/.logos-blockchain-circuits/`), so ensure one of those is present before running it; and it uses cwd discovery, so it must run **inside** the project (no `--project` flag):
 
 ```bash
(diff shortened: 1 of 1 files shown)


weboko:
follow up to the merged feature https://github.com/logos-co/scaffold/pull/219

Issue #174, opened 2026-05-22
umbrella: lgs basecamp verb granularity — build [--variant --module], run <module>


danisharora099:
## TL;DR

Three holes in `lgs basecamp`'s verb surface force every consumer project to drop back to raw `nix build` / `nix run` for the module build + dev-iteration workflow:

1. No `lgs basecamp build` (default `#lgx` variant, no install side-effect).
2. No per-module or per-variant filter on the existing `build-portable`.
3. No `lgs basecamp run <module>` to launch a UI module in `logos-standalone-app` (the dev smoke-test host).

Adding these three closes the "everything goes through `lgs`" loop for the module/UI dev cycle. Surfaced while bringing up `eth-lez-atomic-swaps` against scaffold master. Full per-ask context in [`docs/scaffold-upstream-tracker.md`](https://github.com/logos-co/eth-lez-atomic-swaps/blob/master/docs/scaffold-upstream-tracker.md) entries TR-10, TR-14, TR-15.

## Today's build matrix

| Verb | Builds | Installs? | Per-module filter? | Per-variant filter? |
|---|---|---|---|---|
| `lgs basecamp install` | both variants of every module | yes (all modules, all roles) | no | no |
| `lgs basecamp build-portable` | `#lgx-portable` of every `role = "project"` module | no | no | no |
| (missing) `lgs basecamp build` | — | — | — | — |

So a downstream that wants to build *just one* module's `#lgx` (e.g. for a `logos-standalone-app` smoke test, where `bin-macos-app` isn't the consumer), or wants to build both variants in one shot, has nowhere to go but raw `nix build`.

`lgs basecamp install` is blocked on the distributed stack by the cross-repo variant-flavor mismatch — see [`logos-package-manager#14`](https://github.com/logos-co/logos-package-manager/issues/14) / [`logos-basecamp#197`](https://github.com/logos-co/logos-basecamp/issues/197). Even when that lands, the no-install build path is still useful.

### 1. `lgs basecamp build [--variant lgx|lgx-portable|all] [--module NAME]…` (TR-14)

[Tracker](https://github.com/logos-co/eth-lez-atomic-swaps/blob/master/docs/scaffold-upstream-tracker.md#tr-14) · Pain: [`Makefile#L113-L119`](https://github.com/logos-co/eth-lez-atomic-swaps/blob/master/Makefile#L113-L119) (`swap-module-build`, `swap-ui-build` — both raw `nix build`)

[...]

Issue #173, opened 2026-05-22
[circuits] schema + lgs setup auto-fetch for logos-blockchain-circuits release bundles


danisharora099:
## TL;DR

Every LEZ project needs to fetch a matching `logos-blockchain-circuits` release bundle and point `LOGOS_BLOCKCHAIN_CIRCUITS` at it. There is no scaffold surface for this today — every project re-implements platform detection, version pinning, tarball download, version verification, and env-var export in shell. This project does it in 68 lines of Makefile.

Surfaced while bringing up `eth-lez-atomic-swaps` against scaffold master. Originating tracker entry: [`docs/scaffold-upstream-tracker.md#tr-07`](https://github.com/logos-co/eth-lez-atomic-swaps/blob/master/docs/scaffold-upstream-tracker.md#tr-07).

## Background

The default `lssa` behaviour is to land circuits under `~/.logos-blockchain-circuits/`, which is actively hostile to working on two LEZ projects at different pins simultaneously. The workaround every downstream picks: isolate circuits to a project-local dir and export `LOGOS_BLOCKCHAIN_CIRCUITS` so cargo + scaffold + child processes all see the same dir.

This project's Makefile at [`Makefile#L11-L68`](https://github.com/logos-co/eth-lez-atomic-swaps/blob/master/Makefile#L11-L68) demonstrates the full pattern:

- A `CIRCUITS_VERSION := v0.4.2` pin sourced manually (must be bumped whenever the `lssa` pin in `scaffold.toml` moves to a release requiring newer circuits — nothing automates the link).
- A 12-line platform-detection nested-`ifeq` block (`Darwin/arm64` → `macos-aarch64`, `Linux/x86_64` → `linux-x86_64`, `Linux/aarch64` → `linux-aarch64`).
- A 22-line `circuits:` target that: short-circuits on a matching `VERSION` file, refuses with a useful error on unsupported platforms (macOS Intel is not published upstream), curls the tarball from `github.com/logos-blockchain/logos-blockchain-circuits/releases/download/<ver>/...`, extracts with `--strip-components=1`, verifies the extracted `VERSION` file matches.
- An `export LOGOS_BLOCKCHAIN_CIRCUITS := $(CIRCUITS_DIR)` line so cargo, `logos-scaffold`, and any child processes inherit it.

68 lines of shell for a function every LEZ project needs verbatim.

### `[circuits]` schema in `scaffold.toml`

```toml
[circuits]
version = "v0.4.2"
[...]

Issue #171, opened 2026-05-22
umbrella: [basecamp.profiles.<name>] schema for per-profile launch config (env, runtime dir, log file, paths, custom names)


danisharora099:
## TL;DR

`lgs basecamp launch <profile>` covers the dev-stack happy path, but every downstream basecamp project that ships the distributed stack or runs more than one instance ends up re-implementing the same per-profile launch glue in shell/Makefile. We propose adding a `[basecamp.profiles.<name>]` TOML table that owns the per-profile launch contract, plus a small set of related fields on the surrounding `[repos.basecamp]` and `lgs basecamp launch` CLI.

Surfaced while bringing up `eth-lez-atomic-swaps` against scaffold master. Full per-ask context in [`docs/scaffold-upstream-tracker.md`](https://github.com/logos-co/eth-lez-atomic-swaps/blob/master/docs/scaffold-upstream-tracker.md) entries TR-04 / TR-05 / TR-08 / TR-12 / TR-16 / TR-17.

The umbrella shape mirrors #163 (which bundled `[basecamp.env]` + `[basecamp.env_append]` + `[basecamp.profiles.<name>.env]` in one issue). We'd be happy to split if maintainers prefer N issues — flagging the design coherence first.

## Six concrete asks

Each links a one-paragraph rationale in the originating tracker entry, plus the exact downstream file/line where the workaround lives today.

### 1. `[basecamp.profiles.<name>].env_file` + `.env = {…}` — per-profile env files (TR-05)

[Tracker](https://github.com/logos-co/eth-lez-atomic-swaps/blob/master/docs/scaffold-upstream-tracker.md#tr-05) · Pain: [`scripts/basecamp-instance.sh#L252-L291`](https://github.com/logos-co/eth-lez-atomic-swaps/blob/master/scripts/basecamp-instance.sh#L252-L291)

Two-instance dogfooding needs maker to load `.env` and taker to load `.env.taker`, plus per-profile overrides (`SWAP_UI_AUTO_ENV_FILE`, `SWAP_UI_AUTO_ROLE`). Scaffold has no schema for that mapping today; the launcher hard-codes it.

```toml
[basecamp.profiles.alice]
env_file = ".env"
env = { SWAP_UI_AUTO_ROLE = "maker" }

[basecamp.profiles.bob]
env_file = ".env.taker"
env = { SWAP_UI_AUTO_ROLE = "taker" }
```

[...]

PR #27, opened 2026-03-25
feat: add --json output to wallet list and topup commands

ygd58:
## Summary

Closes #23.

`wallet list` and `wallet topup` only produced human-readable text. This adds `--json` flags to both commands for programmatic consumption, mirroring `localnet status --json` and `doctor --json`.

## Real User Need

Issue #23 (opened by @danisharora099): programmatic consumers must parse stdout strings — no structured output exists for wallet commands.

## End-to-End Verification

Tested on a fresh scaffolded project (after setup):

```bash
$ logos-scaffold wallet list --json
[
  { "account": "Preconfigured Public/2RHZhw9h534Zr3eq2RGhQete2Hh667foECzXPmSkGni2" },
  { "account": "Preconfigured Public/CbgR6tj5kWx5oziiFptM7jMvrQeYY3Mzaao6ciuhSr2r" },
  ...
]
```

stdout is valid JSON, command echoes suppressed in --json mode.

## Local Checks

- `cargo build` ✅
- `cargo test`: 59 passed, 3 pre-existing failures on master ✅
- `cargo fmt --check` ✅
- Tests: `wallet_list_json_outputs_valid_json` ✅

PR #222, opened 2026-06-16
feat(basecamp): add a module build/run workflow

danisharora099:
Implements scaffold **#174** for the Phase 2 atomic-swaps work: move the missing Basecamp module build/run loop out of project-specific scripts and into Scaffold.

## Feature outcome

Developers can now use Scaffold itself to build Basecamp module artifacts and run captured modules during local development.

For atomic-swaps, this replaces the manual Makefile targets and shell logic that currently know how to build portable/non-portable module outputs and run modules from their flakes. Once this lands, that workflow becomes a normal Scaffold capability instead of atomic-swaps-specific glue.

## Before / after

Before this PR, `lgs basecamp` was already the place for Basecamp project setup and launch workflows. It could:
- `setup`: fetch/build Basecamp and LGPM, then seed the standard profiles
- `modules`: capture project modules and dependencies into `scaffold.toml`
- `install`: replay the captured modules into Basecamp profiles
- `launch`: start Basecamp for a profile with clean-slate state
- `develop`: enter a captured module's Nix dev shell
- `build-portable`: build captured project modules as `lgx-portable` artifacts for hand-loading into a portable Basecamp build
- `doctor` / `docs`: inspect Basecamp project state and print compatibility guidance

The missing part was the normal module inner loop. Atomic-swaps still needed its own Makefile/script logic for "build this module output" and "run this captured module from its flake."

With this PR:
- `lgs basecamp build` generalizes module artifact builds beyond the old portable-only helper
- `--variant lgx|lgx-portable|all` selects the output type, and `--module NAME` scopes the build to one module
- `build-portable` remains as a backward-compatible alias for the portable variant
- `lgs basecamp run <module>` launches a captured module from its flake for the local inner loop
- future Scaffold projects get the same Basecamp module workflow without copying atomic-swaps scripts

### `lgs basecamp build [--variant lgx|lgx-portable|all] [--module NAME]`

- Builds project-module `.lgx` artifacts per flake variant.
- Symlinks outputs into `.scaffold/basecamp/<variant-dir>/` in load order.
- Defaults `--variant` to `all`.
- Keeps `build-portable` as a backward-compatible alias for `--variant lgx-portable`.
- Validates and de-duplicates variants before filesystem work.

### `lgs basecamp run <module> [--host standalone|basecamp]`

- Resolves the module flake from `[modules.<name>]` and runs it from the project root.
- Defaults the host from module metadata: `ui_qml` -> `standalone`, otherwise `basecamp`.
- Adds optional `[modules.<name>].standalone_app`.
- Rejects prebuilt `.lgx` path captures with a clearer re-capture/editing hint.

Both verbs are mirrored into the public Rust API via `BasecampCommand::Build` and `BasecampCommand::Run`.

## Verification

- `cargo test` (462 lib + 171 CLI integration + 5 doc tests) passed on `danisharora099/basecamp-verb-granularity`; existing warnings are the pre-existing `sha2` cfg warnings in `setup.rs`.
- DOGFOODING `B5`/`B6` runbook updated: `B5` now covers `basecamp build --variant lgx|lgx-portable|all`, `--module`, and `build-portable`; `B6` covers `basecamp run <module> [--host standalone|basecamp]` and `.lgx` path rejection.
- DOGFOODING `E1`/Basecamp command smoke: verified `basecamp build --help`, `basecamp build-portable --help`, and `basecamp run --help` from the freshly built binary.
- Outside-project smoke: `basecamp run missing` fails with the scaffold project-scoped error and `create/new` remediation.
- Full Nix/Basecamp artefact builds and GUI run-loop dogfooding were not rerun in this docs follow-up.

> Note: branched off `origin/master`. The Phase 2 sibling work (#173 circuits, #171 profiles) lives on separate branches; this PR is scoped to #174 only.

---

## Merge order (scaffold Phase 2)

Part of the scaffold Phase 2 trio for atomic-swaps [#27](https://github.com/logos-co/eth-lez-atomic-swaps/issues/27). Land in order:

**#221 (circuits) -> #222 (verbs) -> #223 (profiles)**

- #221 is independent of the other two.
- #222 and #223 both extend the same `basecamp` enums / subcommands (and #223 also changes the `Launch` variant), so **whichever merges second needs a small rebase**.
- Each PR is isolated by concern and individually mergeable against `master`.

### DIFF
--- a/DOGFOODING.md
+++ b/DOGFOODING.md
@@ -157,7 +157,8 @@ If any of these is missing, do not "skip the real run" — go back and fix the s
 | B2 | external module project | Core | Module capture, install, and single-instance launch | `basecamp modules`, `basecamp modules --show`, `basecamp install`, `basecamp launch alice` |
 | B3 | external module project | Core | Two-instance p2p dogfooding | `basecamp launch alice`, `basecamp launch bob` (parallel) |
 | B4 | external module project | Advanced | Clean-slate scrub semantics on relaunch | `basecamp launch alice` (×2) |
-| B5 | external module project | Advanced | Portable artefact build for AppImage hand-loading | `basecamp build-portable` |
+| B5 | external module project | Advanced | Module artefact builds by variant | `basecamp build`, `basecamp build-portable`, `--variant`, `--module` |
+| B6 | external module project | Advanced | Captured module run loop | `basecamp run <module>`, `--host standalone|basecamp` |
 | A1 | N/A | Advanced | Public Rust API surface for embedding scaffold in tests/tooling | `logos_scaffold::api::Project`, `cargo doc`, doctests |
 | T1 | `default` | Advanced | Isolated test-node lifecycle and caller-project pins | `test-node pins`, `test-node prepare`, `test-node doctor`, `test-node start`, `test-node status`, `test-node stop`, `test-node run` |
 | T2 | `default` | Advanced | Typed RPC reads against a running test node | `test-node tx submit-and-wait`, `test-node blocks head/range/wait`, `test-node clock read/wait-stable`, `test-node account get/batch-get`, `test-node proof get`, `test-node snapshot accounts` |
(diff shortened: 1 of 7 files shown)


weboko (approved):
Reviewed and exercised the non-Nix surface end-to-end. Clean work — approving. 🟢

**Prior review feedback is addressed** (confirmed against HEAD `1e27e73`):
- `--variant` now defaults to `all` per #174 (`[default: all]` in `build --help`).
- `normalize_build_variants` rejects an empty list and validates each value against the `{lgx, lgx-portable}` allowlist (and de-dupes) *before* anything reaches a nix attr or output-dir name — so the public-API `variants: Vec<String>` can't smuggle an unexpected attr. Good, that closes the injection/empty-list concerns.
- `basecamp run` on a `--path`-captured (prebuilt `.lgx`) module now bails with a precise fix-it hint instead of a confusing `nix run path:/…` failure.
- The unimplemented `--host basecamp` stub was dropped; `--host` accepts only `standalone` (with the API-level `other` arm still erroring clearly).

**Security:** the variant allowlist + the module lookup being a plain config-map key (no path built from the module name) keep this tight. `nix run <flake>` executes the user's own captured module flake — inherent and consistent with the existing `develop`/`launch`/`install` verbs, not a new surface.

**Error UX (verified end-to-end, no Nix needed):** `basecamp build` with nothing captured and `basecamp run <unknown>` both emit the specific 'run `basecamp modules` first' / known-module-list guidance *before* the `ensure_nix_present` preflight, so a Nix-less user with an empty project gets the actionable message first. Nice ordering.

**Build/tests:** `cargo fmt --check` clean, no new warnings, 133 basecamp + 28 CLI basecamp tests pass. DOGFOODING.md is updated to cover `build`/`run`.

**Dogfooding limitation (disclosure):** no Nix in this environment, so the real `nix`-backed `basecamp build` / `run` couldn't be executed; I reviewed those code paths and drove everything up to the Nix boundary.

**Coordination note:** this and #223 both substantially rewrite `src/commands/basecamp.rs` and `src/config.rs`; they'll conflict on merge. Whichever lands second will need a careful reconcile (both touch the profile model and the same serializer).

PR #228, opened 2026-06-29
Fix default-template build/deploy against pinned LEZ v0.1.2 + spel v0.5.0

weboko:
## Summary
Dogfooding the default template against the current pinned deps (LEZ v0.1.2,
spel v0.5.0) surfaced two breakages in the generated-project happy path. Both
are fixed and verified clean-room (fresh `new` → `setup` → `build` → `deploy`).

## Changes
- **Fix PDA runner compile error** — `run_hello_world_with_authorization_through_tail_call_with_pda`
  called `AccountId::from((&program.id(), &PDA_SEED))`, which the pinned LEZ
  `AccountId` doesn't implement, so `logos-scaffold build` failed (E0277).
  Switched to the canonical `AccountId::for_public_pda(...)`.
- **Fix deploy program-id resolution** — spel v0.5.0 renamed its ELF command
  (`spel inspect` now decodes account data; `spel program-id <FILE>` extracts
  the image ID). `extract_program_id` still called `inspect`, so every deploy
  reported `program_id: unavailable` and `$SCAFFOLD_PROGRAM_ID` was unset for
  post-deploy hooks. Switched to `program-id`.
- **Docs** — added the missing risc0 Rust toolchain provisioning step to
  DOGFOODING.md (needed to compile guests for build/deploy).

## Testing
E1–E3, A1, D1–D7, and T1–T4 pass end-to-end against a real sequencer,
including a real committed wallet transaction and exact-state seeding.

### DIFF
--- a/.github/workflows/template-e2e.yml
+++ b/.github/workflows/template-e2e.yml
@@ -46,17 +46,21 @@ jobs:
         uses: Swatinem/rust-cache@v2
 
       # The bootstrap cache holds the cloned lez/spel repos and their target/
-      # dirs (~/.cache/logos-scaffold) AND the extracted circuits release
-      # (~/.logos-blockchain-circuits, where the fetch step below puts it).
-      # Both must be cached, or the multi-MB circuits tarball re-downloads on
-      # every run. Keying on constants.rs means a pin or circuits-version bump
+      # dirs (~/.cache/logos-scaffold), the extracted circuits release
+      # (~/.logos-blockchain-circuits, where the fetch step below puts it),
+      # and the risc0 toolchains (~/.risc0, installed further below). All
+      # must be cached, or the multi-MB circuits tarball and the ~500 MB
+      # risc0 Rust toolchain re-download on every run. rzup is idempotent,
+      # so a cache hit on ~/.risc0 turns the install step into a fast no-op.
+      # Keying on constants.rs means a pin or circuits-version bump
       # invalidates it (rebuild from scratch) while same-pin re-runs stay warm.
-      - name: Cache scaffold bootstrap (repos + circuits)
+      - name: Cache scaffold bootstrap (repos + circuits + risc0)
         uses: actions/cache@v4
         with:
           path: |
             ~/.cache/logos-scaffold
             ~/.logos-blockchain-circuits
+            ~/.risc0
           key: scaffold-bootstrap-${{ runner.os }}-${{ hashFiles('src/constants.rs') }}
 
       - name: Build scaffold CLI
--- a/DOGFOODING.md
+++ b/DOGFOODING.md
@@ -110,6 +110,21 @@ mkdir -p "$EXT" && cp /tmp/cr/r0vm "$EXT/r0vm" && chmod +x "$EXT/r0vm"
 
 `find_r0vm_path_for_lez` looks at `~/.risc0/extensions/v<risc0-zkvm-version>-cargo-risczero-<arch>-<os>/r0vm`; placing r0vm there is what wires it into a spawned sequencer (scaffold sets `RISC0_SERVER_PATH` from it). The sequencer runs with `RISC0_DEV_MODE=1` (the test-node default), so r0vm executes guests without real proving — which is why no GPU/prover is needed.
 
+**2b. Install the risc0 Rust toolchain (needed for `build`/`deploy`, not for the T-series).** `r0vm` alone runs guests; it does *not* compile them. `logos-scaffold build` (and therefore `deploy`/`run`/`D`/`L` scenarios) shells out through `risc0-build`, which asks the `rzup` library for a default-installed Rust toolchain and panics with `Risc Zero Rust toolchain not found. Try running 'rzup install rust'` if none exists. `rzup install rust` fails the same way behind a TLS-intercepting proxy, so install it by hand from the `risc0/rust` release and register it where the `rzup` library looks (`~/.risc0/toolchains/` + `~/.risc0/settings.toml`):
+
+```bash
+RVER=1.94.1   # match the toolchain the risc0 release line ships; `rzup install rust` would pick the latest r0.* tag
+curl -sSL -H "Authorization: Bearer $GH_TOKEN" -o /tmp/rust-toolchain.tgz \
+  "https://github.com/risc0/rust/releases/download/r0.$RVER/rust-toolchain-$TRIPLE.tar.gz"
+VDIR="$HOME/.risc0/toolchains/v$RVER-rust-$TRIPLE"
+mkdir -p "$VDIR" && tar xzf /tmp/rust-toolchain.tgz -C "$VDIR"
+printf '[default_versions]\nrust = "%s"\n' "$RVER" > "$HOME/.risc0/settings.toml"
+ln -sfn "$VDIR" "$HOME/.rustup/toolchains/risc0"
+ls "$VDIR/lib/rustlib" | grep riscv32im   # → riscv32im-risc0-zkvm-elf (guest target present)
+```
+
+Circuits need no equivalent hand-provisioning for generated projects: `build`/`setup` resolve the `[circuits]` table in `scaffold.toml`, materialise the configured release into `[circuits].install_dir` (default `.scaffold/circuits`), and export `LOGOS_BLOCKCHAIN_CIRCUITS` for the rest of the process. A pre-set `LOGOS_BLOCKCHAIN_CIRCUITS` pointing at a populated release short-circuits the download; only older projects without `[circuits]` still require it.
+
 **3. Build the real sequencer.** `test-node prepare` downloads the circuits release (via `curl`, automatically) and builds `sequencer_service`. It is long (~6 min); run it, then confirm doctor is green:
 
 ```bash
@@ -117,9 +132,10 @@ mkdir -p "$EXT" && cp /tmp/cr/r0vm "$EXT/r0vm" && chmod +x "$EXT/r0vm"
 "$SCAFFOLD_BIN" test-node doctor  --project "$P" --json | jq .ok   # → true (all checks pass)
 ```
 
-**4. (Only for real transactions — T4) build the wallet via `setup`.** Two gotchas distinguish `setup` from `test-node prepare`: its circuits precheck does **not** consult the scaffold cache (it only accepts `LOGOS_BLOCKCHAIN_CIRCUITS` or `$HOME/.logos-blockchain-circuits/`), so ensure one of those is present before running it; and it uses cwd discovery, so it must run **inside** the project (no `--project` flag):
+**4. (Only for real transactions — T4) build the wallet via `setup`.** One gotcha distinguishes `setup` from `test-node prepare`: it uses cwd discovery, so it must run **inside** the project (no `--project` flag). Projects with a `[circuits]` table materialise their configured release during `setup` itself; only for older projects without `[circuits]` must `LOGOS_BLOCKCHAIN_CIRCUITS` be exported first:
 
 ```bash
+# Only needed for older projects without [circuits] in scaffold.toml:
 export LOGOS_BLOCKCHAIN_CIRCUITS="$("$SCAFFOLD_BIN" test-node pins --project "$P" --json | jq -r .circuits_path)"
 ( cd "$P" && "$SCAFFOLD_BIN" setup )   # builds wallet + spel, seeds the default wallet (~3 min) → "setup complete"
 ```
--- a/DOGFOODING.md
+++ b/DOGFOODING.md
@@ -114,6 +114,7 @@ mkdir -p "$EXT" && cp /tmp/cr/r0vm "$EXT/r0vm" && chmod +x "$EXT/r0vm"
 
 ```bash
 RVER=1.94.1   # match the toolchain the risc0 release line ships; `rzup install rust` would pick the latest r0.* tag
+TRIPLE=${TRIPLE:-x86_64-unknown-linux-gnu}   # aarch64-apple-darwin on macOS; matches step 2
 curl -sSL -H "Authorization: Bearer $GH_TOKEN" -o /tmp/rust-toolchain.tgz \
   "https://github.com/risc0/rust/releases/download/r0.$RVER/rust-toolchain-$TRIPLE.tar.gz"
 VDIR="$HOME/.risc0/toolchains/v$RVER-rust-$TRIPLE"
--- a/src/commands/deploy.rs
+++ b/src/commands/deploy.rs
@@ -456,8 +456,10 @@ fn deploy_single_program(
 
 /// Wall-clock cap for `spel program-id`. The CLI typically returns in
 /// milliseconds; a hung binary should not block the deploy summary.
-/// Override with `LOGOS_SCAFFOLD_SPEL_INSPECT_TIMEOUT_MS` if needed.
-const SPEL_INSPECT_TIMEOUT: std::time::Duration = std::time::Duration::from_secs(15);
+/// Override with `LOGOS_SCAFFOLD_SPEL_PROGRAM_ID_TIMEOUT_MS` if needed;
+/// `LOGOS_SCAFFOLD_SPEL_INSPECT_TIMEOUT_MS` is honored as a legacy alias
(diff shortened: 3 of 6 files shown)


weboko (commented):
## Automated dogfooding review — verified end-to-end ✅

I re-ran the default-template happy path against the pinned deps this PR targets (**LEZ v0.1.2 / spel v0.5.0**) on a fresh box (Nix + risc0 toolchain provisioned by hand per `DOGFOODING.md` §2b, real running `sequencer_service`). Both headline fixes hold up:

**1. PDA compile fix (`AccountId::for_public_pda`)** — `new` → `build` compiles clean, no `E0277`; all guest ELFs produced, including `tail_call_with_pda.bin`. The host runner `run_hello_world_with_authorization_through_tail_call_with_pda` builds.

**2. Deploy program-id fix (`spel inspect` → `spel program-id`)** — confirmed against the real spel v0.5.0 binary:
- `spel program-id <ELF>` emits `   ImageID (hex bytes): d56f49c0…7981` — exactly the `SPEL_IMAGE_ID_PREFIX` the parser keys on.
- `spel inspect <ELF>` now errors (`Account inspection requires --idl`), i.e. the old code path would indeed have reported `program_id: unavailable`. Fix reproduces and resolves the reported bug.
- Full `deploy` (real sequencer): **5/5 submitted, every program prints a real `program_id: <hex>`**. `deploy --program-path … --json` carries `"program_id"`; `deploy nonexistent_program` errors with the available-program list. D1/D3 green.

Note (not caused by this PR): during block production the LEZ v0.1.2 `sequencer_service` panics upstream in `logos_blockchain_core::mantle::encoding::encode_channel_inscribe` (`produce_new_block_with_mempool_transactions`). Submissions succeed; the panic is in LEZ, unrelated to scaffold's deploy path.

### Non-blocking findings

**A. [nit] Stale `inspect` naming in `deploy.rs`.** After moving to `program-id`, `SPEL_INSPECT_TIMEOUT` (deploy.rs:460) and the override env var `LOGOS_SCAFFOLD_SPEL_INSPECT_TIMEOUT_MS` (deploy.rs:472) still say "inspect", and the doc comment now says "spel program-id" while pointing at the INSPECT-named var. Cosmetic, but consider `SPEL_PROGRAM_ID_TIMEOUT` / `…_PROGRAM_ID_TIMEOUT_MS` (or a one-line note that the name is retained for back-compat).

**B. [docs] `DOGFOODING.md` L-series not updated for the now-known-broken lez-framework guest.** This PR correctly makes the lez-framework CI leg `continue-on-error` (upstream `ring` can't cross-compile to the zkVM guest target) with a great explanatory comment. But **L1** (line 632) still states *"`build` succeeds for the LEZ project workspace"* with no caveat. Per the runbook's own maintenance note, L1/L4 should record that the lez-framework guest build is currently upstream-blocked and expected to fail, so the runbook matches the CI change.

**C. [transparency] PR description under-states scope.** The body lists PDA + deploy + docs, but the PR also (a) migrates the lez-framework template `lez-framework*` → `spel-framework*` with an API change (`LezOutput::states_only` → `SpelOutput::execute`) and (b) changes CI. Worth mentioning so reviewers/mergers see the real surface.

**D. [coverage] lez-framework guest semantics are unverifiable here.** Because the guest can't compile (upstream block), the migrated `increment` (`SpelOutput::execute(vec![counter, authority], vec![])` relying on framework-derived auto-claims) couldn't be compile- or run-verified. Since it ships in a user-facing template, please (i) confirm the auto-claim behavior matches the previous explicit `new_claimed`/`new`, and (ii) file a tracking issue for the `ring`/`host`-feature upstream block so the `continue-on-error` can be removed later.

Nice touches: the `Cargo.toml.template` tag-vs-rev source-dedup comment and the CI `continue-on-error` rationale are exactly the kind of context that makes this reviewable, and the `tests/cli.rs` stub teaches `program-id` while keeping `inspect` for the passthrough tests.

_Core fixes verified and working; the above are follow-ups, not blockers._

src/commands/deploy.rs
```diff
@@ -454,12 +454,12 @@ fn deploy_single_program(
     })
 }
 
-/// Wall-clock cap for `spel inspect`. The CLI typically returns in
+/// Wall-clock cap for `spel program-id`. The CLI typically returns in
```
weboko, line 457: **[nit] Stale `inspect` naming.** The command is now `program-id`, but the timeout const is still `SPEL_INSPECT_TIMEOUT` and its override var `LOGOS_SCAFFOLD_SPEL_INSPECT_TIMEOUT_MS` (both a few lines below). Consider `SPEL_PROGRAM_ID_TIMEOUT` / `…_PROGRAM_ID_TIMEOUT_MS`, or a one-line note that the var name is kept for back-compat. Purely cosmetic.

.github/workflows/template-e2e.yml
```diff
@@ -129,7 +148,18 @@ jobs:
...
+      # LEZ Framework SDK) declare their `nssa_core` dependency with the
+      # host-only `host` feature unconditionally, which pulls `ring`; `ring`'s
+      # C sources don't cross-compile to the risc0 zkVM guest target (cc-rs
+      # runs the riscv32 gcc with `-m64`). A consumer can't drop a forced
+      # transitive feature, so this can't be fixed in-repo until the framework
+      # gates `host` behind an opt-in feature. The failure stays visible in the
+      # step log. The `default` leg remains strict and gating. Remove the
+      # continue-on-error once spel-framework ships a guest-safe release.
       - name: Build rendered project workspace
+        continue-on-error: ${{ matrix.template == 'lez-framework' }}
```
weboko: Scoping `continue-on-error` to only the `lez-framework` leg (default stays strict), with the upstream-`ring` rationale above, is the right call. Two follow-ups: (1) `DOGFOODING.md` L1/L4 still say the lez-framework `build` succeeds — please add the same upstream-blocked caveat so the runbook matches CI; (2) worth a tracking issue for the upstream `host`-feature/`ring` block so this `continue-on-error` can be removed.

3esmit:
Probably is better to wait for https://github.com/logos-co/spel/pull/238 to merge, and with this PR merged, would be nice to have a release tag.

3esmit:
Spel released to v0.6.0 https://github.com/logos-co/spel/releases/tag/v0.6.0

DOGFOODING.md
```diff
@@ -110,16 +110,33 @@ mkdir -p "$EXT" && cp /tmp/cr/r0vm "$EXT/r0vm" && chmod +x "$EXT/r0vm"
 
 `find_r0vm_path_for_lez` looks at `~/.risc0/extensions/v<risc0-zkvm-version>-cargo-risczero-<arch>-<os>/r0vm`; placing r0vm there is what wires it into a spawned sequencer (scaffold sets `RISC0_SERVER_PATH` from it). The sequencer runs with `RISC0_DEV_MODE=1` (the test-node default), so r0vm executes guests without real proving — which is why no GPU/prover is needed.
 
+**2b. Install the risc0 Rust toolchain (needed for `build`/`deploy`, not for the T-series).** `r0vm` alone runs guests; it does *not* compile them. `logos-scaffold build` (and therefore `deploy`/`run`/`D`/`L` scenarios) shells out through `risc0-build`, which asks the `rzup` library for a default-installed Rust toolchain and panics with `Risc Zero Rust toolchain not found. Try running 'rzup install rust'` if none exists. `rzup install rust` fails the same way behind a TLS-intercepting proxy, so install it by hand from the `risc0/rust` release and register it where the `rzup` library looks (`~/.risc0/toolchains/` + `~/.risc0/settings.toml`):
+
+```bash
+RVER=1.94.1   # match the toolchain the risc0 release line ships; `rzup install rust` would pick the latest r0.* tag
```
weboko, line 116: Minor: the trailing comment says `rzup install rust` "would pick the latest r0.* tag," but the pinned `RVER=1.94.1` is no longer the latest — `risc0/rust` shipped `r0.1.97.0` on 2026-07-16. Pinning is the right call for a deterministic runbook, but a reader who trusts the "latest" hint would end up on 1.97.0 instead. Suggest tweaking the wording to say the pin is intentional and may lag the latest tag. (Verified 1.94.1 builds the guests cleanly, so the pin itself is good.)

weboko (commented):
**Automated independent review + end-to-end verification** (dogfooded on a real LEZ sequencer, `RISC0_DEV_MODE=1`).

### Verdict
The core fix is correct, well-scoped, backward-compatible, and I verified it end-to-end. LGTM aside from two minor, non-blocking nits below.

### End-to-end verification
Provisioned the real toolchain (r0vm 3.0.5 + risc0 Rust 1.94.1 + circuits v0.4.1), then ran the default-template happy path against the pinned deps (LEZ `cf3639d8`, spel v0.5.0):

- `build` → guest cross-compiles to `riscv32im-risc0-zkvm-elf` ✅
- `setup` → builds `sequencer_service` + `wallet` + `spel`, seeds default wallet ✅
- `localnet start` → ready ✅
- **`deploy` (this branch) → 5/5 programs submitted, each with a resolved 64-hex `program_id`** (e.g. `hello_world` → `34903a89…fb20`) ✅

Negative control with the pre-fix binary reproduces the bug: `program_id: unavailable` ×5. Direct check confirms the root cause and the fix:
- `spel inspect <elf>` → `Account inspection requires --idl` (v0.5.0 repurposed `inspect` to account-data decoding)
- `spel program-id <elf>` → `ImageID (hex bytes): 34903a89…fb20` — exactly what `SPEL_IMAGE_ID_PREFIX` parses.

### Code review
- **Correctness** — `inspect` → `program-id` rename is complete and consistent (deploy.rs call site + doctor.rs/run.rs comments); the only remaining `inspect` mention is the intentional legacy-alias doc comment.
- **Backward-compat** — nice touch keeping `LOGOS_SCAFFOLD_SPEL_INSPECT_TIMEOUT_MS` as an alias for the new `LOGOS_SCAFFOLD_SPEL_PROGRAM_ID_TIMEOUT_MS` via `.or_else(...)`.
- **Security/privacy** — no concern: `Command` with separate args (no shell), file path passed positionally, hung-binary timeout preserved, no secrets touched.
- **Tests** — the stub now accepts both `program-id` and `inspect` so the `lgs spel -- inspect` passthrough-proxy tests still exercise arg forwarding; all 4 `program_id` + 10 `spel` cli tests pass on this branch.
- **CI** — caching `~/.risc0` matches the existing idempotent `rzup install rust/cpp` step; coherent.
- **Docs** — new DOGFOODING step 2b (risc0 Rust toolchain hand-provisioning) is genuinely necessary; I reproduced it and it works.

### Minor, non-blocking
1. DOGFOODING.md `RVER=1.94.1` comment — inline note below.
2. PR description lists a "PDA runner compile error" fix (`AccountId::for_public_pda`) as part of this PR, but that change isn't in the diff — it's already on `master`. The description over-describes; harmless, just flagging so the changelog is accurate.

DOGFOODING.md
```diff
@@ -110,16 +110,33 @@ mkdir -p "$EXT" && cp /tmp/cr/r0vm "$EXT/r0vm" && chmod +x "$EXT/r0vm"
 
 `find_r0vm_path_for_lez` looks at `~/.risc0/extensions/v<risc0-zkvm-version>-cargo-risczero-<arch>-<os>/r0vm`; placing r0vm there is what wires it into a spawned sequencer (scaffold sets `RISC0_SERVER_PATH` from it). The sequencer runs with `RISC0_DEV_MODE=1` (the test-node default), so r0vm executes guests without real proving — which is why no GPU/prover is needed.
 
+**2b. Install the risc0 Rust toolchain (needed for `build`/`deploy`, not for the T-series).** `r0vm` alone runs guests; it does *not* compile them. `logos-scaffold build` (and therefore `deploy`/`run`/`D`/`L` scenarios) shells out through `risc0-build`, which asks the `rzup` library for a default-installed Rust toolchain and panics with `Risc Zero Rust toolchain not found. Try running 'rzup install rust'` if none exists. `rzup install rust` fails the same way behind a TLS-intercepting proxy, so install it by hand from the `risc0/rust` release and register it where the `rzup` library looks (`~/.risc0/toolchains/` + `~/.risc0/settings.toml`):
```
danisharora099, line 113: Could we either add the C++ toolchain fallback here or narrow this claim to the default template? This says the step covers the L-series too, but the workflow itself notes that the `lez-framework` guest needs Risc Zero's C++ toolchain and runs `rzup install cpp`. Someone following this manual path for L1/L4 can finish every command here and still hit the missing-C++ failure.

DOGFOODING.md
```diff
@@ -110,16 +110,33 @@ mkdir -p "$EXT" && cp /tmp/cr/r0vm "$EXT/r0vm" && chmod +x "$EXT/r0vm"
...
+
+```bash
+RVER=1.94.1   # match the toolchain the risc0 release line ships; `rzup install rust` would pick the latest r0.* tag
+TRIPLE=${TRIPLE:-x86_64-unknown-linux-gnu}   # aarch64-apple-darwin on macOS; matches step 2
+curl -sSL -H "Authorization: Bearer $GH_TOKEN" -o /tmp/rust-toolchain.tgz \
```
danisharora099, line 118: This works when `GH_TOKEN` is present, but fails surprisingly when it is not: the empty `Authorization: Bearer ` header makes this public asset return 401, while the same URL without that header returns 200. Could we build the auth header conditionally and use `curl -fL --retry 3` so the copy-paste path works both with and without a token and fails clearly on HTTP errors?

tests/cli.rs
```diff
@@ -3615,13 +3615,16 @@ if [ "${SPEL_FAIL:-0}" = "1" ]; then
   exit 7
 fi
 
-if [ "$#" -ge 2 ] && [ "$1" = "inspect" ]; then
+# `program-id <ELF>` is what `extract_program_id` calls (spel v0.5.0 renamed
+# the old `inspect <ELF>`); `inspect` is still accepted so the `lgs spel --
+# inspect` passthrough-proxy tests exercise arg forwarding.
+if [ "$#" -ge 2 ] && { [ "$1" = "program-id" ] || [ "$1" = "inspect" ]; }; then
```
danisharora099, line 3621: Could we make the stub prove which subcommand deploy used? Right now it produces the same image-ID output for both `program-id` and `inspect`, so the deploy tests would stay green if `extract_program_id` accidentally regressed to the old, broken `inspect` call. Recording and asserting `$1`, or making `inspect` valid only in the passthrough tests, would make this a real regression test.

src/commands/deploy.rs
```diff
@@ -469,14 +471,15 @@ pub(crate) fn extract_program_id(spel_bin: &Path, binary_path: &Path) -> Option<
...
+        .or_else(|_| std::env::var("LOGOS_SCAFFOLD_SPEL_INSPECT_TIMEOUT_MS"))
         .ok()
         .and_then(|s| s.parse::<u64>().ok())
         .map(std::time::Duration::from_millis)
-        .unwrap_or(SPEL_INSPECT_TIMEOUT);
+        .unwrap_or(SPEL_PROGRAM_ID_TIMEOUT);
 
     let mut child = Command::new(spel_bin)
-        .arg("inspect")
+        .arg("program-id")
```
danisharora099, line 482: This code change is right, but a few user-facing docs still tell people that deploy uses `spel inspect`: `README.md:256`, `FURPS.md:56/94`, and `ADR.md:158/366`. Could we update those in this PR so users and future contributors do not follow the command that spel v0.5.0 repurposed?

danisharora099 (changes requested):
The `spel program-id` change looks correct, and local verification passed (`cargo fmt --check`; 692 tests).

I found a few places where the review surface has not caught up with the fix yet:

- the new manual Risc Zero setup is not quite copy-paste safe;
- the test stub would still let the old `spel inspect` regression pass; and
- the README, FURPS, and ADR still direct people to `spel inspect`.

One more housekeeping item: the PR description is now stale after rebasing. The PDA change is already on `master` via #224 and is not in this diff, while the `~/.risc0` cache change is in the diff but is not mentioned. Please tighten the description to what this PR currently ships.

The three normal `validate` checks are green. The heavier Template E2E workflow did not run on the current `76fd3a9`, so I would like the targeted fixes below before approval.

weboko:
will re-create PR after dogfooding with never Spel and LEZ

PR #229, opened 2026-06-29
Add fast block-time support to test-node API

### LINKED ISSUE #227: Add fast block-time support to test-node config API
## Problem

Downstream integration suites that use scaffold `test-node` need an opt-in way to run sequencer blocks faster than the pinned LEZ debug defaults.

In `logos-blockchain/lez-programs`, the scaffold-backed local sequencer parity suite now starts isolated test nodes through the Rust API. The suite passes, but full runtime is high because each transition waits on normal block production cadence:

- AMM: `17` tests, `1023.45s`
- ATA: `10` tests, `464.06s`
- stablecoin: `2` tests, `110.47s`
- token: `16` tests, `513.28s`

Total runtime is roughly 35 minutes on a warm cache. The old local helper I'm replacing with scaffold patched the sequencer config with fast block timing:

```json
"block_create_timeout": "100ms",
"retry_pending_blocks_timeout": "100ms"
```

The scaffold test-node path currently handles lifecycle, state seeding, structured outcomes, block context, accounts, and proofs, but it does not expose these timing knobs through `TestNodeConfig` or the CLI.

## Proposal

Add opt-in fast block-time configuration to scaffold test nodes.

Suggested Rust API shape:

```rust
pub struct TestNodeConfig {
    // existing fields...
    pub block_create_timeout: Option<Duration>,
    pub retry_pending_blocks_timeout: Option<Duration>,
}
```

Suggested CLI shape:

```bash
lgs test-node start --block-create-timeout 100ms --retry-pending-blocks-timeout 100ms
lgs test-node run --block-create-timeout 100ms --retry-pending-blocks-timeout 100ms -- <command>
```

Implementation should patch the generated `sequencer_config.json` when values are provided, while preserving current default behavior when they are omitted.

## Downstream use

`logos-blockchain/lez-programs` would set both values to `100ms` in its scaffold-backed local sequencer tests so the full parity suite remains practical as a validation gate.

## Acceptance criteria

- `TestNodeConfig` exposes optional block timing overrides.
- `test-node start` and `test-node run` expose equivalent CLI flags.
- Omitted values preserve the current pinned LEZ debug config behavior.
- Provided values are written into the node's generated `sequencer_config.json`.
- Existing test-node behavior and metadata remain compatible.
- Unit tests cover config patching for default and overridden timing values.

3esmit:

Closes #227.

Downstream integration suites use `test-node` through the Rust API and need opt-in faster block production to keep local sequencer parity tests practical.

## User Story

Local sequencer parity suites start isolated test nodes through `test-node`, but full runtime is high when every transition waits on the pinned LEZ debug block cadence. Previous local test helpers patched `block_create_timeout` and `retry_pending_blocks_timeout` to lower millisecond values; the CLI and Rust API did not expose equivalent controls.

## Change Summary

This adds optional millisecond block timing overrides through the Rust test-node helper API and wires equivalent CLI flags through `test-node start` and `test-node run`. Accepted override values are 1 to 3,600,000 ms and invalid values surface as typed configuration errors. When provided, the runtime config writer sets `block_create_timeout` and `retry_pending_blocks_timeout` as millisecond strings in `sequencer_config.json`; when omitted, the pinned debug config values are preserved. Runtime config writes are validated before replacement and staged through unique same-directory temp files so failed or concurrent overrides do not clobber an existing runtime config.

## Verification

- Relevant DOGFOODING scenarios: `T1`, `A1`
- `cargo fmt --check`
- `git diff --check`
- `cargo check`
- `cargo test public_api_ --lib`
- `cargo test write_node_config --lib`
- `cargo test sequencer_config --lib`
- `cargo test --test cli test_node`
- `cargo test --test testnode_api`
- `cargo run --bin logos-scaffold -- test-node start --help | rg -n "block-create-timeout-ms|retry-pending-blocks-timeout-ms"`
- `cargo run --bin logos-scaffold -- test-node run --help | rg -n "block-create-timeout-ms|retry-pending-blocks-timeout-ms"`
- `cargo test --all-targets`

## Checklist

- [x] CI is green
- [x] Relevant DOGFOODING scenarios rerun (`T1`, `A1`)
- [x] Docs updated (`README.md`, `DOGFOODING.md`) if user-facing behavior changed


DOGFOODING.md
```diff
@@ -1324,7 +1324,7 @@ From the generated project root:
 # capture the node id (state_dir basename) and rpc_url from the JSON
 "$SCAFFOLD_BIN" test-node status --node <node-id> --json
 "$SCAFFOLD_BIN" test-node stop --node <node-id>
-"$SCAFFOLD_BIN" test-node run --serial -- sh -c 'echo "rpc=$LGS_TEST_NODE_RPC_URL port=$LGS_TEST_NODE_PORT"'
+"$SCAFFOLD_BIN" test-node run --serial --block-create-timeout-ms 100 --retry-pending-blocks-timeout-ms 100 -- sh -c 'echo "rpc=$LGS_TEST_NODE_RPC_URL port=$LGS_TEST_NODE_PORT"'
```
danisharora099: **The advertised `100ms` timing silently breaks the stable-read APIs.**

`ClockReadMode::Stable` requires ≥2 *consecutive identical* head+clock samples (`src/testnode/blocks.rs:238`) and sleeps 200ms between them (`:227`, `:269`). A node producing a block every 100ms advances the head at least twice per sample window, so `clock wait-stable` — and the account-boundary reads that reuse the "head identical before/after" logic — can never converge and return a retryable timeout.

This example recommends exactly `--block-create-timeout-ms 100 --retry-pending-blocks-timeout-ms 100`, so the documented value breaks the parity APIs this test-node surface exists to serve, and it surfaces as a vague timeout rather than a clear error.

Suggested (right-sized) fix: move the doc examples off `100ms` to a value comfortably above the 200ms sample interval (e.g. `500ms`) and add a one-line caveat that timings near/below the `wait-stable` sample cadence break stable reads and account-boundary reads. The complete fix (deriving the sample cadence from block timing, or raising `MIN_BLOCK_TIMING_TIMEOUT_MS`) touches `wait-stable` behavior for all callers and belongs in its own change — worth a TODO.

src/cli.rs
```diff
@@ -823,6 +823,24 @@ struct TestNodeStartArgs {
...
     timeout_sec: u64,
+    /// Override sequencer block_create_timeout in milliseconds.
+    #[arg(
+        long,
+        value_name = "MS",
+        value_parser = clap::value_parser!(u64).range(
+            crate::testnode::MIN_BLOCK_TIMING_TIMEOUT_MS..=crate::testnode::MAX_BLOCK_TIMING_TIMEOUT_MS
+        )
+    )]
+    block_create_timeout_ms: Option<u64>,
```
danisharora099, line 842: **DRY: this flag block is duplicated 4×.** The identical `#[arg(long, value_name = "MS", value_parser = clap::value_parser!(u64).range(...))]` definition appears for both flags in `TestNodeStartArgs` (this line and `:843`) and again in `TestNodeRunArgs` (`:898`, `:907`), and the range is repeated in the attribute even though `MIN/MAX_BLOCK_TIMING_TIMEOUT_MS` exist.

Suggest extracting a shared `#[derive(Args)] struct BlockTimingArgs { ... }` and `#[command(flatten)]`-ing it into both. The two `BlockTimingOverrides { .. }` constructions at `:1262` and `:1284` then read from `args.block_timing.*`. Removes ~20 lines and the risk that a future range change updates one copy but not the others (which would make `start` and `run` validate differently).

src/sequencer_config.rs
```diff
@@ -0,0 +1,145 @@
...
+    fn write_source_config(root: &Path) -> PathBuf {
+        let lez = root.join("lez");
+        let config_path = lez.join(SEQUENCER_CONFIG_REL_PATH);
+        fs::create_dir_all(config_path.parent().expect("parent")).expect("create config dir");
+        fs::write(&config_path, r#"{"home": ".", "genesis_id": 1}"#).expect("write source config");
+        lez
+    }
+
+    #[test]
+    fn patch_runtime_sequencer_config_allows_same_process_parallel_writes() {
```
danisharora099: **Test name over-claims.** `patch_runtime_sequencer_config_allows_same_process_parallel_writes` proves the *final JSON is valid* under concurrent writes — it does not prove concurrency *safety*: callers aren't locked across patch → spawn, so two starts against the same runtime dir could overwrite each other's timing config before launch. In practice `allocate_work_dir` mints a unique dir per node, so normal usage never races — but the name implies a guarantee the test doesn't give.

Suggest renaming to something like `..._produces_valid_json_under_concurrent_writes` and adding a one-line note that concurrent starts must target distinct `--work-dir`s.

danisharora099 (commented):
Solid, single-purpose change with strong unit coverage and a clean DRY refactor of the two config patchers. Two runtime issues to resolve before merge, plus two low-cost cleanups.

**Blockers**

1. **The advertised `100ms` timing breaks the stable-read APIs.** See the inline comment on `DOGFOODING.md`. `ClockReadMode::Stable` needs ≥2 consecutive identical head+clock samples (`src/testnode/blocks.rs:238`) and sleeps 200ms between them (`:227`, `:269`). A node making a block every 100ms advances the head twice per sample window, so `clock wait-stable` and the account-boundary reads that reuse the same logic can never converge and time out.

2. **End-to-end behavior is unverified.** Every new CLI/API integration test drives a Python stub (`tests/common/test_node.rs`) that only loads the config JSON — nothing runs a real sequencer. The `"100ms"` string format is fine (LEZ parses it via `humantime_serde`), but no test proves a fast node actually produces blocks and that the tx/clock/snapshot APIs still work at that cadence. The checklist has `CI is green` and `DOGFOODING scenarios rerun (T1, A1)` unchecked — please run T1/A1 against a real sequencer before merge. It's the one path the stub can't cover, and it would also catch the `wait-stable` interaction above.

**Cleanups (non-blocking):** clap flag duplication in `cli.rs`, and an over-claiming test name in `sequencer_config.rs` — both inline.

3esmit:
Thanks for the review. I addressed all your comments and additional copilot automated teviews

weboko (approved):
Reviewed and validated end-to-end against a real sequencer. This is a well-designed, thoroughly-tested PR — approving. 🟢

**Real end-to-end (T1, default pin, real `sequencer_service`)**
Provisioned the real toolchain (r0vm 3.0.5, circuits v0.4.1, built the standalone sequencer), `test-node doctor` fully green, then:
- Started `test-node start --block-create-timeout-ms 500 --retry-pending-blocks-timeout-ms 500`; runtime `sequencer_config.json` correctly carried `"block_create_timeout": "500ms"` / `"retry_pending_blocks_timeout": "500ms"`.
- **The real sequencer honored it**: ~9 blocks produced in ~5s (~1.78 blk/s) vs the pinned 2s default (~0.5 blk/s) — a real ~4× speedup. This also confirms the `"{ms}ms"` encoding is accepted by LEZ's `humantime_serde` duration fields (`sequencer/core/src/config.rs`).

**On @danisharora099's stable-read concern — validated both ways.** I reproduced it: at 100ms, `clock wait-stable` does not converge (hangs to timeout), because `ClockReadMode::Stable` samples 200ms apart and the head advances every 100ms. At the **500ms** the DOGFOODING example now uses, `clock wait-stable` and `account get` both converge cleanly. So the resolution (500ms example + the caveat sentence in README/DOGFOODING) is correct and the caveat's wording is accurate.

**Code review**
- Clean API: `BlockTimingOverrides` builder, `#[non_exhaustive]`, fail-fast `validate()` before spawn, typed `BlockTimingValidationError` wired through `api::error::classify`, and back-compat `start`/`run` shims delegating to the `_with_block_timing` variants.
- `patch_runtime_sequencer_config` centralizes the read/patch/write and upgrades the write to an atomic same-dir `NamedTempFile::persist` (a robustness win over the old direct `fs::write`), with a nice unix symlink-safety regression test.
- Prior review comments all addressed: `BlockTimingArgs` is now a single `#[command(flatten)]` (no 4× duplication), the concurrency test was renamed/clarified, and the CLI tests dropped the non-portable `true` and use explicit UTF-8.
- Tests are strong: 472 lib + 9 CLI (`test_node`) + 3 `testnode_api` + 5 doctests all pass, `cargo fmt --check` clean, docs build. The CLI/API integration tests drive a fake sequencer that records the exact config it was launched with, plus a Drop-based stop guard and exit-status forwarding checks.

**Non-blocking coordination note (re #234).** `patch_runtime_sequencer_config` reads the flat `SEQUENCER_CONFIG_REL_PATH`. That matches current `master`, so it's not a regression here — but #234 makes the sequencer/wallet config reads probe the nested `lez/…` layout for newer LEZ pins. When this rebases after #234 lands, please route this new helper through `first_existing_lez_path` / `SEQUENCER_CONFIG_REL_PATHS` so fast-block nodes don't re-break on nested pins. Whichever merges second should reconcile that single read site.

weboko:
Human review also LGTM

3esmit:
@weboko Rebased and addressed coordination note. PR 234 landed first. Thanks for the review.

weboko (commented):
## Re-review after the fresh commits (post-approval) — verified end-to-end ✅

The four commits that landed after the earlier approval (`feat/support block timing overrides`, docs/args refinement, utf-8 in embedded python, and the platform-command test fix) are all sound. I rebuilt the branch and drove it against a **real sequencer** (LEZ `cf3639d8`, r0vm 3.0.5, circuits v0.4.1) — not just the fake-sequencer unit path.

### Verified
- **Override actually speeds up a real chain.** `test-node start --block-create-timeout-ms 500 --retry-pending-blocks-timeout-ms 500` wrote `"500ms"` into the runtime `sequencer_config.json`, and block height advanced **7 blocks in ~4s** (~2 blk/s) vs the pinned 15s cadence (0 in 4s). The `"{ms}ms"` string is correctly parsed by the sequencer's `humantime_serde` config fields.
- **Omitting the flags preserves the pinned debug values** (`15s`/`5s`) — confirmed on a real node.
- **Validation is pre-spawn and layered:** clap `value_parser.range(1..=3_600_000)` rejects at the CLI, and `BlockTimingOverrides::validate()` rejects on the Rust API path before anything is spawned. Typed `BlockTimingValidationError` → `api::Error::BlockTimingValidation`.
- **Atomic, symlink-safe config write.** Factoring the two duplicate writers into `patch_runtime_sequencer_config` (write to a same-dir `NamedTempFile` + `persist`) is a real robustness/security improvement: no partial configs, and the new test proves a pre-planted predictable `.tmp` symlink is not followed. Existing runtime config is preserved on any validation/parse failure.
- **Tests:** `cargo fmt --check` clean; the block-timing/sequencer_config unit tests (25) and the 9 `test-node` CLI integration tests pass locally. CI `validate` is green.
- **Docs:** DOGFOODING T-series note + the good caveat that values near/below the stable-read sample cadence can keep `clock wait-stable` / account-boundary reads from converging.

No blocking findings. Clean, well-scoped, and well-tested. (Leaving this as a comment, not an approval.)

weboko:
@3esmit thanks for contribution!

PR #234, opened 2026-07-01
fix: support nested LEZ repo layout

### LINKED ISSUE #230: `lgs localnet start` hardcodes a pre-reorg LEZ path and cannot start a sequencer for recent `lez` pins
### Summary

`lgs localnet` reads the sequencer/wallet configs at paths that assume the *old*
flat LEZ repo layout:

- `constants.rs`: `SEQUENCER_CONFIG_REL_PATH = "sequencer/service/configs/debug/sequencer_config.json"`
- `state.rs`: `lez_repo.join("wallet/configs/debug/wallet_config.json")`

But the `lez` repo moved these under a `lez/` prefix
(`lez/sequencer/service/configs/debug/...`, `lez/wallet/configs/debug/...`).
This is already true at the rc5 testnet pin (`27360cb7`) **and** at `v0.2.0`
(`a58fbce2`). With either pin, `lgs localnet start` fails:

```
error: failed to read <lez>/sequencer/service/configs/debug/sequencer_config.json: No such file or directory
```

The latest `scaffold` commit still defaults `DEFAULT_LEZ` to `v0.1.2` (the
pre-reorg layout), so the breakage only appears when a project pins a newer
`lez` — which the public testnet requires.

### Workaround

Symlink the new layout back to the path `lgs` expects:

```bash
ln -sfn lez/sequencer <lez_cache>/sequencer
ln -sfn lez/wallet    <lez_cache>/wallet
```

### Suggested fix

Detect both layouts (probe `lez/sequencer/...` then `sequencer/...`), or make the
config-relative paths configurable in `scaffold.toml`, and bump `DEFAULT_LEZ` to a
pin matching the live testnet.

---

wyf027:
## Summary
- Support newer LEZ pins whose repository payload is nested under a `lez/` prefix.
- Add shared LEZ layout probing for sequencer debug config and wallet debug config.
- Keep the previous flat layout as a fallback for existing checkouts/tests.

## Fix / Approach
- Added `first_existing_lez_path(...)` to resolve the first available LEZ-relative candidate path with a clear error listing attempted paths.
- Updated localnet/testnode sequencer config setup to probe nested then flat paths.
- Updated wallet home seeding to probe nested then flat wallet config paths.
- Added regression tests for nested LEZ layouts in localnet, testnode, and wallet setup.

## Verification
RED before implementation:
- `cargo test prepare_sequencer_config_accepts_nested_lez_layout -- --nocapture` failed with missing flat sequencer config path.
- `cargo test write_node_config_accepts_nested_lez_layout -- --nocapture` failed with missing flat sequencer config path.
- `cargo test prepare_wallet_home_accepts_nested_lez_layout -- --nocapture` failed with missing wallet debug config.

After the fix:
- `cargo fmt --check` — passed.
- `cargo test prepare_sequencer_config_accepts_nested_lez_layout -- --nocapture` — passed.
- `cargo test write_node_config_accepts_nested_lez_layout -- --nocapture` — passed.
- `cargo test prepare_wallet_home_accepts_nested_lez_layout -- --nocapture` — passed.
- `cargo test --all-targets` — passed (`459` unit tests and `168` integration/CLI tests passed).
- `git diff --check` — passed.

Note: the existing `unexpected cfg condition value: sha2` warnings from `src/commands/setup.rs` still appear during tests and are unrelated to this change.

## Related issue
Fixes logos-co/scaffold#230

### DIFF
--- a/src/constants.rs
+++ b/src/constants.rs
@@ -56,6 +56,11 @@ pub(crate) const CIRCUITS_RELEASE_BASE_URL: &str =
 pub(crate) const DEFAULT_HELLO_WORLD_IMAGE_ID_HEX: &str =
     "4880b298f59699c1e4263c5c2245c80123632d608b9116f4b253c63e6c340771";
 pub(crate) const DEFAULT_WALLET_PASSWORD: &str = "logos-scaffold-v0";
+pub(crate) const WALLET_CONFIG_REL_PATH: &str = "wallet/configs/debug/wallet_config.json";
+pub(crate) const WALLET_CONFIG_NESTED_REL_PATH: &str =
+    "lez/wallet/configs/debug/wallet_config.json";
+pub(crate) const WALLET_CONFIG_REL_PATHS: &[&str] =
+    &[WALLET_CONFIG_NESTED_REL_PATH, WALLET_CONFIG_REL_PATH];
 pub(crate) const WALLET_BIN_REL_PATH: &str = "target/release/wallet";
 pub(crate) const FRAMEWORK_KIND_DEFAULT: &str = "default";
 pub(crate) const FRAMEWORK_KIND_LEZ_FRAMEWORK: &str = "lez-framework";
@@ -70,6 +75,10 @@ pub(crate) const SEQUENCER_BIN_REL_PATH: &str = "target/release/sequencer_servic
 pub(crate) const METHODS_DIR: &str = "methods";
 pub(crate) const SEQUENCER_CONFIG_REL_PATH: &str =
     "sequencer/service/configs/debug/sequencer_config.json";
+pub(crate) const SEQUENCER_CONFIG_NESTED_REL_PATH: &str =
+    "lez/sequencer/service/configs/debug/sequencer_config.json";
+pub(crate) const SEQUENCER_CONFIG_REL_PATHS: &[&str] =
+    &[SEQUENCER_CONFIG_NESTED_REL_PATH, SEQUENCER_CONFIG_REL_PATH];
 pub(crate) const SPEL_BIN_REL_PATH: &str = "target/release/spel";
 /// Default seconds to wait for the sequencer to become ready when `lgs run`
 /// has to start localnet itself. Cold first runs (fresh repo clone, cold
--- a/src/lez_layout.rs
+++ b/src/lez_layout.rs
@@ -1,0 +1,32 @@
+use std::path::{Path, PathBuf};
+
+use anyhow::bail;
+
+use crate::DynResult;
+
+/// Return the first existing LEZ-relative path from `rel_paths`.
+///
+/// LEZ has existed in both a flat repository layout (`sequencer/...`,
+/// `wallet/...`) and a nested layout (`lez/sequencer/...`, `lez/wallet/...`).
+/// Callers use this helper at the boundary where scaffold consumes files from a
+/// checked-out LEZ repo so both layouts stay supported without mutating the LEZ
+/// checkout or requiring symlinks.
+pub(crate) fn first_existing_lez_path(
+    lez: &Path,
+    rel_paths: &[&str],
+    label: &str,
+) -> DynResult<PathBuf> {
+    for rel in rel_paths {
+        let candidate = lez.join(rel);
+        if candidate.exists() {
+            return Ok(candidate);
+        }
+    }
+
+    let tried = rel_paths
+        .iter()
+        .map(|rel| lez.join(rel).display().to_string())
+        .collect::<Vec<_>>()
+        .join(", ");
+    bail!("missing {label} in lez repo; tried {tried}");
+}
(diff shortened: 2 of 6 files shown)


src/lez_layout.rs
```diff
@@ -0,0 +1,32 @@
...
+            return Ok(candidate);
+        }
+    }
+
+    let tried = rel_paths
+        .iter()
+        .map(|rel| lez.join(rel).display().to_string())
+        .collect::<Vec<_>>()
+        .join(", ");
+    bail!("missing {label} in lez repo; tried {tried}");
```
danisharora099, line 31: Optional: this error branch is the whole reason the helper exists, and nothing exercises it yet. A ~6-line test (point it at a temp dir with neither layout, assert the message names both candidate paths) would lock in the exact text a stuck user sees. Not a blocker.

src/constants.rs
```diff
@@ -56,6 +56,11 @@ pub(crate) const CIRCUITS_RELEASE_BASE_URL: &str =
 pub(crate) const DEFAULT_HELLO_WORLD_IMAGE_ID_HEX: &str =
     "4880b298f59699c1e4263c5c2245c80123632d608b9116f4b253c63e6c340771";
 pub(crate) const DEFAULT_WALLET_PASSWORD: &str = "logos-scaffold-v0";
+pub(crate) const WALLET_CONFIG_REL_PATH: &str = "wallet/configs/debug/wallet_config.json";
+pub(crate) const WALLET_CONFIG_NESTED_REL_PATH: &str =
+    "lez/wallet/configs/debug/wallet_config.json";
+pub(crate) const WALLET_CONFIG_REL_PATHS: &[&str] =
+    &[WALLET_CONFIG_NESTED_REL_PATH, WALLET_CONFIG_REL_PATH];
```
danisharora099, line 63: Minor: the order here is load-bearing, nested is tried before flat so it wins if both ever coexist. A one-line comment noting that would make the precedence read as intentional rather than incidental. Same goes for `SEQUENCER_CONFIG_REL_PATHS` below.

danisharora099 (approved):
Thanks for this, and for the thorough write-up. RED-before-green plus reproducing against both rc5 and v0.2.0 made it easy to trust the fix.

What I liked:

- Adding the fallback instead of moving `DEFAULT_LEZ` keeps the change backward-compatible and reversible. Existing v0.1.2 checkouts keep working, newer pins just start resolving. Good instinct to keep the two concerns separate.
- Pulling the probing into one `first_existing_lez_path` helper and applying it at all three read sites (localnet, testnode, wallet seeding) means we won't fix one and quietly leave another on the old path.
- The error now lists every path it tried, which is exactly what someone hitting #230 needs.
- Leaving the binaries alone is correct. Those resolve from the checkout root, so the payload reorg never touched them.

Both layouts are covered across all three call sites, so I'm happy to approve. Two small optional nits inline, neither blocks merge.

One follow-up worth its own issue (not this PR): the out-of-box `DEFAULT_LEZ` still points at flat v0.1.2, so a fresh project targeting the testnet has to override the pin until we bump it. This PR is the prerequisite for that bump, so the ordering is right, I just don't want us to lose track of the second half.

Approving. Thanks again for the contribution.

weboko (approved):
Reviewed and verified end-to-end. LGTM. 🟢

**What I checked**
- Built the branch clean (no new warnings), `cargo fmt --check` passes, `cargo clippy --lib` surfaces only the pre-existing `api::error` large-Err warnings (untouched here), and all 459 lib tests pass including the 3 new regression tests (`prepare_sequencer_config_accepts_nested_lez_layout`, `write_node_config_accepts_nested_lez_layout`, `prepare_wallet_home_accepts_nested_lez_layout`).
- **Validated the probe paths against the real repo.** Cloned `logos-execution-zone@v0.2.0` (`a58fbce2`, the nested pin from #230) and confirmed the configs live *only* at the probed candidates — `lez/sequencer/service/configs/debug/sequencer_config.json` and `lez/wallet/configs/debug/wallet_config.json` — while the old flat paths are genuinely absent (the exact root cause of #230). The nested-first ordering is therefore correct.
- **Completeness.** Grepped for every place scaffold consumes a config from the LEZ checkout; all three call sites (localnet `prepare_sequencer_config`, testnode `write_node_config`, `state::prepare_wallet_home`) now go through `first_existing_lez_path`. No production site still hardcodes a flat config path.
- **Binaries correctly left flat.** In v0.2.0 the workspace `Cargo.toml` is still at the repo root (members nested under `lez/`), so `sequencer_service`/`wallet` still build to `<lez>/target/release/`. The binary constants (`SEQUENCER_BIN_REL_PATH`, `WALLET_BIN_REL_PATH`) rightly did not need the same treatment.

**Design**: `first_existing_lez_path` is a clean, well-documented boundary helper, and the error message enumerating the attempted paths is a nice touch for debuggability.

**Non-blocking note**: `DEFAULT_LEZ` is still `v0.1.2` (flat), so this forward-compat path isn't exercised by the default template yet — the bump #230 also suggested is reasonably left as separate scope (it needs a matching spel + circuits release per the cross-framework invariant in `constants.rs`). No DOGFOODING.md change needed — this is a compatibility fix, not new user-facing surface.

weboko:
Human PoV also looks good :) 

Thanks for contribution @wyf027 !

PR #238, opened 2026-07-21
fix(basecamp): set absolute LOGOS_DATA_DIR on macOS portable launch

### LINKED ISSUE #236: basecamp launch: macOS bin-macos-app ignores XDG; needs absolute LOGOS_DATA_DIR to load profile modules + render UI
## Summary

On macOS, `lgs basecamp launch <profile>` isolates per-profile state exclusively via `XDG_DATA_HOME` (`launch_env`, `src/commands/basecamp.rs:712`). But a `bin-macos-app` Basecamp bundle **ignores XDG on macOS** — it loads its modules/plugins from `LOGOS_DATA_DIR`, falling back to `~/Library/Application Support/Logos/LogosBasecamp/` when that variable is unset. So scaffold installs the profile's modules under `<profile>/xdg-data/Logos/LogosBasecamp/{modules,plugins}` (`profile_modules_and_plugins`, `basecamp.rs:2436`), the app boots, but it never sees them: only its built-in `package_manager` + `capability_module` load, and the project's own modules are absent.

## Environment

Found while dogfooding the `eth-lez-atomic-swaps` Phase 2 scaffold migration on macOS `darwin-arm64` (Apple Silicon).

- Scaffold `7c52211a3f40a6ac5829905d4569712f414776ed` (== current `master` HEAD)
- Basecamp `a746cdbc` (= `v0.1.1`), attr `bin-macos-app`
- LGPM `e5c25989`, attr `cli-portable`

This Basecamp build predates the `--user-dir` / `LOGOS_USER_DIR` flag (see the note added in #178), so the older `LOGOS_DATA_DIR` mechanism is the only lever available on it.

## Repro

1. A project with `[repos.basecamp].attr = "bin-macos-app"` and one or more `[modules.*]` captured.
2. `lgs basecamp setup && lgs basecamp install` — modules land under `<profile>/xdg-data/Logos/LogosBasecamp/modules/`.
3. `lgs basecamp launch alice`.
4. Basecamp comes up but the project's modules are not loaded (only `package_manager` + `capability_module`); the app read from `~/Library/Application Support/...`, not the profile's `XDG_DATA_HOME`.

## Root cause

`launch_env` sets `XDG_CONFIG_HOME` / `XDG_DATA_HOME` / `XDG_CACHE_HOME` / `TMPDIR` / `LOGOS_PROFILE`, but never `LOGOS_DATA_DIR`. The dev (non-portable) Basecamp honors XDG on both platforms, so this only bites the macOS portable (`bin-macos-app`) stack.

### Critical subtlety — the value must be **absolute**

Setting a *relative* `LOGOS_DATA_DIR` loads the backend modules but then the dlopen'd `main_ui` / `package_manager_ui` dylibs fail `@rpath` resolution (`"shared library was not found"`) and the shell UI never renders. Only an **absolute** `LOGOS_DATA_DIR` makes the whole portable stack (`main_ui` + the project modules + their UI plugins) load and render.

## Downstream impact

Because scaffold can't portably express an absolute per-profile path in a committed `scaffold.toml`, the downstream repo currently ships a launch bridge (`.context/bc-launch.sh`) that re-implements what `launch` should do: it execs the Basecamp binary with an **absolute** `LOGOS_DATA_DIR = <profile>/xdg-data/Logos/LogosBasecamp` plus the profile's XDG/runtime env. With that in place, both peers load the full portable stack. This glue exists purely to work around the missing launch behavior.

## Proposed fix

In `cmd_basecamp_launch`, on macOS when the configured Basecamp is portable (`is_portable_basecamp(basecamp_repo)`), set an **absolute** `LOGOS_DATA_DIR` = `<profile_dir>/xdg-data/<basecamp_xdg_subpath(basecamp_repo)>` — the exact directory scaffold already installs into (`Logos/LogosBasecamp` for the portable stack). Additionally, resolve any caller-supplied relative `LOGOS_DATA_DIR` (from `[basecamp.env]` / `[basecamp.profiles.<name>.env]`) to absolute against the project root, so a relative value can't silently break the UI. No-op on non-macOS and for the dev stack (which honors XDG).

Happy to PR this (the layout constants and portable-detection helpers already exist in `basecamp.rs`).

danisharora099:

Found while dogfooding the [`eth-lez-atomic-swaps`](https://github.com/logos-co/eth-lez-atomic-swaps) Phase 2 scaffold migration — a two-peer atomic-swap app running the distributed (portable) Basecamp stack on macOS `darwin-arm64`. Fixes #236.

## User Story

`lgs basecamp launch <profile>` isolates per-profile state via `XDG_DATA_HOME`, but the `bin-macos-app` Basecamp bundle (`a746cdbc` / `v0.1.1`) **ignores XDG on macOS** — it reads modules/plugins from `LOGOS_DATA_DIR`, falling back to `~/Library/Application Support/Logos/LogosBasecamp/` when unset. So scaffold installs the profile's modules under `<profile>/xdg-data/Logos/LogosBasecamp/{modules,plugins}` but the launched app never sees them: only its built-in `package_manager` + `capability_module` load.

Critically, the value **must be absolute**: a *relative* `LOGOS_DATA_DIR` loads the backend modules but then the dlopen'd `main_ui` / `package_manager_ui` dylibs fail `@rpath` resolution (`"shared library was not found"`) and the shell UI never renders. Only an absolute path brings up the full stack.

Because scaffold can't portably express an absolute per-profile path in a committed `scaffold.toml`, the downstream repo had to ship a launch bridge that re-implements this — exec'ing Basecamp with an absolute `LOGOS_DATA_DIR` plus the profile's XDG/runtime env. With that in place both peers load the full portable stack (`main_ui` + the swap modules + their UI plugins). This PR moves that behavior into `launch` so the glue can be deleted.

## Change Summary

On the macOS portable stack (`cfg!(target_os = "macos")` + `is_portable_basecamp`), `cmd_basecamp_launch` now sets `LOGOS_DATA_DIR` to the profile's **absolute** module root (`<profile>/xdg-data/<basecamp_xdg_subpath>` — the exact dir scaffold already installs into), and absolutizes any caller-supplied *relative* `LOGOS_DATA_DIR` from `[basecamp.env]` / `[basecamp.profiles.<name>.env]` against the project root. No-op on non-macOS and for the dev (XDG-honoring) stack. The transform is applied after the scaffold.toml env layering so caller overrides are still respected (just absolutized).

## Verification

- `cargo test --lib` — 511 passed, 0 failed (3 new unit tests for `set_absolute_logos_data_dir`: default-when-unset, relative→absolute, absolute-untouched).
- Downstream: verified end-to-end in the atomic-swaps repo — both maker + taker peers launch and load the full portable stack (`main_ui` + `swap` + `swap_ui` + `delivery_module`) with the absolute `LOGOS_DATA_DIR` this PR now sets.

## Checklist

- [x] CI is green
- [x] Relevant DOGFOODING scenarios rerun (macOS portable two-peer launch, downstream atomic-swaps)
- [ ] Docs updated (no user-facing config surface changed; behavior is automatic on the macOS portable stack)

### DIFF
--- a/DOGFOODING.md
+++ b/DOGFOODING.md
@@ -1156,6 +1156,7 @@ Within the running UIs, exercise whatever p2p surface the module exposes (chat e
 - Custom profile pairs open against their own profile dirs under `.scaffold/basecamp/profiles/<profile>/` and their configured runtime/log/env paths.
 - Each window shows the project's `.lgx` modules installed and ready.
 - `LOGOS_PROFILE=alice` and `LOGOS_PROFILE=bob` are visible in each respective process environment (helpful for debugging).
+- On the macOS portable stack, each process environment carries an absolute `LOGOS_DATA_DIR` pointing at its own profile's module root — set automatically by `launch`, no manual export needed.
 - The two instances do not collide on Qt remote-objects or any non-module port; per-profile port-override env vars (per the spec) are set on each `launch`.
 - A p2p interaction triggered from `alice` is observable in `bob` (and vice versa) within the module's expected latency window.
 
--- a/src/commands/basecamp.rs
+++ b/src/commands/basecamp.rs
@@ -784,7 +784,9 @@ fn launch_env(
 ///     `.../Logos/LogosBasecamp` for the portable stack) when unset, and
 ///   * rewrites any caller-supplied relative `LOGOS_DATA_DIR` (from
 ///     `[basecamp.env]` / `[basecamp.profiles.<name>.env]`) to absolute against
-///     the project root, so a relative value can't silently break the UI.
+///     the project root, so a relative value can't silently break the UI, and
+///   * treats an empty (or whitespace-only) caller value as unset — absolutizing
+///     `""` would collapse to the project root, which is not a module root.
 ///
 /// The caller gates this to the macOS portable stack; the transform itself is
 /// platform-independent so it stays unit-testable on any host.
@@ -795,7 +797,10 @@ fn set_absolute_logos_data_dir(
     basecamp_repo: Option<&RepoRef>,
 ) {
     const KEY: &str = "LOGOS_DATA_DIR";
-    match env.get(KEY) {
+    match env
+        .get(KEY)
+        .filter(|v| !v.to_string_lossy().trim().is_empty())
+    {
         Some(existing) => {
             let path = PathBuf::from(existing);
             if path.is_relative() {
@@ -816,13 +821,19 @@ fn set_absolute_logos_data_dir(
 
 /// Resolve `path` to an absolute path. Absolute inputs pass through unchanged; a
 /// relative input is joined onto `base` (the project root, itself absolute for
-/// the normal `load_project` discovery path).
+/// the normal `load_project` discovery path). A relative `base` (possible via
+/// `load_project_at`, which accepts any root) is best-effort canonicalized
+/// first so the result is absolute whenever `base` exists on disk.
 fn absolutize(base: &Path, path: &Path) -> PathBuf {
     if path.is_absolute() {
-        path.to_path_buf()
-    } else {
-        base.join(path)
+        return path.to_path_buf();
     }
+    if base.is_relative() {
+        if let Ok(abs_base) = base.canonicalize() {
+            return abs_base.join(path);
+        }
+    }
+    base.join(path)
 }
 
 /// Resolve the per-profile runtime dir (`TMPDIR` / `XDG_RUNTIME_DIR` root).
@@ -3813,6 +3824,41 @@ mod tests {
         assert_eq!(
             env.get("LOGOS_DATA_DIR").map(PathBuf::from),
             Some(PathBuf::from("/somewhere/else"))
+        );
+    }
+
+    #[test]
+    fn set_absolute_logos_data_dir_treats_empty_caller_value_as_unset() {
+        let portable = repo_with_attr("bin-macos-app");
+        let project_root = Path::new("/abs/project");
+        let profile_dir = project_root.join("profiles/alice");
+        let mut env: BTreeMap<String, OsString> = BTreeMap::new();
+        // An empty value (e.g. `LOGOS_DATA_DIR = ""` in [basecamp.env]) would
+        // absolutize to the project root itself, which is not a module root;
+        // it must fall back to the profile default instead.
+        env.insert("LOGOS_DATA_DIR".into(), OsString::new());
+        set_absolute_logos_data_dir(&mut env, project_root, &profile_dir, Some(&portable));
+        assert_eq!(
+            env.get("LOGOS_DATA_DIR").map(PathBuf::from),
+            Some(
+                profile_dir
+                    .join("xdg-data")
+                    .join(BASECAMP_XDG_APP_SUBPATH_PORTABLE)
+            )
+        );
+    }
+
+    #[test]
+    fn absolutize_canonicalizes_relative_base() {
+        // `.` always exists, so canonicalize resolves it against the test cwd
+        // without the test itself having to change directories.
+        let resolved = absolutize(Path::new("."), Path::new("custom/data"));
+        assert!(resolved.is_absolute(), "got: {}", resolved.display());
+        assert_eq!(
+            resolved,
+            std::env::current_dir()
+                .expect("test cwd")
+                .join("custom/data")
         );
     }
 
--- a/src/commands/basecamp.rs
+++ b/src/commands/basecamp.rs
@@ -514,6 +514,13 @@ fn cmd_basecamp_launch(
             std::env::var_os(k)
         });
     }
+    // macOS `bin-macos-app` Basecamp ignores XDG and loads its modules from
+    // `LOGOS_DATA_DIR` (see `set_absolute_logos_data_dir`). Apply this after the
+    // scaffold.toml env layering so a user-supplied relative value is absolutized
+    // too. Scoped to the macOS portable stack; a no-op elsewhere.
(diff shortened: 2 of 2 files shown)


weboko (commented):
## Review — full code read + unit tests (macOS-gated, so runtime-verified only where possible on Linux)

I reviewed this by reading the change in context and running the suite. Because the whole behavior is gated on `cfg!(target_os = "macos") && is_portable_basecamp(...)`, on my Linux host the new code path is a compile-checked no-op — I could not exercise the actual macOS `bin-macos-app` launch. I relied on the unit tests + the surrounding invariants for correctness, and on your downstream atomic-swaps verification (noted in the PR) for the real end-to-end.

### What I verified
- **Builds/tests:** `cargo test --lib` → **511 passed / 0 failed**, including the 3 new `set_absolute_logos_data_dir` tests (default-when-unset, relative→absolute, absolute-untouched); `cargo fmt --check` clean.
- **Path correctness:** the default `LOGOS_DATA_DIR = <profile>/xdg-data/<basecamp_xdg_subpath()>` is exactly where scaffold installs a portable profile's modules — `seed_profiles` seeds `<profile>/xdg-data/BASECAMP_XDG_APP_SUBPATH_PORTABLE` (see `seed_profiles_uses_portable_subpath_when_basecamp_is_portable`), and `XDG_DATA_HOME` is set to `<profile>/xdg-data`. So the finalizer points the app at the same tree XDG would resolve to. 
- **Ordering:** applying the transform *after* the `scaffold.toml` env layering (so a caller-supplied relative override is absolutized rather than ignored) is the right call and matches the comment.

### Security / privacy
No concerns. It only derives a local filesystem path from the existing profile dir + project root; no new external input, network, or secret handling.

### Non-blocking nits (both echo Copilot; I confirmed them by reading)
1. **Empty `LOGOS_DATA_DIR = ""`** (e.g. from `[basecamp.env]`) lands in the `Some(existing)` arm, is `is_relative() == true`, and `base.join("")` collapses to the **project root** — not a valid module/plugin root, so it would silently misconfigure the launch. Suggest treating an empty value as *unset* (fall through to the module-root default) and adding a test for it. One-liner: `if existing.is_empty() { /* default */ } else { ... }`.
2. **`absolutize` with a relative `base`:** if the project root is ever relative (possible via `load_project_at`, which doesn't require an absolute root), the result stays relative even though `LOGOS_DATA_DIR` must be absolute. The doc-comment acknowledges this and the normal `load_project` path yields an absolute root, so it's low-risk — a best-effort `canonicalize()` of a relative `base` would harden it.

### Docs
No `README`/`DOGFOODING` change, which is defensible since there's no new config surface and the behavior is automatic. Since it does change observable launch behavior on the macOS portable stack, a one-line note (e.g. in the DOGFOODING **B3** launch scenario, or a pointer comment for macOS portable users) wouldn't hurt — optional.

Well-scoped, well-documented in-code (the `set_absolute_logos_data_dir` doc-comment explaining *why absolute* and the `@rpath` failure mode is excellent), and the reasoning is sound. LGTM from my side, modulo the empty-string nit which is worth a quick follow-up. Leaving the formal approval to you.

danisharora099:
Addressed both nits in febb74b.

Empty `LOGOS_DATA_DIR` is now treated as unset, so it falls back to the profile module root instead of resolving to the project root. Added a test for it.

For absolutize, a relative base now gets canonicalized (best effort) before the join, and falls back to the plain join if that fails. Updated the doc comment and added a test that uses . as the base so it doesn't need to mess with the cwd.

Also added the one-liner to B3's success signals about launch setting the absolute `LOGOS_DATA_DIR` automatically.

`cargo test --lib` is at 513 passing with the two new tests.

@weboko thanks for the careful read. Since the nits are in, can you drop a formal approval when you get a chance?

weboko (commented):
## Re-review — follow-up commit resolves both nits from my prior review ✅

Re-reviewed the fresh commit `febb74b` ("treat empty `LOGOS_DATA_DIR` as unset and harden absolutize"), which landed after my 2026-07-21 review. It resolves both non-blocking items I raised, cleanly and with test coverage.

### Verified locally (Linux — Rust 1.94.1, Nix 2.35.1 w/ flakes, r0vm 3.0.5 matching pinned LEZ v0.1.2)
- **Builds clean, `cargo fmt --check` clean, `cargo test --lib` → 513 passed / 0 failed** — including the 2 new tests:
  - `set_absolute_logos_data_dir_treats_empty_caller_value_as_unset`
  - `absolutize_canonicalizes_relative_base`
- The behavior stays gated on `cfg!(target_os = "macos") && is_portable_basecamp(...)`, so on a Linux host it is a compile-checked no-op — I could not exercise the real `bin-macos-app` launch (same limitation as my last pass; the downstream atomic-swaps run covers the real macOS end-to-end).

### Both nits are addressed correctly
1. **Empty `LOGOS_DATA_DIR = ""`** — now filtered via `.filter(|v| !v.to_string_lossy().trim().is_empty())` before the `Some(existing)` arm, so an empty *or whitespace-only* caller value falls through to the profile module-root default instead of collapsing to the project root via `base.join("")`. Nice touch covering whitespace-only too; the new test asserts the fall-through to `<profile>/xdg-data/<portable subpath>`.
2. **Relative `base` in `absolutize`** — now best-effort `canonicalize()`d before joining, so a relative project root (reachable via `load_project_at`) still yields an absolute `LOGOS_DATA_DIR`. When a relative base doesn't exist on disk it falls back to `base.join(path)` (stays relative) — acceptable and documented, since the normal `load_project` discovery path is always absolute.

### Docs
- `DOGFOODING.md` **B3** now notes the automatic absolute `LOGOS_DATA_DIR` on the macOS portable stack — the one-line note I suggested.

No further blocking findings — this is exactly the follow-up I asked for. (Comment only — not an approval.)

weboko (approved):

Issue #89, opened 2026-04-30
`lgs basecamp launch`: set per-profile `TMPDIR` to prevent cross-profile basecamp SIGSEGV in QRO `onClientRead`


fryorcraken:
# `lgs basecamp launch` should set per-profile `TMPDIR` to prevent cross-profile basecamp SIGSEGV

## Symptom

Running `lgs basecamp launch alice` and `lgs basecamp launch bob` concurrently on the same Linux host (so both profiles eventually load the same module — e.g. `delivery_module`) causes one or both basecamp processes to crash with `SIGSEGV` shortly after the second profile starts spawning that module's `logos_host` child. The crashing thread is always inside Qt RemoteObjects' `onClientRead` path:

```
#0  QMetaObject::methodOffset()                    ← null/dangling vtable read
#1  QMetaObject::method(int)
#2  QRemoteObjectNodePrivate::onClientRead(QObject*)
#3  doActivate<false>(QObject*, int, void**)
#6  QAbstractSocketPrivate::canReadNotification()
#7  QReadNotifier::event(QEvent*)
#8  QApplicationPrivate::notify_helper(QObject*, QEvent*)
#9  QCoreApplication::notifyInternal2(QObject*, QEvent*)
#10 socketNotifierSourceDispatch(...)
…
#16 QConnectedReplicaImplementation::waitForSource(int)
#17 LogosAPIConsumer::requestObject(...)
#18 LogosAPIConsumer::invokeRemoteMethod(...)
#19 LogosAPIClient::invokeRemoteMethod(...)
#20 LogosQmlBridge::callModule(...)               ← basecamp main_ui plugin
```

I observed this 3 times in a row across several launch orderings; the second instance to come up tends to take the first one with it within 1–2 seconds of its own modules spawning.

## Root cause (what I traced)

`liblogos_core` exposes each `logos_host` child to basecamp via a `QLocalServer`-bound socket whose name is hardcoded as `logos_token_<modulename>` (no random suffix). I confirmed this by stringsing `liblogos_core.so` (`logos_token_%1`).

Qt's `QLocalServer` resolves that name through `QStandardPaths::TempLocation`, which on Linux is the value of `TMPDIR`, falling back to `/tmp`. `lgs basecamp launch` exports `XDG_CONFIG_HOME / XDG_DATA_HOME / XDG_CACHE_HOME` per profile but does **not** export `TMPDIR`. So both profiles end up pointing at `/tmp/logos_token_<module>`.

[...]

PR #244, opened 2026-07-30
fix(basecamp): set absolute LOGOS_USER_DIR for Basecamp 0.2.x on macOS portable launch

### LINKED ISSUE #242: basecamp launch: set LOGOS_USER_DIR for Basecamp 0.2.x (renamed data-dir override; profiles collapse onto the shared dir without it)
## Problem

`lgs basecamp launch` isolates each profile's Basecamp data tree by exporting an absolute per-profile `LOGOS_DATA_DIR` (#236, fixed by #238). That env var is a Basecamp **0.1.x** name.

Basecamp 0.2.x renamed the override: [logos-basecamp#164](https://github.com/logos-co/logos-basecamp/pull/164) (commit `9bc9114`, "allow overriding user directory via launch argument") replaced it with `LOGOS_USER_DIR` / `--user-dir`, resolved in `app/utils/LogosBasecampPaths.h` `baseDirectory()`. No `LOGOS_DATA_DIR` reference survives anywhere in the 0.2.x tree.

Because Qt's `AppDataLocation` ignores `XDG_DATA_HOME` on macOS, the failure against a 0.2.x pin is silent and total: every profile launched via `lgs basecamp launch` collapses onto the shared `~/Library/Application Support/Logos/LogosBasecamp`, none of the project's per-profile modules appear (only the embedded bundle modules load), per-profile isolation for multi-peer flows is lost, and any stale 0.1.x-era dylibs left in the shared tree get loaded into the new host.

## Subtlety: Basecamp consumes the env var verbatim

Basecamp 0.2.x absolutizes the `--user-dir` **flag** (`app/main.cpp` runs it through `QFileInfo::absoluteFilePath()` and `qputenv`s it) but consumes the `LOGOS_USER_DIR` **env var** verbatim (`LogosBasecampPaths.h:baseDirectory()` returns it as-is). A relative env value therefore scatters state under the app's cwd. Scaffold must export an **absolute** path — exactly the discipline `set_absolute_logos_data_dir` (#238) already applies to the old key.

## Proposed fix

In `cmd_basecamp_launch`, where `set_absolute_logos_data_dir` is applied (macOS portable stack), also ensure `LOGOS_USER_DIR` is set with identical semantics:

- default to the profile's absolute module root (`<project>/.scaffold/basecamp/profiles/<profile>/xdg-data/Logos/LogosBasecamp` for the portable stack) when unset — the same path #238 computes for `LOGOS_DATA_DIR`;
- absolutize any caller-supplied relative value (from `[basecamp.env]` / `[basecamp.profiles.<name>.env]`) against the project root;
- treat an empty value as unset.

Keep emitting `LOGOS_DATA_DIR` so projects on 0.1.x pins are unaffected; the two keys are read by disjoint basecamp versions, so setting both is harmless and version-agnostic.

## Downstream evidence

Hit in `eth-lez-atomic-swaps` (tracker entry TR-21) while bumping its basecamp pin `a746cdbc` (v0.1.1) → `8a36a839` (0.2.1), verified live on aarch64-darwin:

- Unbridged `lgs basecamp launch maker` logs `Base data directory: ~/Library/Application Support/Logos/LogosBasecamp` and loads only the embedded modules; the project's `swap` / `swap_ui` / `delivery_module` are absent.
- With `LOGOS_USER_DIR` exported (absolute, per profile) the app logs the profile path, `setUserUiPluginsDirectory` receives the profile's `plugins` dir, and the project modules load.
- The repo currently bridges this with a Makefile target (`basecamp-launch-%`) that prepends `LOGOS_USER_DIR=<profile module root>` to `lgs basecamp launch` — verified against basecamp 0.2.1 with a full two-peer GUI swap on 2026-07-30. This issue is about retiring that bridge.

## Relationship to #236 / #238

Same bug class, other side of the version boundary: #236/#238 fixed profile isolation for the 0.1.x key on the macOS portable stack; this extends the same mechanism to the 0.2.x key so `lgs basecamp launch` works across both basecamp generations.

danisharora099:
Closes #242.

## Problem

`lgs basecamp launch` exports an absolute per-profile `LOGOS_DATA_DIR` (#236 / #238) — a Basecamp **0.1.x** name. Basecamp 0.2.x renamed the data-tree override to `LOGOS_USER_DIR` / `--user-dir` (logos-basecamp#164, resolved in `LogosBasecampPaths.h::baseDirectory()`) and drops the old key entirely. Since Qt's `AppDataLocation` ignores `XDG_DATA_HOME` on macOS, launching against a 0.2.x pin silently collapses every profile onto the shared `~/Library/Application Support/Logos/LogosBasecamp`: only the embedded bundle modules load, none of the project's modules appear, and multi-peer profile isolation is gone.

Extra subtlety: basecamp absolutizes the `--user-dir` **flag** (`QFileInfo::absoluteFilePath()` + `qputenv`) but consumes the `LOGOS_USER_DIR` **env var** verbatim, so a relative env value scatters state under the app's cwd. Scaffold must export an absolute path, exactly as #238 already does for the old key.

## Fix

Generalizes `set_absolute_logos_data_dir` into `set_absolute_basecamp_data_dirs`, which applies the identical per-key transform (private helper `set_absolute_module_root_var`) to both `LOGOS_DATA_DIR` and `LOGOS_USER_DIR` independently:

- default to the profile's absolute module root (`<profile>/xdg-data/<basecamp_xdg_subpath>`) when unset;
- absolutize a caller-supplied relative value (from `[basecamp.env]` / `[basecamp.profiles.<name>.env]`) against the project root;
- treat an empty/whitespace-only value as unset.

The two keys are read by disjoint basecamp generations, so setting both keeps the launch pin-agnostic — projects on 0.1.x pins see no behavior change.

### Gating rationale

Kept #238's gate unchanged (`cfg!(target_os = "macos") && is_portable_basecamp(...)`): the profile-collapse failure only exists where `XDG_DATA_HOME` is ignored, i.e. the macOS portable stack. On Linux, Qt honors `XDG_DATA_HOME`, so `launch_env`'s existing XDG isolation already covers both basecamp generations there. The transform itself stays platform-independent for unit-testability, as before.

## Tests

The four `set_absolute_logos_data_dir_*` tests are extended into six `set_absolute_basecamp_data_dirs_*` tests: default-both-keys-when-unset, absolutize-relative for each key, absolute-values-untouched for both, empty-treated-as-unset for both, and a per-key independence test (one key overridden relative, the other unset). `cargo fmt --check`, `cargo check`, and `cargo test` (709 tests) all pass locally.

## Downstream verification

Verified in `eth-lez-atomic-swaps` (tracker entry TR-21; basecamp pin `8a36a839`, tag 0.2.1, aarch64-darwin), which until now bridged this with a `LOGOS_USER_DIR=... lgs basecamp launch` Makefile wrapper — the wrapped stack was verified there with a full two-peer GUI swap on 2026-07-30. With this branch's `lgs` installed (`cargo install --locked`), a **bare** `lgs basecamp launch maker --log-file` (no wrapper):

`ps eww <app pid>`:

```
LOGOS_DATA_DIR=/Users/.../eth-lez-atomic-swaps/nashville/.scaffold/basecamp/profiles/maker/xdg-data/Logos/LogosBasecamp
LOGOS_USER_DIR=/Users/.../eth-lez-atomic-swaps/nashville/.scaffold/basecamp/profiles/maker/xdg-data/Logos/LogosBasecamp
LOGOS_PROFILE=maker
```

Profile `basecamp.log` (previously this line read the shared `~/Library/Application Support/Logos/LogosBasecamp` — the failure signature):

```
Base data directory: /Users/.../nashville/.scaffold/basecamp/profiles/maker/xdg-data/Logos/LogosBasecamp
... ModuleProxy: callRemoteMethod "setUserModulesDirectory" args: QList(QVariant(QString, ".../profiles/maker/xdg-data/Logos/LogosBasecamp/modules"))
... ModuleProxy: callRemoteMethod "setUserUiPluginsDirectory" args: QList(QVariant(QString, ".../profiles/maker/xdg-data/Logos/LogosBasecamp/plugins"))
... ModuleProxy: callRemoteMethod "getInstalledUiPlugins" args: QList()
```

Zero `shared library was not found` occurrences. Enumerating the exact dirs the app was pointed at (bundled 0.2.1 lgpm, the same enumeration source `package_manager` reads):

```
NAME                           VERSION         TYPE       CATEGORY
delivery_module                1.1.0           core       protocol
swap                           0.2.0           core       defi
swap_ui                        0.2.0           ui_qml     defi
```

`swap_ui`'s installed manifest carries `"view": "qml/Main.qml"`, so it passes 0.2.x's `UIPluginManager` `ui_qml` empty-view filter and shows up in the launcher.

### DIFF
--- a/DOGFOODING.md
+++ b/DOGFOODING.md
@@ -1171,7 +1171,7 @@ Within the running UIs, exercise whatever p2p surface the module exposes (chat e
 - Each window shows the project's `.lgx` modules installed and ready.
 - `LOGOS_PROFILE=alice` and `LOGOS_PROFILE=bob` are visible in each respective process environment (helpful for debugging).
 - On the macOS portable stack, each process environment carries an absolute `LOGOS_DATA_DIR` *and* an absolute `LOGOS_USER_DIR`, both pointing at that profile's own module root (`.scaffold/basecamp/profiles/<profile>/xdg-data/Logos/LogosBasecamp`) — set automatically by `launch`, no manual export needed. Basecamp 0.1.x reads the first and 0.2.x the second, so which one the app actually honors depends on the `[repos.basecamp]` pin; `launch` sets both so the check is the same either way.
-- The two instances do not collide on Qt remote-objects or any non-module port; per-profile port-override env vars (per the spec) are set on each `launch`.
+- The two instances do not collide on Qt remote-objects or any non-module port. Do not go hunting for per-module port-override env vars in the process environment: the registry they would flow in through is empty in v1 (no module has published a name yet), so `launch` exports none. A module-level port collision between `alice` and `bob` is therefore still possible, and belongs to the owning module rather than to scaffold.
 - A p2p interaction triggered from `alice` is observable in `bob` (and vice versa) within the module's expected latency window.
 
 ### Failure Signals / Common Pitfalls
@@ -1187,7 +1187,9 @@ Within the running UIs, exercise whatever p2p surface the module exposes (chat e
 
 - The exact two-terminal command sequence used.
 - A short transcript or screenshot pair showing a p2p interaction propagating from one instance to the other.
-- The env block of each running process (e.g., `tr '\0' '\n' < /proc/<pid>/environ | grep -E 'XDG_|LOGOS_'`).
+- The env block of each running process. `launch` records the basecamp PID in `.scaffold/basecamp/profiles/<profile>/launch.state` as `pid=<pid>`; read the env with:
+  - Linux: `tr '\0' '\n' < /proc/<pid>/environ | grep -E 'XDG_|LOGOS_'`
+  - macOS (no `/proc`): `ps eww -p <pid> | tr ' ' '\n' | grep -E 'XDG_|LOGOS_'`
 - Any port-collision error text verbatim, with the module that owns the colliding port.
 - If custom profiles are used, `basecamp paths <profile> --json` for each profile and the resolved log/runtime dirs.
 
--- a/DOGFOODING.md
+++ b/DOGFOODING.md
@@ -1179,7 +1179,7 @@ Within the running UIs, exercise whatever p2p surface the module exposes (chat e
 - Two windows opening but sharing identity keys, profile state, or message history is a clean-slate / XDG-isolation regression.
 - On the macOS portable stack, both windows showing only basecamp's bundled modules — none of the project's `.lgx` modules — while their logs report a base data directory under the shared `~/Library/Application Support/Logos/LogosBasecamp` is the profile-collapse signature: the app is not reading the per-profile module root at all. Check that both `LOGOS_DATA_DIR` and `LOGOS_USER_DIR` are present, absolute, and distinct per profile in each process environment.
 - A non-module port collision (Qt remote objects, etc.) is a real finding — file upstream against the affected component, do not patch around it inside scaffold.
-- A module that does not honor an externally-provided port override is documented as a known gap pending an upstream fix on that module; capture the module name, the env var that should have worked, and the observed collision.
+- A module that hardcodes its port is a known gap pending an upstream fix on that module. Scaffold exports no override for it to honor (see the signal above), so capture the module name and the observed collision — not a missing env var.
 - One window crashing while the other survives is recordable evidence; capture the crashing instance's logs from `.scaffold/basecamp/profiles/<name>/` before relaunching.
 - Running `basecamp launch alice` twice in parallel is undefined in v1 — record the behavior if you trip it accidentally, but don't treat it as a supported scenario.
 
@@ -1189,7 +1189,7 @@ Within the running UIs, exercise whatever p2p surface the module exposes (chat e
 - A short transcript or screenshot pair showing a p2p interaction propagating from one instance to the other.
 - The env block of each running process. `launch` records the basecamp PID in `.scaffold/basecamp/profiles/<profile>/launch.state` as `pid=<pid>`; read the env with:
   - Linux: `tr '\0' '\n' < /proc/<pid>/environ | grep -E 'XDG_|LOGOS_'`
-  - macOS (no `/proc`): `ps eww -p <pid> | tr ' ' '\n' | grep -E 'XDG_|LOGOS_'`
+  - macOS (no `/proc`): `ps eww -p <pid>` — read the `XDG_*` / `LOGOS_*` assignments off the line directly rather than splitting on spaces, since the profile-collapse target (`~/Library/Application Support/…`) contains one and would be truncated at `Application`.
 - Any port-collision error text verbatim, with the module that owns the colliding port.
 - If custom profiles are used, `basecamp paths <profile> --json` for each profile and the resolved log/runtime dirs.
 
--- a/DOGFOODING.md
+++ b/DOGFOODING.md
@@ -1170,13 +1170,14 @@ Within the running UIs, exercise whatever p2p surface the module exposes (chat e
 - Custom profile pairs open against their own profile dirs under `.scaffold/basecamp/profiles/<profile>/` and their configured runtime/log/env paths.
 - Each window shows the project's `.lgx` modules installed and ready.
 - `LOGOS_PROFILE=alice` and `LOGOS_PROFILE=bob` are visible in each respective process environment (helpful for debugging).
-- On the macOS portable stack, each process environment carries an absolute `LOGOS_DATA_DIR` pointing at its own profile's module root — set automatically by `launch`, no manual export needed.
+- On the macOS portable stack, each process environment carries an absolute `LOGOS_DATA_DIR` *and* an absolute `LOGOS_USER_DIR`, both pointing at that profile's own module root (`.scaffold/basecamp/profiles/<profile>/xdg-data/Logos/LogosBasecamp`) — set automatically by `launch`, no manual export needed. Basecamp 0.1.x reads the first and 0.2.x the second, so which one the app actually honors depends on the `[repos.basecamp]` pin; `launch` sets both so the check is the same either way.
 - The two instances do not collide on Qt remote-objects or any non-module port; per-profile port-override env vars (per the spec) are set on each `launch`.
 - A p2p interaction triggered from `alice` is observable in `bob` (and vice versa) within the module's expected latency window.
 
 ### Failure Signals / Common Pitfalls
 
 - Two windows opening but sharing identity keys, profile state, or message history is a clean-slate / XDG-isolation regression.
+- On the macOS portable stack, both windows showing only basecamp's bundled modules — none of the project's `.lgx` modules — while their logs report a base data directory under the shared `~/Library/Application Support/Logos/LogosBasecamp` is the profile-collapse signature: the app is not reading the per-profile module root at all. Check that both `LOGOS_DATA_DIR` and `LOGOS_USER_DIR` are present, absolute, and distinct per profile in each process environment.
 - A non-module port collision (Qt remote objects, etc.) is a real finding — file upstream against the affected component, do not patch around it inside scaffold.
 - A module that does not honor an externally-provided port override is documented as a known gap pending an upstream fix on that module; capture the module name, the env var that should have worked, and the observed collision.
 - One window crashing while the other survives is recordable evidence; capture the crashing instance's logs from `.scaffold/basecamp/profiles/<name>/` before relaunching.
--- a/docs/basecamp-module-requirements.md
+++ b/docs/basecamp-module-requirements.md
@@ -173,7 +173,7 @@ logos-scaffold basecamp build-portable
 
 `build-portable` does not touch profiles, `basecamp.state`, or the AppImage itself — it only produces artefacts. Load them into your AppImage in the printed order via its "install lgx" button; scaffold is intentionally unaware of the AppImage's install path.
 
-If you launch the AppImage with `--user-dir <path>` (the `bin-macos-app` flag that sets `LOGOS_USER_DIR`), the AppImage stores its state at `<path>` rather than its default XDG data root — orthogonal to scaffold's per-profile XDG isolation under `.scaffold/basecamp/profiles/{alice,bob}/`. The two are independent mechanisms at different layers: scaffold's profiles isolate dev-stack runs; `--user-dir` isolates a single AppImage launch's installed-modules + identity state.
+If you launch a portable basecamp build by hand with `--user-dir <path>` (basecamp 0.2.x; `LOGOS_USER_DIR` is its env equivalent), the app stores its installed-modules + identity state at `<path>` rather than at its default data root. `build-portable` never sets that — it only produces artefacts — but `launch` does: on the macOS portable stack it exports an absolute per-profile `LOGOS_USER_DIR` (and `LOGOS_DATA_DIR`, the 0.1.x name for the same override) pointing at that profile's own module root, because the portable bundle does not honor `XDG_DATA_HOME` on macOS and would otherwise collapse every profile onto the shared `~/Library/Application Support/Logos/LogosBasecamp`. So hand-launching with `--user-dir` and `launch`-ing a profile are the same data-tree redirect reached from two entry points, not independent mechanisms: the flag isolates one ad-hoc launch, while `launch` wires the env equivalent per profile so scaffold's isolation under `.scaffold/basecamp/profiles/{alice,bob}/` actually reaches the app. A `LOGOS_USER_DIR` / `LOGOS_DATA_DIR` you declare yourself in `[basecamp.env]` or `[basecamp.profiles.<name>.env]` is honored rather than overwritten — `launch` only rewrites it to absolute against the project root when it is relative (see "Env exported to the basecamp process" below).
 
 The `.scaffold/basecamp/portable/` directory is wiped and recreated on every `build-portable` run, so re-running after you've removed a module via `basecamp modules` doesn't leave stale symlinks behind.
 
@@ -206,6 +206,25 @@ A project that ships more than one basecamp variant can map the flake attr per h
 aarch64-darwin = "bin-macos-app"
 aarch64-linux  = "bin-appimage"
 ```
+
+### Env exported to the basecamp process
+
+`launch` sets the variables below on the basecamp child, on top of the environment `lgs` itself inherited (nothing is cleared). `<profile-dir>` is `.scaffold/basecamp/profiles/<profile>/`, and the env layering above can override any of them.
+
+| Variable | Value | Set when |
+|---|---|---|
+| `XDG_CONFIG_HOME` | `<profile-dir>/xdg-config` | always |
+| `XDG_DATA_HOME` | `<profile-dir>/xdg-data` | always |
+| `XDG_CACHE_HOME` | `<profile-dir>/xdg-cache` | always |
+| `TMPDIR` | the resolved `runtime_dir`, else `<profile-dir>/xdg-tmp` | always |
+| `XDG_RUNTIME_DIR` | the resolved `runtime_dir` | only when one resolves — a configured `runtime_dir`, or the `/tmp/lgs-<profile>` macOS default |
+| `LOGOS_PROFILE` | the profile name | always |
+| `LOGOS_DATA_DIR` | `<profile-dir>/xdg-data/Logos/LogosBasecamp` | macOS **and** a portable `[repos.basecamp].attr` (`bin-macos-app`, `bin-appimage`, `bin-bundle-dir`) |
+| `LOGOS_USER_DIR` | same as `LOGOS_DATA_DIR` | as above |
+
+Module-owned port-override variables are not in this list: no module has published a name yet, so scaffold exports none.
+
+The last two rows are finalized *after* the env layering, so they are the one place a declared value is post-processed rather than simply taken as-is: an absolute value you declared is kept, a relative one is rewritten to absolute against the project root, and an empty (or whitespace-only) one is treated as unset and replaced by the profile default. Both keys are written because basecamp 0.1.x reads `LOGOS_DATA_DIR` while 0.2.x reads `LOGOS_USER_DIR`, so setting both keeps `launch` agnostic to the pinned generation.
 
 ### The macOS `sun_path == 104` socket-path budget
 
--- a/docs/basecamp-module-requirements.md
+++ b/docs/basecamp-module-requirements.md
@@ -173,7 +173,7 @@ logos-scaffold basecamp build-portable
 
 `build-portable` does not touch profiles, `basecamp.state`, or the AppImage itself — it only produces artefacts. Load them into your AppImage in the printed order via its "install lgx" button; scaffold is intentionally unaware of the AppImage's install path.
 
-If you launch a portable basecamp build by hand with `--user-dir <path>` (basecamp 0.2.x; `LOGOS_USER_DIR` is its env equivalent), the app stores its installed-modules + identity state at `<path>` rather than at its default data root. `build-portable` never sets that — it only produces artefacts — but `launch` does: on the macOS portable stack it exports an absolute per-profile `LOGOS_USER_DIR` (and `LOGOS_DATA_DIR`, the 0.1.x name for the same override) pointing at that profile's own module root, because the portable bundle does not honor `XDG_DATA_HOME` on macOS and would otherwise collapse every profile onto the shared `~/Library/Application Support/Logos/LogosBasecamp`. So hand-launching with `--user-dir` and `launch`-ing a profile are the same data-tree redirect reached from two entry points, not independent mechanisms: the flag isolates one ad-hoc launch, while `launch` wires the env equivalent per profile so scaffold's isolation under `.scaffold/basecamp/profiles/{alice,bob}/` actually reaches the app. A `LOGOS_USER_DIR` / `LOGOS_DATA_DIR` you declare yourself in `[basecamp.env]` or `[basecamp.profiles.<name>.env]` is honored rather than overwritten — `launch` only rewrites it to absolute against the project root when it is relative (see "Env exported to the basecamp process" below).
+If you launch a portable basecamp build by hand with `--user-dir <path>` (basecamp 0.2.x; `LOGOS_USER_DIR` is its env equivalent), the app stores its installed-modules + identity state at `<path>` rather than at its default data root. `build-portable` never sets that — it only produces artefacts — but `launch` does: on the macOS portable stack it exports an absolute per-profile `LOGOS_USER_DIR` (and `LOGOS_DATA_DIR`, the 0.1.x name for the same override) pointing at that profile's own module root, because the portable bundle does not honor `XDG_DATA_HOME` on macOS and would otherwise collapse every profile onto the shared `~/Library/Application Support/Logos/LogosBasecamp`. So hand-launching with `--user-dir` and `launch`-ing a profile are the same data-tree redirect reached from two entry points, not independent mechanisms: the flag isolates one ad-hoc launch, while `launch` wires the env equivalent per profile so scaffold's isolation under `.scaffold/basecamp/profiles/{alice,bob}/` actually reaches the app. A non-empty `LOGOS_USER_DIR` / `LOGOS_DATA_DIR` you declare yourself in `[basecamp.env]` or `[basecamp.profiles.<name>.env]` is honored rather than overwritten — `launch` only rewrites it to absolute against the project root when it is relative. An empty or whitespace-only value is the exception: it counts as unset and is replaced by the profile default (see "Env exported to the basecamp process" below).
 
 The `.scaffold/basecamp/portable/` directory is wiped and recreated on every `build-portable` run, so re-running after you've removed a module via `basecamp modules` doesn't leave stale symlinks behind.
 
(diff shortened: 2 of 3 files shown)


weboko:
## Review — `fix(basecamp): set absolute LOGOS_USER_DIR for Basecamp 0.2.x on macOS portable launch`

Reviewed the diff, ran the full suite locally on this branch, and checked how the change lands against the repo's docs surface.

**Verification performed** (Linux x86_64 container, provisioned per `DOGFOODING.md` "Provisioning the Real LEZ Sequencer Toolchain"):

- `cargo test` — **709 green** (520 lib incl. the 6 new `set_absolute_basecamp_data_dirs_*` tests, 181 CLI, 3 testnode, 5 doc), `cargo fmt --check` clean. Baseline `master` is 707, so the delta is exactly the two added tests with no displacement.
- Nix (flakes enabled) verified working in the environment.
- The runtime path itself is **not reachable on Linux**: the call site is gated `cfg!(target_os = "macos") && is_portable_basecamp(...)`, so `basecamp launch` on this host never invokes the finalizer. The behavioural claims rest entirely on the unit tests plus the `eth-lez-atomic-swaps` aarch64-darwin evidence in the PR body — which is thorough and I have no reason to doubt, but I could not independently reproduce the macOS portable launch here. Flagging that as a limit of my verification, not as a concern about the change.

The refactor itself is clean: `set_absolute_module_root_var` is a faithful per-key extraction of the previous single-key logic (relative→absolutize, empty→unset, absolute→untouched), the two keys are genuinely independent, and the new independence test pins that down. `check`/`fmt` clean, CI green.

Three things I'd like addressed before this lands, all outside the Rust change:

---

### 1. `docs/basecamp-module-requirements.md:176` now contradicts the code — and it ships inside the binary

That file is `include_str!`'d at `src/commands/basecamp.rs:143` and printed verbatim by `lgs basecamp docs`, and `basecamp.rs:149` advertises it as the canonical compatibility contract. I confirmed the stale text ships in the CLI built **from this branch**:

```
$ lgs basecamp docs | grep -n "independent mechanisms"
176:If you launch the AppImage with `--user-dir <path>` ... — orthogonal to scaffold's
per-profile XDG isolation ... The two are independent mechanisms at different layers ...
```

It currently says:

> If you launch the AppImage with `--user-dir <path>` (the `bin-macos-app` flag that sets `LOGOS_USER_DIR`), the AppImage stores its state at `<path>` rather than its default XDG data root — **orthogonal to scaffold's per-profile XDG isolation** under `.scaffold/basecamp/profiles/{alice,bob}/`. **The two are independent mechanisms at different layers**: scaffold's profiles isolate dev-stack runs; `--user-dir` isolates a single AppImage launch's installed-modules + identity state.

After this PR that is no longer true — scaffold sets `LOGOS_USER_DIR` itself, per profile, pointed at the profile's own module root. The "independent mechanisms" framing is exactly the mental model this PR replaces, and a module author reading `lgs basecamp docs` would now be told the opposite of what `launch` does. Worth rewriting that paragraph to say scaffold owns `LOGOS_USER_DIR` on the portable stack and a caller-supplied value is absolutized rather than ignored.

### 2. `DOGFOODING.md:1173` (B3) still names only `LOGOS_DATA_DIR`

```
- On the macOS portable stack, each process environment carries an absolute `LOGOS_DATA_DIR` pointing at its own profile's module root — set automatically by `launch`, no manual export needed.
```

For a 0.2.x pin `LOGOS_USER_DIR` is *the* variable that decides whether profile isolation works at all — `LOGOS_DATA_DIR` is inert there. B3's own "Evidence to Capture" step is `grep -E 'XDG_|LOGOS_'` over `/proc/<pid>/environ`, so a dogfooder following the runbook literally would see a variable the expected-signals list never mentions, and would have no documented signal for the failure this PR fixes (profiles collapsing onto the shared `~/Library/Application Support/Logos/LogosBasecamp`).

The runbook's own maintenance note asks for exactly this ("update this document whenever ... major user-facing behaviors are added, removed, or materially changed"). Suggested: name both keys, and add the collapse signature to B3's Failure Signals. The "Change → Scenario" map at the end of the file also has no entry routing basecamp-launch-env changes to `B2`/`B3` — worth adding while you're there.

### 3. Is the macOS-only gate right for 0.2.x, or should it be all portable stacks?

The rationale given is that Linux is covered because Qt honours `XDG_DATA_HOME` there, so `launch_env`'s existing isolation suffices for both generations. That was demonstrably true for 0.1.x. But the premise of this fix is that 0.2.x moved data-tree resolution into `LogosBasecampPaths.h::baseDirectory()` — so the question is whether *that function* still falls through to `QStandardPaths::AppDataLocation` on Linux, or resolves its own default. If it's the latter, then `bin-appimage` and `bin-bundle-dir` (both in `BASECAMP_PORTABLE_ATTRS`, both Linux) collapse profiles in exactly the same way, and B3 two-instance p2p on Linux silently loses isolation — the same bug, just unreported because the downstream project that surfaced it is on darwin.

Have you read the Linux branch of `baseDirectory()`? If it's not clearly XDG-backed, I'd widen the gate to `is_portable_basecamp(basecamp_repo)` alone and drop the `cfg!(target_os = "macos")`. That's low-risk: on Linux the value the finalizer would write is the same `<profile>/xdg-data/<subpath>` root that `XDG_DATA_HOME` already implies, so setting it explicitly is a no-op where XDG works and a fix where it doesn't. Happy to be told the Linux path is verified — a sentence in the doc comment recording *why* Linux is exempt would then keep the next reader from re-asking this.

---

### Nit (non-blocking)

The key list is an inline literal:

```rust
for key in ["LOGOS_DATA_DIR", "LOGOS_USER_DIR"] {
```

#246 solves the structurally identical "one var got renamed, set both" problem via `constants::WALLET_HOME_ENV_VARS`. Hoisting this to a `BASECAMP_DATA_DIR_ENV_VARS` const in `constants.rs` would keep the two-generation compatibility lists discoverable in one place and match the pattern landing in the sibling PR.

---

Nothing here is a correctness objection to the Rust — the code and tests are good and I'd be happy to see it land once the two doc surfaces are updated and the gating question is answered.


---

weboko (commented):
## Review: `fix(basecamp): set absolute LOGOS_USER_DIR for Basecamp 0.2.x on macOS portable launch`

Reviewed the diff, built it, and ran the suite locally (Linux x86_64, `cargo test --locked` green; the six new `set_absolute_basecamp_data_dirs_*` tests pass). The core change is correct and the reasoning in the description holds up against the code.

### What checks out

- **The generalization is the right shape.** `set_absolute_module_root_var` applies one transform per key with no shared state, and `set_absolute_basecamp_data_dirs` just iterates `["LOGOS_DATA_DIR", "LOGOS_USER_DIR"]`. The `handles_each_key_independently` test is exactly the case that would catch cross-key leakage — good addition.
- **The empty/whitespace guard is preserved per key.** `.filter(|v| !v.to_string_lossy().trim().is_empty())` still runs before the `is_relative()` branch, so `LOGOS_USER_DIR=""` falls back to the profile default rather than absolutizing to the project root. That subtlety was easy to lose in the refactor and it didn't get lost.
- **The macOS gate is defensible.** I confirmed `launch_env` sets `XDG_DATA_HOME` unconditionally on every platform (`src/commands/basecamp.rs:729`), so on Linux the per-profile isolation is already carried by XDG for both basecamp generations. Keeping #238's gate unchanged rather than widening it is the conservative call, and the transform staying platform-independent keeps it unit-testable. No objection.
- **Setting both keys is safe.** Since the two names are read by disjoint basecamp generations, a 0.1.x pin sees no behavior change. Worth it to stay pin-agnostic.

### Docs are out of date — please fold these in

This is the one thing I'd want fixed before merge. The PR touches only `src/commands/basecamp.rs`, but it changes behavior that two docs describe, and both are now wrong:

**1. `DOGFOODING.md:1173`** (B3 "Expected Success Signals") still reads:

> - On the macOS portable stack, each process environment carries an absolute `LOGOS_DATA_DIR` pointing at its own profile's module root — set automatically by `launch`, no manual export needed.

A dogfooder validating B3 against a 0.2.x pin would check `LOGOS_DATA_DIR`, see it set, and pass the scenario — while the key that actually determines whether the profile isolates (`LOGOS_USER_DIR`) goes unverified. That's the exact failure this PR fixes, left unobservable in the scenario that's supposed to catch it. Suggested replacement:

```suggestion
- On the macOS portable stack, each process environment carries an absolute `LOGOS_DATA_DIR` **and** `LOGOS_USER_DIR`, both pointing at its own profile's module root — set automatically by `launch`, no manual export needed. Basecamp 0.1.x reads the former, 0.2.x the latter; on a 0.2.x pin, a missing or relative `LOGOS_USER_DIR` collapses every profile onto the shared `~/Library/Application Support/Logos/LogosBasecamp` and is a real finding.
```

**2. `docs/basecamp-module-requirements.md:176`** is now actively contradicted by the code. It says:

> If you launch the AppImage with `--user-dir <path>` (the `bin-macos-app` flag that sets `LOGOS_USER_DIR`), the AppImage stores its state at `<path>` rather than its default XDG data root — **orthogonal to scaffold's per-profile XDG isolation** … The two are **independent mechanisms at different layers**.

After this PR they are no longer independent or orthogonal — `lgs basecamp launch` sets `LOGOS_USER_DIR` itself on the macOS portable stack, and it deliberately points at the same per-profile module root as the XDG isolation. This file is user-facing via `lgs basecamp docs`, so a reader following it will form the wrong mental model of who owns that variable and may set a conflicting value in `[basecamp.env]`. Please update the paragraph to say scaffold now manages `LOGOS_USER_DIR` on that stack, and that a caller-supplied value in `[basecamp.env]` / `[basecamp.profiles.<name>.env]` is honored but absolutized against the project root.

- The doc comment on `set_absolute_basecamp_data_dirs` is genuinely good — it records *why* absolute matters differently per generation (`@rpath` resolution on 0.1.x vs. verbatim env consumption on 0.2.x). That asymmetry is the kind of thing that gets "simplified" away in a year; having it written down next to the loop is worth the lines.
- `for key in ["LOGOS_DATA_DIR", "LOGOS_USER_DIR"]` hardcodes the list inline. Every other multi-name env set in this repo that I could find lives in `constants.rs` (cf. `WALLET_HOME_ENV_VARS`, introduced in #246). Not worth blocking on, but a `BASECAMP_MODULE_ROOT_ENV_VARS` constant would match the house style and give the next rename one place to edit.
- No security or privacy concern: the values are project-local absolute paths derived from the project root, nothing is logged that wasn't already, and absolutizing strictly narrows where the app can write compared to a relative value.

### Not verified here

The gate is `cfg!(target_os = "macos")`, so the runtime path cannot execute on this Linux host — I exercised the transform through its unit tests only. The downstream `ps eww` / `basecamp.log` evidence in the description is what covers the real launch, and it looks like the right evidence to have captured.

Happy to see this land once the two doc updates are in.

---

weboko:
some good stuff in AI review above that would be good to address

weboko:
## Re-review on the new base — findings still stand, with one correction and refreshed anchors

No new commits since `67f9f00`, so this is not a re-review of changed code. I'm posting because **the base moved underneath this PR after my 2026-07-31 pass**: `#241` (`9fcc376`) landed on `master` on 2026-07-31 21:39 UTC, roughly 18 hours *after* that review, and it rewrote `DOGFOODING.md`, `ADR.md`, `FURPS.md` and `README.md` — the exact doc surfaces my findings anchor to. I re-verified everything against `master@9fcc376`.

### Re-verified on current `master`

Fresh Linux x86_64 container, provisioned per `DOGFOODING.md` (Rust 1.94.1, Nix 2.35.1 with flakes enabled, r0vm 3.0.5 matched to the pinned LEZ `cf3639d8`'s `risc0-zkvm = 3.0.5`, risc0 guest toolchain 1.91.1).

- The branch **still merges cleanly into `master@9fcc376`** — `src/commands/basecamp.rs` is untouched by `#241`, so there is no interaction at all.
- On the merge result: `cargo test` — **716 green** (526 lib incl. the 6 `set_absolute_basecamp_data_dirs_*` tests, 182 CLI, 3 testnode, 5 doc), zero failures.
- As before, the runtime path is unreachable on Linux (`cfg!(target_os = "macos") && is_portable_basecamp(...)`), so this remains unit-test + downstream-evidence verification, not a live macOS reproduction.

### Correction to my own §2 — the Change → Scenario map already routes this

I wrote:

> The "Change → Scenario" map at the end of the file also has no entry routing basecamp-launch-env changes to `B2`/`B3` — worth adding while you're there.

**That was wrong, and I'd rather retract it than have you add a redundant line.** The entry exists and explicitly covers env export:

```
- Changes to `basecamp launch` (kill-and-scrub semantics, XDG isolation, runtime/log/env export,
  port-override env vars, p2p surface): rerun `B3`.
```

I confirmed it is not something `#241` added — it is present at this PR's own merge-base (`:1669` there, `:1698` on current `master`). A change to `set_absolute_basecamp_data_dirs` is squarely "runtime/log/env export", so the map already routes this diff to `B3` correctly. **No map edit needed.** Sorry for the noise.

### Refreshed anchors for the findings that do still stand

Both remaining doc findings reproduce verbatim on `master@9fcc376`; only one line number moved.

**§1 — `docs/basecamp-module-requirements.md:176` (unchanged anchor).** Still reads:

> …orthogonal to scaffold's per-profile XDG isolation… **The two are independent mechanisms at different layers**…

Still `include_str!`'d at `src/commands/basecamp.rs:143` and printed by `lgs basecamp docs`, so the contradicting text still ships inside the binary built from this branch. This is the one I'd most want fixed — it tells a module author the opposite of what `launch` now does.

**§2 — `DOGFOODING.md` B3 expected-signals: `:1173` → now `:1202`** (shifted +29 by `#241`). Still names only `LOGOS_DATA_DIR`:

```
- On the macOS portable stack, each process environment carries an absolute `LOGOS_DATA_DIR`
  pointing at its own profile's module root — set automatically by `launch`, no manual export needed.
```

With the map entry already correct (above), the ask here narrows to just this: name **both** keys in B3's expected signals, and add the profile-collapse signature (`~/Library/Application Support/Logos/LogosBasecamp` shared across profiles) to B3's Failure Signals. B3's own evidence step greps `LOGOS_` out of `/proc/<pid>/environ`, so a dogfooder will see `LOGOS_USER_DIR` and currently has nothing in the runbook to check it against.

**§3 — the Linux gating question is still open and still unanswered.** I want to be straight that I *cannot* settle this one myself: my review access is scoped to `logos-co/scaffold`, so I can't read `LogosBasecampPaths.h::baseDirectory()` in the basecamp repo to see whether its Linux branch falls through to `QStandardPaths::AppDataLocation` (XDG-backed, gate is right) or resolves its own default (gate is too narrow, and `bin-appimage` / `bin-bundle-dir` collapse profiles on Linux the same way). This needs an answer from someone who can read that file — a sentence in the doc comment recording *why* Linux is exempt would close it permanently.

### Nit, re-stated now that its sibling is concrete

My nit about hoisting the inline `["LOGOS_DATA_DIR", "LOGOS_USER_DIR"]` list to a `constants.rs` const has a live precedent to point at: `#246` is carrying exactly that shape as `constants::WALLET_HOME_ENV_VARS`, with a doc comment explaining the two-generation rename. If both land, having one file where "these are the compat env-var lists and here's which pin renamed what" is discoverable is worth the five lines. Still non-blocking.

---

**Net:** the Rust is unchanged and still good — 716 green, clean merge, no `#241` interaction. One of my three asks (§2's map half) was my error and is withdrawn. What remains before I'd approve is small and entirely documentation: the `basecamp docs` paragraph at `docs/basecamp-module-requirements.md:176` that ships stale inside the CLI, B3's expected/failure signals at `DOGFOODING.md:1202`, and a yes/no on the Linux gate.

---

danisharora099:
Pushed `6296b38` + `9e4c0a7` + `7d22a62`.

### §1 — `docs/basecamp-module-requirements.md`

Two things here, and the second turned out bigger than the finding.

**Line 176** rewrote: hand-launching with `--user-dir` and `basecamp launch` aren't independent mechanisms, they're the same data-tree redirect reached from two entry points. It now says what `launch` exports on the macOS portable stack and why, and that a caller-declared value is honoured (and only absolutized when relative — an empty or whitespace-only value counts as unset and gets the profile default).

**The gap underneath it:** grepping `LOGOS_` in that file hit *only* line 176. `LOGOS_DATA_DIR` and `LOGOS_PROFILE` — both of which `launch` actually sets — were never mentioned. So there was no portable-launch env list at all, not a stale one. Added a table under "Per-profile launch configuration" derived from `launch_env()`, `resolve_profile_runtime_dir()` and the launch gate: the XDG trio, `TMPDIR`, `XDG_RUNTIME_DIR` (only when one resolves), `LOGOS_PROFILE`, and the two data-root keys. It also records that nothing is `env_clear`'d and that `scaffold.toml` layering can override any row.

Since the file is `include_str!`-embedded, here's the proof it moved in the binary rather than just on disk:

```
$ cargo build && ./target/debug/logos-scaffold basecamp docs | grep -nE "LOGOS_|Env exported"
210:### Env exported to the basecamp process
221:| `LOGOS_PROFILE` | the profile name | always |
222:| `LOGOS_DATA_DIR` | `<profile-dir>/xdg-data/Logos/LogosBasecamp` | macOS **and** a portable attr |
223:| `LOGOS_USER_DIR` | same default as `LOGOS_DATA_DIR`, resolved independently of it |
```

### §2 — you were right to withdraw it

`DOGFOODING.md:1670` does already route `basecamp launch` env-export changes to B3. I updated B3's signals anyway: the expected-signal line now names both vars with the concrete per-profile path, and there's a new failure signal for the profile-collapse signature (both windows showing only bundled modules while logs report a base data directory under the shared `~/Library/Application Support/Logos/LogosBasecamp`).

### §3 — the gate: kept macOS-only, and here's the honest reasoning

I did **not** widen it, and I want to flag that your "low-risk" framing may understate it slightly. `set_absolute_basecamp_data_dirs` doesn't only fill an unset key — it also rewrites any *relative* caller-supplied `LOGOS_DATA_DIR` / `LOGOS_USER_DIR` from `[basecamp.env]` to absolute. So widening isn't purely additive; it changes observable behaviour for a Linux user who deliberately declares a relative value. Small blast radius, but a behaviour change on a platform none of us can test right now.

The call-site comment now records the scope explicitly: `is_portable_basecamp` matches all three of `bin-macos-app` / `bin-appimage` / `bin-bundle-dir`, the XDG-ignoring behaviour has only been *observed* on `bin-macos-app`, and the Linux stacks are **untested rather than known-good**.

**The question that would settle it** is what the Linux branch of `LogosBasecampPaths.h::baseDirectory()` does. If it falls through to `QStandardPaths::AppDataLocation`, XDG is honoured and the gate is right as written. If it resolves its own default, then `bin-appimage` and `bin-bundle-dir` have the same profile-collapse bug — and since B3 is normally run on Linux, the runbook would be passing a two-instance p2p scenario whose isolation guarantee doesn't actually hold. That's the more alarming branch, and it's why I'd rather get a yes/no than widen speculatively. Do you have that file to hand?

### Four things I found while in there

Not from your findings, but same defect class — shipped docs disagreeing with the code:

- **The socket name has no PID.** Docs said `logos_token_<module>_<pid>`; the code comment at `basecamp.rs:753` says it's a hardcoded `logos_token_<module>`. The whole #89 per-profile-`TMPDIR` rationale depends on that — with a PID in the name, alice and bob couldn't collide in the first place. (Authority for this is our own comment, not upstream liblogos source, which I didn't read. Worth a second pair of eyes if you have it.)
- **Linux `sun_path` is 108 bytes, not "far larger"** than macOS's 104. Four more. A deep project root can overflow there too.
- **`XDG_RUNTIME_DIR`:** the precedence list implied all three cases export it; the third doesn't. `launch` exports none of its own there — though the child still inherits an ambient one, since `launch` clears nothing.
- **The macOS env-capture recipe was broken for its own purpose.** B3 only offered `/proc/<pid>/environ`, which macOS doesn't have. My first attempt at a `ps eww` replacement split on spaces — which truncates `~/Library/Application Support/…` at `Application`, i.e. exactly the value B3 is hunting for. It now says to read the assignments off the line directly. B3 also gains the PID source (`launch.state` records `pid=<pid>`).

`cargo test` 709 (520 lib + 181 CLI + 3 testnode + 5 doc), `cargo fmt --check` clean. Docs-only apart from the gate comment.

weboko:
## Re-review of the three docs commits — the semantic question I had is now answered by your own evidence, and the corrections all land

Re-verified on a fresh Linux x86_64 container provisioned per `DOGFOODING.md` (Rust 1.94.1, Nix 2.35.1 with flakes, r0vm 3.0.5, risc0 guest toolchain 1.91.1, risc0 C++ `v2024.1.5`). Branch merges cleanly into `master@9fcc376`; on the merge result `cargo test` is **716 green** (526 lib incl. the 6 `set_absolute_basecamp_data_dirs_*` tests, 182 CLI, 3 testnode, 5 doc), `cargo fmt --check` clean.

The code is unchanged since `67f9f00`; the three new commits are docs. Taking the substantive one first.

### The question the docs answered for me: is `LOGOS_USER_DIR` the module root or its parent?

This was the one thing in the code I could not settle by reading it. `set_absolute_module_root_var` defaults **both** keys to the same value — `<profile>/xdg-data/<basecamp_xdg_subpath>`, i.e. `.../Logos/LogosBasecamp`. That is unambiguously right for `LOGOS_DATA_DIR` (0.1.x), which is what #238 established. But 0.2.x calls its variable a *user* dir and resolves it in `LogosBasecampPaths.h::baseDirectory()`, and "base directory" is exactly the shape of thing an app appends its own `Logos/LogosBasecamp` suffix to. If 0.2.x did that, this would produce `.../Logos/LogosBasecamp/Logos/LogosBasecamp`, no modules would load, and every unit test here would still pass — they all assert the two keys are *equal*, never that the value is the right depth.

Your downstream log excerpt settles it, and it is the right piece of evidence to have captured:

```
Base data directory: /Users/.../nashville/.scaffold/basecamp/profiles/maker/xdg-data/Logos/LogosBasecamp
... setUserModulesDirectory ... ".../profiles/maker/xdg-data/Logos/LogosBasecamp/modules"
... setUserUiPluginsDirectory ... ".../profiles/maker/xdg-data/Logos/LogosBasecamp/plugins"
```

0.2.1 consumes the value verbatim as the base and appends `modules/` and `plugins/` under it — which is exactly where `basecamp install` puts them. Confirmed by the lgpm enumeration listing `swap`, `swap_ui`, `delivery_module` from those dirs. So the shared default is correct, not a coincidence of the two names happening to be equal.

I'd suggest one durable improvement, not blocking: that depth relationship is currently pinned only by a log excerpt in a PR description. A unit test asserting the default value **ends in** `Logos/LogosBasecamp` and equals the dir `install` writes into would keep a future `basecamp_xdg_subpath` change from silently breaking 0.2.x while all six existing tests stay green.

### The gating comment is the most valuable thing in this diff

> The Linux portable stacks are *untested* here rather than known-good … Note that widening it is not purely additive — it would also start rewriting relative `LOGOS_*_DIR` values declared in `[basecamp.env]` on Linux — so it needs a live check, not a guess.

This is the right way to write that comment. It records the *limit of what was verified* rather than asserting a conclusion, and it names the specific side effect that makes widening the gate a behavior change rather than a safety net. It also correctly resolves the tension I raised earlier between `is_portable_basecamp` matching three attrs and the `cfg!(target_os = "macos")` conjunct narrowing it to one.

### The three corrections

All three check out against the code and against what `launch` actually does:

1. **Socket name** — `logos_token_<module>` without the `_<pid>` suffix, and rooted at `TMPDIR` rather than `XDG_RUNTIME_DIR`. This matters for the budget arithmetic, not just for accuracy: the old text overstated the name length and pointed at a variable that `launch` does not always export.
2. **Linux `sun_path` budget** — 108 vs 104 is right, and "four more than macOS, so a deep project root can overflow there too" is the correct conclusion to draw. The previous "not a practical concern" was the kind of claim that turns into a confusing bug report. The follow-on detail — that when nothing resolves, `TMPDIR` points at `xdg-tmp` and no `XDG_RUNTIME_DIR` is exported *of scaffold's own*, with the child still inheriting the ambient one because `launch` clears nothing — matches the code.
3. **Env recipes** — the `ps eww -p <pid>` guidance for macOS, including *why* not to split on spaces (the profile-collapse target contains a space in `Application Support` and would truncate at `Application`), is a genuinely useful piece of runbook. Pointing at `launch.state`'s `pid=` for the PID makes the step executable rather than aspirational.

The new "Env exported to the basecamp process" table in `docs/basecamp-module-requirements.md` is the right addition — it is the first place the full launch env surface is written down in one view, and the note that the last two rows are post-processed after layering (absolute kept, relative rewritten, empty treated as unset, each key independently) documents the one genuinely surprising behavior.

### The port-override correction

Replacing "a module that does not honor an externally-provided port override" with "scaffold exports no override for it to honor" is a real correctness fix, not a wording change — the old text sent a dogfooder hunting for env vars that do not exist and would have them file a scaffold bug for a module-owned collision. Good catch.

### Scope limit on my verification

To be explicit about what I did **not** do: the runtime path here is `cfg!(target_os = "macos") && is_portable_basecamp(...)`, so on this Linux container it is unreachable by construction. I did not reproduce the launch. This review is unit-test verification plus a read of the transform, the gate, and the docs — the live confirmation is your `eth-lez-atomic-swaps` TR-21 run, which is why the `ps eww` and `basecamp.log` excerpts carry the weight here. The B2/B3 scenarios that would exercise it also need a module project exposing `packages.<system>.lgx`, which this container does not have.

### Verdict

No blocking findings. The one suggestion above (a depth assertion for the shared default) is a follow-up, not a condition. The code was already right; these commits make the docs describe it accurately, and the gating comment is better than what it replaced.

---

danisharora099:
@weboko thanks — and the depth question you raised is a better catch than the findings it came with. You're right that all six `set_absolute_basecamp_data_dirs_*` tests assert the two keys are *equal* and none assert the value is at the right depth, so a `basecamp_xdg_subpath` change could break 0.2.x with the suite fully green. Agreed the log excerpt is thin evidence to leave that resting on.

Happy to add the assertion here if you'd rather it land with the PR — say the word and it's a few lines. Otherwise I'll track it separately, since you flagged it as a follow-up rather than a condition.

Since this one lands with no blocking findings: could you drop a formal approval when you get a moment? Branch protection is the only thing holding it — CI is green and it merges clean into `master@9fcc376`.

weboko (approved):

PR #239, opened 2026-07-21
feat(run): per-profile deploy = false to skip scaffold's deploy step

danisharora099:
## Your Project or User Demand

Found while dogfooding the [`eth-lez-atomic-swaps`](https://github.com/logos-co/eth-lez-atomic-swaps) Phase 2 scaffold migration on macOS `darwin-arm64`. Fixes #237.

## User Story

`lgs run` discovers deployable programs only under the hardcoded `<root>/methods/guest/src/bin` (`discover_deployable_programs`, `deploy.rs:303`), and step 5 always runs it via `compute_program_hashes`. The function **bails** when that directory is absent:

```
missing deployable program directory at <root>/methods/guest/src/bin
```

The atomic-swaps repo keeps its guest program at `programs/lez-htlc/methods/guest/` and deploys it from its own post-deploy command, so scaffold's deploy step is neither wanted nor possible via the hardcoded path — yet `lgs run` bails before ever reaching `post_deploy`, even though the run profile only needs build → localnet → topup → `post_deploy`. There was no way to tell a run profile "the project owns deployment; skip scaffold's deploy step." The workaround was to not use `lgs run` at all.

(Related but distinct from #231 / #59, which concern *binary* auto-discovery paths; this is the *source* dir gate blocking the whole pipeline for self-deploying apps.)

## Change Summary

Adds a `deploy` boolean to `[run]` / `[run.profiles.<name>]`, defaulting to `true`. When `deploy = false`, `run_pipeline_once` skips step 5 entirely — including the program-hash computation that otherwise bails on the missing default program directory — and proceeds straight to the `post_deploy` hooks. Parse/serialize/resolve mirror the existing `reset` boolean; the `true` default is never serialized to keep a fresh `scaffold.toml` minimal.

```toml
[run.profiles.demo]
deploy = false
post_deploy = ["scripts/deploy-and-demo.sh"]
```

## Verification

- `cargo test --lib` — 513 passed, 0 failed. 5 new config tests (default-true, inline `deploy = false`, profile `deploy = false`, round-trip, and "default true is not serialized"); updated the one direct `RunProfile { .. }` literal for the new field.
- `RunProfile` gets a manual `Default` (deploy = true) so a project with no `[run]` config keeps the historical deploy behavior — `derive(Default)` would have given `false` and silently disabled deploy everywhere.
- README: new "Self-deploying projects (`deploy = false`)" subsection under the `run` docs.

## Checklist

- [x] CI is green
[...]

Issue #237, opened 2026-07-21
lgs run hardcodes methods/guest/src/bin; self-deploying apps can't run (add per-profile deploy = false)


danisharora099:
## Summary

`lgs run` hardcodes the deployable-program directory to `<project_root>/methods/guest/src/bin` (`discover_deployable_programs`, `src/commands/deploy.rs:303-310`). The `run` pipeline calls this unconditionally in step 5 via `compute_program_hashes` (`src/commands/run_state.rs:63` → `run.rs:206`), and `discover_deployable_programs` **bails** when that directory is absent:

```
missing deployable program directory at <root>/methods/guest/src/bin
```

So a project whose guest program lives elsewhere, or that deploys its program itself (e.g. from a `post_deploy` hook), cannot use `lgs run` at all — even when its `[run.profiles.*]` only needs build → localnet → topup → post_deploy and never wants scaffold to run the deploy step.

## Environment

Found while dogfooding the `eth-lez-atomic-swaps` Phase 2 scaffold migration on macOS `darwin-arm64`, with Scaffold `7c52211a3f40a6ac5829905d4569712f414776ed` (== current `master` HEAD). The repo keeps its guest program at `programs/lez-htlc/methods/guest/` and deploys it from its own post-deploy command, so scaffold's deploy step is neither wanted nor possible via the hardcoded path.

## Root cause

- `discover_deployable_programs` hardcodes `methods/guest/src/bin` and bails if it doesn't exist (`deploy.rs:304`).
- `run_pipeline_once` step 5 always computes program hashes / runs deploy (`run.rs:198-225`); there is no way to tell a run profile "the project owns deployment, skip scaffold's deploy step."

(Note: `collect_deployed_programs` and `print_deploy_summary` already tolerate a missing dir gracefully; only the deploy/hash path bails. There's also a stale comment at `run_state.rs:59` claiming `discover_deployable_programs` "returns an empty Vec when the bin dir is missing" — it actually bails.)

Related but distinct: #231 and #59 are about *binary* auto-discovery paths; this is about the *source* dir gate blocking `lgs run` entirely for self-deploying apps.

## Proposed fix

[...]

PR #245, opened 2026-07-30
feat(run): per-profile topup = false to skip scaffold's wallet-topup step

### LINKED ISSUE #243: lgs run: mandatory wallet-topup step blocks self-funding projects — add topup = false run-profile toggle
## Problem

`lgs run` runs its pipeline as build → IDL → localnet → **topup** → deploy → post_deploy. Step 4 (`[4/N] Topping up wallet...`) is unconditional: it always calls the wallet-topup path, and when the project has no default wallet address configured it hard-fails the whole run with:

```
wallet topup requires a destination address.
Next step: run `logos-scaffold wallet list` to inspect available wallets, then run `logos-scaffold wallet default set <address>` or pass `--address <address>`.
```

For a project that funds its own accounts, this step is pure overhead at best and a hard blocker at worst.

## Downstream case (eth-lez-atomic-swaps)

The `eth-lez-atomic-swaps` demo binary funds its own LEZ accounts via bounded Piñata faucet claims from inside the app (`src/demo.rs`) — the accounts it uses are created and funded at demo time, not ahead of time. There is no meaningful "default wallet" to top up, yet the `lgs run` pipeline forces a topup (and its confirmation wait / timeout-abort semantics) before the project's own logic ever runs.

This is the exact shape of the problem #237 described for the deploy step, which #239 solved with a per-profile `deploy = false` toggle: a project that owns a pipeline step itself needs a way to tell scaffold to skip that step.

## Proposed fix

Add a `topup` boolean to `RunProfile`, mirroring `deploy` from #239:

- Defaults to `true` (historical behavior for every existing project; manual `Default` impl already exists for exactly this reason).
- Settable inline under `[run]` or per profile under `[run.profiles.<name>]`:

```toml
[run.profiles.demo]
topup = false
post_deploy = ["cargo run --bin demo"]
```

- When `false`, step 4 prints a `Topup skipped (...)` step line (mirroring the deploy-skip message style) instead of calling the wallet-topup path, and the pipeline proceeds to deploy/hooks.
- Serialization only emits the key when it deviates from the default, keeping fresh `scaffold.toml` files minimal (same rule as `deploy`).

## Precedent

- #237 — the equivalent problem statement for the deploy step
- #239 — the merged `deploy = false` per-profile toggle this proposal structurally mirrors

danisharora099:
Closes #243.

## What

Adds a `topup` boolean to run profiles — the structural twin of the `deploy = false` toggle from #239 (which solved the same problem for the deploy step, per #237):

```toml
[run.profiles.demo]
topup = false
post_deploy = ["cargo run --bin demo"]
```

- **Defaults to `true`** — every existing project and every profile that omits the key keeps the historical build → IDL → localnet → topup → deploy pipeline. Enforced by the manual `Default` impl on `RunProfile` (same rationale as `deploy`: `derive(Default)` would give `false` and silently disable the step everywhere).
- **Parsed** inline under `[run]` and per profile under `[run.profiles.<name>]` (`src/config.rs`), mirroring the `deploy` parse sites.
- **Serialized** only when it deviates from the default, keeping fresh `scaffold.toml` files minimal (same emit rule as `deploy`).
- **Pipeline gate** (`src/commands/run.rs`): when `false`, step 4 prints
  ``[4/N] Topup skipped (`topup = false` in the run profile; this project funds its own accounts — e.g. via faucet claims at runtime)``
  instead of `[4/N] Topping up wallet...`, skips the wallet-topup call entirely, and proceeds to deploy/hooks.
- **`SCAFFOLD_TOPUP_SKIPPED`** is exported to every post-deploy hook as `1`/`0`, alongside the existing `SCAFFOLD_DEPLOY_SKIPPED`. A project whose funding lives in a `post_deploy` hook needs to know whether scaffold already funded the wallet — without this it would have to be written "always claim, tolerate already-funded" rather than "claim only if scaffold didn't". The flag is produced by the same expression that executes (or skips) step 4, so what a hook is told cannot drift from what actually happened.
- `resolve_profile` needs no changes — it clones whole profiles, so the new field flows through inline/default-profile/`--profile` selection automatically.

## Why

Self-funding projects have no use for step 4. Downstream, the eth-lez-atomic-swaps demo binary funds its own LEZ accounts via bounded faucet claims at runtime (`src/demo.rs`), so scaffold's topup is duplicated work ahead of the project's own funding logic — and it splits responsibility for funding across two places. `topup = false` keeps it in one. Details in #243.

To be precise about a claim in an earlier revision of this description: step 4 *does* require a destination address, and `resolve_wallet_address` bails with `wallet topup requires a destination address.` when it can't find one. But on the scaffold-default LEZ v0.1.2 pin that path is effectively unreachable — `lgs setup` (which step 1 of `run` chains) seeds `.scaffold/state/wallet.state` from the first preconfigured public account in the wallet config, and v0.1.x debug configs ship those accounts. It becomes reachable on a pin whose config ships none, i.e. LEZ v0.2.0, which is what #246 addresses from the other side. So this PR stands on "my project owns funding", not on "otherwise it breaks". Thanks @weboko for catching that the original framing didn't reproduce.

## Tests

New unit tests in `src/config.rs`, mirroring the #239 deploy tests one-for-one:

- `run_profile_topup_defaults_to_true`
- `parse_config_run_profile_topup_false`
- `parse_config_inline_run_topup_false`
- `run_profile_topup_round_trips_through_parse_serialize`
- `run_profile_topup_true_is_not_serialized`

Plus hook-env tests in `src/commands/run.rs` that drive `build_hook_command` through a real `sh -c` and read the value back, covering both the `1` and `0` cases, and a test pinning the step-4 skip decision to the profile flag.

`cargo test` green (526 lib + 181 CLI + 3 testnode + 5 doc = 715, 0 failures); `cargo fmt --check` clean; `cargo clippy --all-targets` introduces no new warnings vs master.

**Residual risk worth naming:** the *polarity* of the step-4 decision is guarded by one test at the decision site, but the pipeline itself (`run_pipeline_once`) is not covered end-to-end without a real LEZ toolchain, so an inverted branch condition would not be caught by unit tests alone. The D7 scenario below is what actually exercises it.

Docs updated in README (new "Self-funding projects (`topup = false`)" section next to the self-deploying one, plus table rows for `SCAFFOLD_TOPUP_SKIPPED` and `SCAFFOLD_DEPLOY_SKIPPED` — the latter was previously undocumented outside DOGFOODING) and DOGFOODING D7 (self-fund profile scenario asserting the new env var in both states, expected `[4/6] Topup skipped` signal, rerun-trigger line).

### DIFF
--- a/DOGFOODING.md
+++ b/DOGFOODING.md
@@ -601,6 +601,18 @@ post_deploy = ["echo 'deploy skipped:' $SCAFFOLD_DEPLOY_SKIPPED"]
 "$SCAFFOLD_BIN" run --profile self-deploy   # skips step 5, still fires hooks
 ```
 
+Then add a profile that funds its own accounts (`topup = false`) and run it:
+
+```toml
+[run.profiles.self-fund]
+topup = false
+post_deploy = ["echo 'self-fund hook ran'"]
+```
+
+```bash
+"$SCAFFOLD_BIN" run --profile self-fund   # skips step 4, still deploys + fires hooks
+```
+
 ### Expected Success Signals
 
 - The first `run` (no hooks configured) prints a numbered step header for each phase (`[1/5] Building...` through `[5/5] Deploying...`) and ends with a deployed-programs summary.
@@ -611,6 +623,7 @@ post_deploy = ["echo 'deploy skipped:' $SCAFFOLD_DEPLOY_SKIPPED"]
 - `--post-deploy` with `--no-post-deploy` errors at clap parse time with a `cannot be used with` message; exit code is non-zero.
 - A non-zero hook exit aborts the run with a clear `post-deploy hook exited with status N` message.
 - With `deploy = false` in the selected profile, `run --profile self-deploy` prints ``[5/6] Deploy skipped (`deploy = false` in the run profile; ...)`` instead of `[5/6] Deploying...`, skips the program-hash/deploy work entirely, and still runs the `post_deploy` hooks. Each hook sees `SCAFFOLD_DEPLOY_SKIPPED=1` (here `echo 'deploy skipped:' $SCAFFOLD_DEPLOY_SKIPPED` prints `deploy skipped: 1`).
+- With `topup = false` in the selected profile, `run --profile self-fund` prints ``[4/6] Topup skipped (`topup = false` in the run profile; ...)`` instead of `[4/6] Topping up wallet...`, skips the wallet-topup call entirely (no default-wallet-address requirement), and still runs deploy and the `post_deploy` hooks.
 
 ### Failure Signals / Common Pitfalls
 
@@ -625,6 +638,7 @@ post_deploy = ["echo 'deploy skipped:' $SCAFFOLD_DEPLOY_SKIPPED"]
 - Output of `run --post-deploy "echo override"` showing only the override hook fires.
 - Output of `run --no-post-deploy` showing the deployed-programs summary instead of hooks.
 - Output of `run --profile self-deploy` showing the ``[5/6] Deploy skipped (`deploy = false` ...)`` header and the `post_deploy` hook reporting `deploy skipped: 1`.
+- Output of `run --profile self-fund` showing the ``[4/6] Topup skipped (`topup = false` ...)`` header followed by the deploy step and hooks.
 
 ## L1. LEZ Template Bootstrap
 
@@ -1687,7 +1701,7 @@ HEAD0=$("$SCAFFOLD_BIN" test-node blocks head --url "$URL" --json | jq -r .block
 - Changes to wallet flows or wallet-related defaults: rerun `D4`.
 - Changes to diagnostics, report contents, or redaction logic: rerun `D5`.
 - Changes to example runner binaries or template `src/bin/*` code: rerun `D6`.
-- Changes to `run` step ordering, the `deploy = false` deploy-skip branch, post-deploy env vars, post-deploy CLI override flag handling, or `[run]` config parsing: rerun `D7`.
+- Changes to `run` step ordering, the `deploy = false` deploy-skip branch, the `topup = false` topup-skip branch, post-deploy env vars, post-deploy CLI override flag handling, or `[run]` config parsing: rerun `D7`.
 - Changes to LEZ template scaffolding or generated outputs: rerun `L1`, `L2`, `L3`, and `L4`.
 - Changes to CLI argument parsing, help text, or error messages: rerun `E1`.
 - Changes to `create`/`new` flags or template selection logic: rerun `E2`.
--- a/DOGFOODING.md
+++ b/DOGFOODING.md
@@ -606,11 +606,13 @@ Then add a profile that funds its own accounts (`topup = false`) and run it:
 ```toml
 [run.profiles.self-fund]
 topup = false
-post_deploy = ["echo 'self-fund hook ran'"]
+post_deploy = ["echo 'topup skipped:' $SCAFFOLD_TOPUP_SKIPPED"]
 ```
 
 ```bash
 "$SCAFFOLD_BIN" run --profile self-fund   # skips step 4, still deploys + fires hooks
+# same hook without the profile: topup runs, so the variable reports 0
+"$SCAFFOLD_BIN" run --post-deploy 'echo "topup skipped:" $SCAFFOLD_TOPUP_SKIPPED'
 ```
 
 ### Expected Success Signals
@@ -623,7 +625,7 @@ post_deploy = ["echo 'self-fund hook ran'"]
 - `--post-deploy` with `--no-post-deploy` errors at clap parse time with a `cannot be used with` message; exit code is non-zero.
 - A non-zero hook exit aborts the run with a clear `post-deploy hook exited with status N` message.
 - With `deploy = false` in the selected profile, `run --profile self-deploy` prints ``[5/6] Deploy skipped (`deploy = false` in the run profile; ...)`` instead of `[5/6] Deploying...`, skips the program-hash/deploy work entirely, and still runs the `post_deploy` hooks. Each hook sees `SCAFFOLD_DEPLOY_SKIPPED=1` (here `echo 'deploy skipped:' $SCAFFOLD_DEPLOY_SKIPPED` prints `deploy skipped: 1`).
-- With `topup = false` in the selected profile, `run --profile self-fund` prints ``[4/6] Topup skipped (`topup = false` in the run profile; ...)`` instead of `[4/6] Topping up wallet...`, skips the wallet-topup call entirely (no default-wallet-address requirement), and still runs deploy and the `post_deploy` hooks.
+- With `topup = false` in the selected profile, `run --profile self-fund` prints ``[4/6] Topup skipped (`topup = false` in the run profile; ...)`` instead of `[4/6] Topping up wallet...`, skips the wallet-topup call entirely (no destination-address requirement), and still runs deploy and the `post_deploy` hooks. Each hook sees `SCAFFOLD_TOPUP_SKIPPED=1` (here `echo 'topup skipped:' $SCAFFOLD_TOPUP_SKIPPED` prints `topup skipped: 1`). The same hook run without `--profile self-fund` prints `topup skipped: 0` — the variable is always set, never absent.
 
 ### Failure Signals / Common Pitfalls
 
@@ -638,7 +640,7 @@ post_deploy = ["echo 'self-fund hook ran'"]
 - Output of `run --post-deploy "echo override"` showing only the override hook fires.
 - Output of `run --no-post-deploy` showing the deployed-programs summary instead of hooks.
 - Output of `run --profile self-deploy` showing the ``[5/6] Deploy skipped (`deploy = false` ...)`` header and the `post_deploy` hook reporting `deploy skipped: 1`.
-- Output of `run --profile self-fund` showing the ``[4/6] Topup skipped (`topup = false` ...)`` header followed by the deploy step and hooks.
+- Output of `run --profile self-fund` showing the ``[4/6] Topup skipped (`topup = false` ...)`` header followed by the deploy step and the `post_deploy` hook reporting `topup skipped: 1`, plus the profile-less run of the same hook reporting `topup skipped: 0`.
 
 ## L1. LEZ Template Bootstrap
 
--- a/README.md
+++ b/README.md
@@ -311,11 +311,11 @@ post_deploy = ["cargo run --bin demo"]
 
 Hooks see `SCAFFOLD_TOPUP_SKIPPED=1` on such a run, so a funding hook can
 claim only when scaffold did not. The topup step itself needs a destination
-address, but step 1 of every `lgs run` chains `lgs setup`, which seeds one
-into `.scaffold/state/wallet.state` from the first preconfigured public
-account in the wallet config whenever that file is missing. So on a pin that
-ships those accounts the address is in place before step 4 — including after
-the file is deleted, and after `--reset` wipes the wallet.
+address, but on a pin whose wallet config ships preconfigured accounts one is
+already in place by step 4: step 1 of every `lgs run` chains `lgs setup`,
+which seeds `.scaffold/state/wallet.state` from the first preconfigured
+public account whenever that file is missing — and `--reset`, which wipes the
+wallet later, at step 3, re-seeds it before the run continues.
 
 `topup` defaults to `true`; omit it for the normal topup-then-deploy loop.
 
--- a/README.md
+++ b/README.md
@@ -259,9 +259,10 @@ environment variables pre-set:
 | `SCAFFOLD_GUEST_BIN` | Absolute path to the guest `.bin`. Set only when the project has exactly one deployable program |
 
 `SCAFFOLD_TOPUP_SKIPPED` and `SCAFFOLD_DEPLOY_SKIPPED` describe the run as a
-whole, so unlike the program-specific variables they are always set — including
-for projects with no deployable programs. Hooks should branch on their value,
-not on whether they exist.
+whole, so they are set on every hook invocation — including for projects with
+no deployable programs, where `SCAFFOLD_PROGRAM_ID` and `SCAFFOLD_GUEST_BIN`
+are absent. Hooks should branch on their `1`/`0` value, not on whether they
+exist.
 
 `SCAFFOLD_PROGRAM_ID` and `SCAFFOLD_GUEST_BIN` are unset for
 multi-program projects so hooks fail loudly rather than silently
@@ -310,9 +311,11 @@ post_deploy = ["cargo run --bin demo"]
 
 Hooks see `SCAFFOLD_TOPUP_SKIPPED=1` on such a run, so a funding hook can
 claim only when scaffold did not. The topup step itself needs a destination
-address, but `lgs setup` seeds one into `.scaffold/state/wallet.state` from
-the first preconfigured public account in the wallet config, so on a pin
-that ships those accounts it is already there.
+address, but step 1 of every `lgs run` chains `lgs setup`, which seeds one
+into `.scaffold/state/wallet.state` from the first preconfigured public
+account in the wallet config whenever that file is missing. So on a pin that
+ships those accounts the address is in place before step 4 — including after
+the file is deleted, and after `--reset` wipes the wallet.
 
 `topup` defaults to `true`; omit it for the normal topup-then-deploy loop.
 
--- a/README.md
+++ b/README.md
@@ -285,6 +285,25 @@ post_deploy = ["scripts/deploy-and-demo.sh"]
 
 `deploy` defaults to `true`; omit it for the normal deploy loop.
 
+#### Self-funding projects (`topup = false`)
+
+`run` tops up the project's default wallet before deploying (step 4), and
+hard-fails when no default wallet address is configured. A project that
+funds its own accounts — e.g. its demo binary claims from the faucet at
+runtime, or a `post_deploy` hook handles funding — can set `topup = false`
+to skip scaffold's topup step. The pipeline then runs
+build → IDL → localnet → deploy → `post_deploy`, without requiring a
+default wallet address to be set. It works inline under `[run]` or per
+profile, and combines with `deploy = false`:
+
+```toml
+[run.profiles.demo]
+topup = false
+post_deploy = ["cargo run --bin demo"]
+```
+
+`topup` defaults to `true`; omit it for the normal topup-then-deploy loop.
+
 #### One-off override / skip
 
 To run a different hook without editing `scaffold.toml`:
--- a/README.md
+++ b/README.md
@@ -253,8 +253,15 @@ environment variables pre-set:
 | `NSSA_WALLET_HOME_DIR` | Absolute path to project wallet directory |
 | `SCAFFOLD_PROJECT_ROOT` | Absolute path to project root |
 | `SCAFFOLD_IDL_DIR` | Absolute path to IDL output directory |
+| `SCAFFOLD_TOPUP_SKIPPED` | `1` when step 4 was skipped (`topup = false`), `0` when scaffold topped up the wallet. Always set |
+| `SCAFFOLD_DEPLOY_SKIPPED` | `1` when step 5 deployed nothing — either `deploy = false`, or the deploy cache found guest binaries, IDL, config and sequencer unchanged — and `0` when it deployed. Always set |
 | `SCAFFOLD_PROGRAM_ID` | risc0 image ID (hex) of the deployed program. Set only when the project has exactly one deployable program; unset if `spel program-id` cannot extract the ID |
 | `SCAFFOLD_GUEST_BIN` | Absolute path to the guest `.bin`. Set only when the project has exactly one deployable program |
+
+`SCAFFOLD_TOPUP_SKIPPED` and `SCAFFOLD_DEPLOY_SKIPPED` describe the run as a
+whole, so unlike the program-specific variables they are always set — including
+for projects with no deployable programs. Hooks should branch on their value,
+not on whether they exist.
 
 `SCAFFOLD_PROGRAM_ID` and `SCAFFOLD_GUEST_BIN` are unset for
 multi-program projects so hooks fail loudly rather than silently
@@ -287,20 +294,25 @@ post_deploy = ["scripts/deploy-and-demo.sh"]
 
 #### Self-funding projects (`topup = false`)
 
-`run` tops up the project's default wallet before deploying (step 4), and
-hard-fails when no default wallet address is configured. A project that
-funds its own accounts — e.g. its demo binary claims from the faucet at
-runtime, or a `post_deploy` hook handles funding — can set `topup = false`
-to skip scaffold's topup step. The pipeline then runs
-build → IDL → localnet → deploy → `post_deploy`, without requiring a
-default wallet address to be set. It works inline under `[run]` or per
-profile, and combines with `deploy = false`:
+`run` tops up the project's default wallet before deploying (step 4). A
+project that funds its own accounts — e.g. its demo binary claims from the
+faucet at runtime, or a `post_deploy` hook handles funding — can set
+`topup = false` to skip that step and keep funding in one place instead of
+splitting it between scaffold and the project. The pipeline then runs
+build → IDL → localnet → deploy → `post_deploy`. It works inline under
+`[run]` or per profile, and combines with `deploy = false`:
 
 ```toml
 [run.profiles.demo]
 topup = false
 post_deploy = ["cargo run --bin demo"]
 ```
+
+Hooks see `SCAFFOLD_TOPUP_SKIPPED=1` on such a run, so a funding hook can
+claim only when scaffold did not. The topup step itself needs a destination
+address, but `lgs setup` seeds one into `.scaffold/state/wallet.state` from
+the first preconfigured public account in the wallet config, so on a pin
+that ships those accounts it is already there.
 
 `topup` defaults to `true`; omit it for the normal topup-then-deploy loop.
 
--- a/src/commands/run.rs
+++ b/src/commands/run.rs
@@ -185,8 +185,19 @@ fn run_pipeline_once(project: &Project, params: &PipelineParams) -> DynResult<()
     }
 
     // Step 4: Wallet topup. Skipped entirely when the run profile sets
-    // `topup = false` (project funds its own accounts).
-    let topup_skipped = if params.resolved.topup {
+    // `topup = false` (project funds its own accounts). The branch below and
+    // the value hooks read from `SCAFFOLD_TOPUP_SKIPPED` come from the same
+    // expression, so what the hook is told can't drift from what step 4 did.
+    let topup_skipped = topup_step_skipped(&params.resolved);
+    if topup_skipped {
+        // `topup = false`: the project funds its own accounts (e.g. claims
+        // from the faucet at runtime, from a demo binary or a post_deploy
+        // hook). Skip scaffold's topup — including its requirement that a
+        // destination address be resolvable — and proceed to deploy/hooks.
+        println!(
+            "[4/{total_steps}] Topup skipped (`topup = false` in the run profile; this project funds its own accounts — e.g. via faucet claims at runtime)"
+        );
+    } else {
         println!("[4/{total_steps}] Topping up wallet...");
         let outcome = cmd_wallet_topup_inner(project, None, false, false)?;
         if let TopupOutcome::ConfirmationTimeout { message } = outcome {
@@ -196,17 +207,7 @@ fn run_pipeline_once(project: &Project, params: &PipelineParams) -> DynResult<()
                  Hint: retry `logos-scaffold run` or run `logos-scaffold wallet topup` manually."
             );
         }
-        false
-    } else {
-        // `topup = false`: the project funds its own accounts (e.g. claims
-        // from the faucet at runtime, from a demo binary or a post_deploy
-        // hook). Skip scaffold's topup — including its requirement that a
-        // destination address be resolvable — and proceed to deploy/hooks.
-        println!(
-            "[4/{total_steps}] Topup skipped (`topup = false` in the run profile; this project funds its own accounts — e.g. via faucet claims at runtime)"
-        );
-        true
-    };
+    }
 
     // Step 5: Deploy. Skipped entirely when the run profile sets
     // `deploy = false` (project owns deployment); otherwise idempotent —
@@ -275,6 +276,16 @@ fn run_pipeline_once(project: &Project, params: &PipelineParams) -> DynResult<()
     }
 
     Ok(())
+}
+
+/// Whether step 4 skips scaffold's wallet topup for `resolved` — which is
+/// also the value hooks read from `SCAFFOLD_TOPUP_SKIPPED`.
+///
+/// Extracted as its own function so a unit test can pin the polarity
+/// directly: `run_pipeline_once` needs a real sequencer to drive, so an
+/// inverted signal here would otherwise reach every hook unchallenged.
+fn topup_step_skipped(resolved: &RunProfile) -> bool {
+    !resolved.topup
 }
 
 fn reset_for_run(project: &Project, verify_timeout_sec: u64) -> DynResult<()> {
@@ -1394,6 +1405,27 @@ mod tests {
 
         let content = std::fs::read_to_string(&env_file).expect("read env output");
         assert_eq!(content.trim(), "0");
+    }
+
+    /// The two tests above pass `topup_skipped` in explicitly, so they pin
+    /// how `build_hook_command` renders the flag but not how step 4 decides
+    /// it. This one pins the decision: swapping the polarity would hand every
+    /// hook the exact inverse of what the pipeline did, and no test that
+    /// stops at `build_hook_command` can notice.
+    #[test]
+    fn topup_step_skipped_follows_the_profile_flag() {
+        // Default profile (`topup = true`) tops up, so nothing was skipped.
+        assert!(
+            !topup_step_skipped(&RunProfile::default()),
+            "`topup = true` must report SCAFFOLD_TOPUP_SKIPPED=0"
+        );
+        assert!(
+            topup_step_skipped(&RunProfile {
+                topup: false,
+                ..RunProfile::default()
+            }),
+            "`topup = false` must report SCAFFOLD_TOPUP_SKIPPED=1"
+        );
     }
 
     #[test]
--- a/src/commands/run.rs
+++ b/src/commands/run.rs
@@ -184,14 +184,25 @@ fn run_pipeline_once(project: &Project, params: &PipelineParams) -> DynResult<()
         ensure_localnet(project, params.localnet_timeout_sec)?;
     }
 
-    // Step 4: Wallet topup
-    println!("[4/{total_steps}] Topping up wallet...");
-    let outcome = cmd_wallet_topup_inner(project, None, false, false)?;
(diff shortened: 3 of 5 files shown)


weboko:
## Review — `feat(run): per-profile topup = false to skip scaffold's wallet-topup step`

This is a clean, faithful twin of #239, and it's the best-documented of the three PRs currently open — README, D7, and the "Change → Scenario" map are all updated in the same commit, which is exactly right. I verified the feature end-to-end on real infrastructure rather than reading it.

### Verification performed

Full toolchain provisioned per `DOGFOODING.md` on a fresh Linux x86_64 container: r0vm 3.0.5 + the risc0 Rust guest toolchain (both via the curl fallback — see the note at the bottom), circuits v0.4.1, real `sequencer_service` and `wallet` built from the pinned LEZ v0.1.2, `setup` complete, default wallet seeded. Then **D7 executed against a live localnet with real deploys**, using this branch's binary.

`cargo test` — **714 green** (523 lib incl. the 5 new `run_profile_topup_*` tests, 181 CLI, 3 testnode, 5 doc); `cargo fmt --check` clean. Baseline `master` is 707.

**The new branch works exactly as specified:**

```
$ lgs run --profile self-fund
[1/6] Building...
[2/6] Building IDL...
[3/6] Ensuring localnet...
[4/6] Topup skipped (`topup = false` in the run profile; this project funds its own accounts — e.g. via faucet claims at runtime)
[5/6] Deploying programs...
[6/6] Running 2 post-deploy hook(s)...
EXIT=0
```

No wallet subprocess is spawned at step 4 (contrast the default profile, which shells out to `wallet account get` + `wallet pinata claim`), deploy still runs, all 5 example programs submit, and hooks fire. The pre-existing `deploy = false` branch is unaffected — `run --profile self-deploy` still prints `[5/6] Deploy skipped (...)` and its hook still sees `SCAFFOLD_DEPLOY_SKIPPED=1`.

I also confirmed the claim that `resolve_profile` needs no changes: it `.cloned()`s whole `RunProfile` values in all three selection paths (explicit `--profile`, `default_profile`, inline), so `topup` propagates for free. And the serialization rule holds — `topup = true` never appears in a written `scaffold.toml`, `topup = false` round-trips.

Two observations and a nit.

---

### 1. Hooks can't tell whether scaffold topped up — no `SCAFFOLD_TOPUP_SKIPPED`

The deploy toggle exposes its state to hooks:

```rust
.env("SCAFFOLD_DEPLOY_SKIPPED", if run_deploy_skipped { "1" } else { "0" })
```

The topup toggle exposes nothing — there's no `SCAFFOLD_TOPUP_SKIPPED` anywhere in the branch. That's an asymmetry worth closing, because it bites precisely the use case this PR is *for*. Your own README text says the toggle is for a project where "its demo binary claims from the faucet at runtime, **or a `post_deploy` hook handles funding**". A hook in that second category is the one piece of code that genuinely needs to know whether scaffold already funded the wallet — and after this PR it's the only pipeline decision a hook can't observe.

Concretely, a hook that wants to be safe under both profiles has to be written as "always claim, tolerate already-funded" rather than "claim only if scaffold didn't", because it has no signal. Adding the var alongside the existing one is a couple of lines and keeps the two toggles teaching the same lesson.

If you add it, D7's self-fund scenario should assert it the way the self-deploy scenario asserts `SCAFFOLD_DEPLOY_SKIPPED` — right now the new scenario's hook is `echo 'self-fund hook ran'`, which passes whether or not the env contract is correct. That's the one place the added D7 coverage is weaker than the block it mirrors.

### 2. The motivating hard-fail isn't reachable on the default pin — worth stating where it *is*

The PR (and #243) frame step 4 as hard-failing "when no default wallet address is configured":

> ```
> wallet topup requires a destination address.
> ```

I tried to reproduce that as the negative half of the test and **could not**, on the scaffold-default LEZ v0.1.2 pin. Deleting `.scaffold/state/wallet.state` entirely and running bare `lgs run`:

```
[4/6] Topping up wallet...
wallet topup complete
EXIT=0
```

`resolve_wallet_address` falls back to `first_public_wallet_address`, which reads the preconfigured account out of the wallet config — and v0.1.x debug configs ship `initial_accounts`, so there's always an address to resolve. The hard-fail only becomes reachable on a pin whose config ships **no** preconfigured accounts, i.e. LEZ v0.2.0 — which is exactly the gap #246 is fixing from the other direction.

This isn't a defect in this PR: the feature does what it says, and "skip a step my project doesn't want" stands on its own merits regardless of whether the step would have failed. But it means **#245 and #246 are two answers to the same downstream pain**, and a reader of either issue would benefit from that being said out loud. It also means the `topup = false` escape hatch is most valuable *before* #246 lands, and mostly a convenience after. Worth a line in the README section noting that scaffold can resolve an address from the wallet config when `wallet.state` is absent, so `topup = false` is about "my project owns funding", not "otherwise it breaks".

### 3. Nit — the skip branch has no CLI-level test

The 5 new tests are all `config.rs` parse/serialize; nothing exercises `run_pipeline_once`'s new `if params.resolved.topup` branch. I checked, and #239's `deploy = false` branch has no CLI test either, so this is consistent with precedent rather than a regression — flagging only because it means both skip branches are covered exclusively by manual D7, and D7's new self-fund step (per §1) doesn't assert much. Not blocking.

---

### Unrelated, but you'll hit it: two pre-existing bugs surfaced during this D7 run

Neither is caused by this PR — both reproduced on `master` — and both are already in flight in #241. Noting them so the D7 evidence above is read correctly:

- **`program_id: unavailable`** in every deploy summary and `$SCAFFOLD_PROGRAM_ID` unset in hooks, despite spel being built. That's D7's own documented Failure Signal, and it's #241's spel `inspect` → `program-id` rename.
- **The LEZ cache is wiped and re-cloned on every repo sync** in any container with a git `url.<base>.insteadOf` rewrite (standard for proxied CI sandboxes, including this one), discarding the just-built `sequencer_service` and `wallet` under `<lez>/target`. I confirmed the exact mechanism #241 describes — `git remote get-url origin` returns the *rewritten* URL while `git config --get remote.origin.url` returns the configured one:

  ```
  remote get-url: http://local_proxy@127.0.0.1:41729/git/logos-blockchain/logos-execution-zone.git
  configured:     https://github.com/logos-blockchain/logos-execution-zone.git
  ```

  I had to drop the rewrite from the environment before D7 could complete at all — otherwise each of the ~8 `run` invocations re-cloned and rebuilt from scratch. **Practical consequence for merge order: #241 should land before this, or nobody in a proxied environment can run the D7 scenario this PR adds.**

Also worth folding back into `DOGFOODING.md` independently of this PR: the "Provisioning the Real LEZ Sequencer Toolchain" section installs `r0vm` only, but `lgs build` also needs the risc0 **Rust guest toolchain**, without which step 1 dies with `Risc Zero Rust toolchain not found. Try running 'rzup install rust'`. `rzup` can't fetch it behind the TLS-intercepting proxy (its rustls build uses compiled-in roots and ignores `SSL_CERT_FILE`, so even `no_proxy` doesn't help), so it needs the same curl fallback the runbook already documents for r0vm — download `rust-toolchain-<triple>.tar.gz` from `risc0/rust`, extract to `~/.risc0/toolchains/v<ver>-rust-<triple>/`, and write `~/.risc0/settings.toml`. Happy to open that as a separate docs PR if useful.

---

Feature-wise this is good work and I'd be comfortable with it landing once §1 is resolved (or explicitly declined with a reason).


---

weboko (commented):
## Review: `feat(run): per-profile topup = false to skip scaffold's wallet-topup step`

Built and tested locally (Linux x86_64): `cargo fmt --check` clean and the full suite green — 523 lib + 181 CLI + 3 API + 5 doc, exactly the counts your description claims, which I appreciate being able to confirm rather than take on faith. The change is small, symmetric with #239, and I could not find a correctness defect in it. Two things I'd like addressed — one concrete gap, one architectural question — plus verification notes.

### What checks out

- **Default handling is right, and for the right reason.** The manual `Default for RunProfile` impl with `topup: true` (rather than `derive(Default)`) is the load-bearing detail here — `false` would silently disable topup for every existing project. The comment says exactly that. `run_profile_topup_defaults_to_true` covers the profile, the inline profile, *and* `RunProfile::default()` in one test, which is the right coverage.
- **Serialize-only-when-deviating is consistent.** `has_inline` gained the `|| !run.inline.topup` term, so a project that sets only `topup = false` inline still emits a `[run]` table instead of losing the setting on round-trip. `run_profile_topup_round_trips_through_parse_serialize` pins that, and `run_profile_topup_true_is_not_serialized` pins the minimal-file rule. Both parse sites (inline + per-profile) mirror `deploy` exactly.
- **Step numbering is preserved.** `total_steps` is untouched, and the skip branch keeps step 4's slot (`[4/N] Topup skipped …`) rather than renumbering — same as the `deploy = false` branch keeps step 5. That's the behavior the DOGFOODING D7 signal now asserts.
- **The README's central claim is accurate.** I verified that "without requiring a default wallet address to be set" is true and not just plausible: the `wallet topup requires a destination address` bail lives in `resolve_wallet_address` (`src/commands/wallet_support.rs:189`), which is reached only from `cmd_wallet_topup_inner`. `deploy.rs` never calls `resolve_wallet_address` or `read_default_wallet_address`, so skipping step 4 genuinely removes the hard requirement rather than deferring it to step 5.
- **`resolve_profile` needing no change is correct**, not an oversight — it clones whole `RunProfile` values, so the new field flows through inline/default/`--profile` selection for free.

### 1. The `deploy = false` twin sets a hook env var; `topup = false` sets none

This is the gap I'd want closed before merge. When `deploy = false` skips step 5, hooks are told about it:

- `SCAFFOLD_DEPLOY_SKIPPED` (run-level, set unconditionally — `src/commands/run.rs:782`)
- `SCAFFOLD_DEPLOY_SKIPPED_<name>` per program (`src/commands/run.rs:800`)

`topup = false` adds no equivalent, so a `post_deploy` hook has no way to know whether scaffold funded the wallet or left it to the project. I confirmed this on a live run of D7 against a real localnet — note the last line of each block:

```
############ default profile (topup runs) ############
[4/6] Topping up wallet...
[5/6] Deploying programs...   Succeeded: 5  Failed: 0
===> post_deploy[3/4]: echo 'deploy skipped:' $SCAFFOLD_DEPLOY_SKIPPED
deploy skipped: 0
===> post_deploy[4/4]: echo 'topup skipped:' ${SCAFFOLD_TOPUP_SKIPPED:-<UNSET>}
topup skipped: <UNSET>

############ self-fund profile (topup = false) ############
[4/6] Topup skipped (`topup = false` in the run profile; this project funds its own accounts — e.g. via faucet claims at runtime)
[5/6] Deploying programs...   Succeeded: 5  Failed: 0
===> post_deploy[3/3]: echo 'topup skipped:' ${SCAFFOLD_TOPUP_SKIPPED:-<UNSET>}
topup skipped: <UNSET>
```

The hook sees an identical, unset value whether topup ran or was skipped — the two cases are indistinguishable from inside the hook. That matters precisely because of this PR's own motivating case: the downstream demo binary that "funds its own LEZ accounts via bounded faucet claims at runtime" runs *as a post-deploy hook*. Today that hook must either claim unconditionally (wasting faucet budget and doing redundant work when `topup = true`) or hardcode the assumption that it's always the funder — which silently breaks the moment someone flips the profile back.

Suggest adding the run-level twin alongside the existing one in `build_hook_command`:

```rust
.env(
    "SCAFFOLD_TOPUP_SKIPPED",
    if run_topup_skipped { "1" } else { "0" },
)
```

If you add it, three docs need the same one-line follow-through, since the hook env is treated as a public surface:
- `README.md:157` — the prose list of hook env vars.
- `DOGFOODING.md:605` — the "missing any of …" env-contract failure signal.
- `ADR.md:349-352` — the "Hook Env Contract is a Documented Public Surface" bullet list.

If you'd rather *not* add it, that's a defensible call too — but then please say so explicitly in the PR description, because "the structural twin of the `deploy = false` toggle" sets the expectation that the hook-visible half came along with it.

### 2. Architectural: this is the second skip-knob, and the ADR still says there are none

`ADR.md:342-343` states the pipeline principle as:

> Step ordering is fixed; if a new step is needed, it joins the chain rather than offering a knob.

`deploy = false` (#239) was the first knob; `topup = false` is the second. Neither updated that ADR section, so the recorded architecture now says the opposite of what the code does, and there's no written boundary for which of the six steps may become skippable next. I don't think that blocks this PR — the precedent is already set and this change is a faithful application of it — but someone should reconcile it, ideally here since this is the change that turns a one-off into a pattern. A sentence would do: step *ordering* is fixed, and individual steps may expose a boolean skip when the project legitimately owns that concern itself, with the step header still printed so the pipeline shape stays visible.

- The skip message is long enough to wrap in most terminals. The parenthetical explains both the cause and a rationale (`this project funds its own accounts — e.g. via faucet claims at runtime`); the `deploy` twin is similarly long, so this is consistent — just noting that both are now near the edge.
- `topup = false` combined with the default `deploy = true` on an unfunded wallet will fail at step 5 rather than step 4. That's the user's explicit choice and the error surfaces from deploy normally, so no change requested — but it may be worth one clause in the README section noting that the project is then responsible for funding *before* deploy if its deploy path needs funds.
- No security or privacy concern. The toggle only removes a faucet-claim subprocess; it grants nothing, widens no path, and cannot cause funds to move anywhere they wouldn't already.

### Verification notes

I ran D7 end to end against a real localnet rather than relying on the unit tests. Environment provisioned per DOGFOODING's "Provisioning the Real LEZ Sequencer Toolchain": Nix 2.35.1, r0vm 3.0.5 pinned to the LEZ tree's `risc0-zkvm` version, real `sequencer_service` built from the pinned checkout, `lgs setup` completed (wallet + spel built, default wallet seeded), localnet ready on `:3040`.

Results, all matching the DOGFOODING signals this PR added:

- **Bare `lgs run`** — `[1/5]`…`[5/5]`, 5 guest programs built and deployed, `Succeeded: 5  Failed: 0`.
- **Default profile with hooks** — `[4/6] Topping up wallet...` ran the real faucet claim, deploy succeeded, all 4 hooks fired with correct `cwd` and env.
- **`--profile self-fund` (`topup = false`)** — printed exactly the header your DOGFOODING entry specifies:
  ```
  [4/6] Topup skipped (`topup = false` in the run profile; this project funds its own accounts — e.g. via faucet claims at runtime)
  ```
  Step 4's slot is preserved, the topup call is skipped entirely, deploy still ran (`Succeeded: 5  Failed: 0`), and the `post_deploy` hooks still fired. The localnet was reused rather than restarted.

- **`topup = false` + `deploy = false` together** — I checked the README's "combines with `deploy = false`" claim rather than take it on trust, and it holds:
  ```
  [4/6] Topup skipped (`topup = false` in the run profile; ...)
  [5/6] Deploy skipped (`deploy = false` in the run profile; ...)
  [6/6] Running 1 post-deploy hook(s)...
  both-off hook ran; deploy_skipped= 1
  ```
  Both slots keep their numbers and the hooks still fire. Note the asymmetry in the last line: the hook can read `deploy_skipped=1`, but there is no corresponding way for it to learn that topup was skipped in the very same run.

No problems observed in the pipeline itself — the behavior is exactly what the PR and the DOGFOODING entry describe. The only gap is the missing hook-visible signal in §1 above.

---

weboko:
I'd like this `The motivating hard-fail isn't reachable on the default pin — worth stating where it is` to be addressed

weboko:
## Re-review on the new base — the merge-order precondition I named has cleared, and D7 now runs unmodified

No new commits since `fd90c34`, so this is not a re-review of changed code. I'm posting because **the thing I said had to happen first has happened**: `#241` (`9fcc376`) landed on `master` on 2026-07-31 21:39 UTC, about 17 hours *after* my review. I closed that review with:

> **Practical consequence for merge order: #241 should land before this, or nobody in a proxied environment can run the D7 scenario this PR adds.**

That is now satisfied, so I re-ran the whole thing against `master@9fcc376` to replace the caveated evidence from last time with clean evidence.

### Re-verified on current `master`

Fresh Linux x86_64 container provisioned per `DOGFOODING.md` (Rust 1.94.1, Nix 2.35.1 with flakes, r0vm 3.0.5 matched to the pinned LEZ `cf3639d8`'s `risc0-zkvm = 3.0.5`, risc0 guest toolchain 1.91.1, circuits v0.4.1, real `sequencer_service` + `wallet` built from the pin).

- Branch **merges cleanly into `master@9fcc376`**. `#241` rewrote `src/commands/run.rs`, which this PR also touches, so I checked the result rather than trusting the auto-merge: no semantic drift, the `topup` gate lands exactly where it did before.
- `cargo test` on the merge result — **719 green, 0 failed**.
- **Full D7 re-executed end to end** against a real localnet with real deploys, using the merged binary.

### D7: every step passes, and the two caveats on my last run are gone

`setup` → seeded wallet → bare `run` → hooks → overrides → both profiles. Bare run:

```
[1/5] Building...  [2/5] Building IDL...  [3/5] Ensuring localnet...
localnet ready (sequencer pid=377)
[4/5] Topping up wallet...   wallet topup complete
[5/5] Deploying programs...  Succeeded: 5  Failed: 0
```

**Caveat 1 retired — `program_id` is populated now.** Last time every deploy summary read `program_id: unavailable` (the spel `inspect` → `program-id` rename). On the new base the summary carries real IDs:

```
program_id: dba9478d28982969512dd78575a651df2d7499e27234991b279d776872f8c9c2
program_id: fd1a70223063c97e539289b1e3a0b9c0fad57fbe36e38560a12a1f845798d42a
```

`#241`'s deploy pacing is also visible and working (`Waiting for a new block past 7 before the next deployment...`). D7's documented Failure Signal for this no longer fires.

*(In the 7-hook block, `$SCAFFOLD_PROGRAM_ID` still prints `unavailable` — that is correct, not a regression: the default template deploys 5 programs and the shortcut is documented as single-program-only.)*

**Caveat 2 — I have to be precise, and it is narrower than I implied.** I could not reproduce the cache-wipe here: this container has no `insteadOf` rewrite affecting the LEZ origin (`git remote get-url` and `git config --get remote.origin.url` both return `https://github.com/logos-blockchain/logos-execution-zone.git`), so the trigger condition simply isn't present. What I can state: the LEZ build was reused across ~8 `run` invocations with no re-clone (both binaries kept their original build mtimes), the whole scenario ran with no environment surgery on my part, and `#241`'s regression test `repo::tests::insteadof_rewrite_does_not_trigger_source_mismatch` passes on the merge result. So the fix is in and tested — I just didn't independently re-trigger the original failure.

**The feature itself is correct.** `topup = false` skips step 4 and still deploys:

```
[4/6] Topup skipped (`topup = false` in the run profile; this project funds its own accounts — e.g. via faucet claims at runtime)
[5/6] Deploying programs...   OK hello_world submitted ... (5/5 submitted)
[6/6] Running 3 post-deploy hook(s)...
```

Objective check that the step is genuinely skipped rather than merely relabelled — counting wallet subprocesses in the same run:

| profile | `wallet account get` / `pinata claim` invocations |
| --- | --- |
| `self-fund` (`topup = false`) | **0** |
| default | **2** |

`deploy = false` is unaffected (`[5/6] Deploy skipped (...)`, hook still sees `SCAFFOLD_DEPLOY_SKIPPED=1`), and the two toggles compose.

---

### §1 is still open, and I now have runtime evidence for it rather than an argument

My one substantive ask was the missing `SCAFFOLD_TOPUP_SKIPPED`. I put both variables in a single `self-fund` hook block so the asymmetry shows up side by side in one run:

```
===> post_deploy[2/3]: echo 'topup skipped var:' ${SCAFFOLD_TOPUP_SKIPPED:-UNSET}
topup skipped var: UNSET
===> post_deploy[3/3]: echo 'deploy skipped var:' ${SCAFFOLD_DEPLOY_SKIPPED:-UNSET}
deploy skipped var: 0
```

In the profile whose *entire purpose* is "my project owns funding", the hook that owns funding is told nothing about topup — while being told, in the same breath, that deploy ran. `SCAFFOLD_DEPLOY_SKIPPED` is even emitted as `0` on the not-skipped path, so the contract already commits to reporting both states of a toggle; `topup` is the one that reports neither. A hook can't ask "did scaffold already fund this?" and must be written defensively instead. Still a couple of lines, and it keeps the two toggles teaching the same lesson.

### §2 — still the one I'd actually like addressed

Restating because it's the item I flagged on 2026-07-31 and it's unchanged: the README section should note that `run` can resolve a topup destination from the wallet config when `wallet.state` is absent, so `topup = false` reads as *"my project owns funding"* rather than *"otherwise it breaks"*. My finding stands exactly as written — on the v0.1.2 default pin the hard-fail this PR is framed against is **not reachable**, because `resolve_wallet_address` falls back to `first_public_wallet_address` and v0.1.x debug configs ship `initial_accounts`. I re-confirmed the mechanism on this run: `setup` printed `default wallet seeded from preconfigured account`, and `run --profile self-fund` printed `default wallet already configured: Public/CbgR6tj5kWx5oziiFptM7jMvrQeYY3Mzaao6ciuhSr2r`. The hard-fail only becomes reachable on a pin shipping no preconfigured accounts — i.e. v0.2.0, which is what `#246` addresses from the other side.

§3 (no CLI-level test for the skip branch) remains non-blocking and consistent with `#239`'s precedent.

---

**Net:** the feature is correct, well documented (README + D7 + the Change → Scenario map, all in-commit — still the best-documented of the open PRs), 719 green, clean merge, and D7 now runs unmodified on the new base. The blocker I raised was environmental and has cleared. What's left is §1 (~2 lines + a D7 assertion, and I'd now call it the more valuable of the two given the evidence above) and §2 (one README sentence). Neither is a correctness objection.

---

danisharora099:
Thanks — and thanks for the merge-order call on #241, which has since landed. Pushed `e0bfe8b` + `e91fe05` + `d1df445`.

### §1 — `SCAFFOLD_TOPUP_SKIPPED`

Added. You were right that this bites exactly the case the feature is for.

I threaded it as an explicit parameter rather than hanging it off `DeployedPrograms` the way `deploy_skipped` is carried — topup-skip isn't a property of the deployed programs, and unlike deploy it has no per-program dimension, so there's no `_<name>` indexed form. The flag is produced by the same expression that executes or skips step 4, so what a hook is told can't drift from what ran.

Two tests drive `build_hook_command` through a real `sh -c` and read the value back, covering `1` and `0`, plus one pinning the step-4 decision to the profile flag.

While in there: `SCAFFOLD_DEPLOY_SKIPPED` turned out to be documented **nowhere** outside DOGFOODING — zero hits in README, `docs/`, `cli.rs`, `cli_help.rs`. Both are now in the README env table.

### §2 — the hard-fail that doesn't reproduce

You were right, and I've rewritten both the README section and the PR description. Your repro was correct; the mechanism is slightly different from the one in your note. There's no runtime fallback in `resolve_wallet_address` — `cmd_wallet_topup_inner` only ever reads `wallet.state` (`wallet.rs:242`). The address gets there earlier, at `lgs setup` time, via `ensure_default_wallet_seeded` → `first_public_wallet_address`, which reads `initial_accounts` out of the wallet config. Same practical consequence you observed: unreachable on any pin that ships those accounts.

So the section now stands on "my project owns funding — keep it in one place" rather than "otherwise it breaks", and notes where the hard-fail *is* reachable (a pin shipping no preconfigured accounts, i.e. v0.2.0 — #246 from the other side), which is the connection you asked to be said out loud.

### §3 — the D7 nit

The self-fund hook now asserts `$SCAFFOLD_TOPUP_SKIPPED` the way the self-deploy block asserts its sibling, and the scenario runs a second profile-less invocation so the `0` case is actually executed rather than asserted in prose.

**One residual I'd rather name than paper over.** The two hook-env tests pass the bool in explicitly, so they'd survive an inverted step-4 branch condition. There's now a test at the decision site, but `run_pipeline_once` itself isn't covered end-to-end without a real LEZ toolchain — so D7 is what genuinely exercises the polarity, not the unit suite. I didn't want "mutation-checked" to imply more coverage than exists.

### Not done

`SCAFFOLD_DEPLOY_SKIPPED_<name>` and the rest of the indexed family (`SCAFFOLD_PROGRAMS`, `SCAFFOLD_PROGRAM_ID_<name>`, `SCAFFOLD_GUEST_BIN_<name>`) are still undocumented in README. Adding one member would imply the set was covered; that's a separate docs pass.

`cargo test` 715 (526 lib + 181 CLI + 3 testnode + 5 doc), `cargo fmt --check` clean, no new clippy warnings. `tests/cli.rs` is flaky on macOS via localnet/test-node port contention — different failures per run, all green in isolation, none touching `run.rs`.

Your DOGFOODING note about the risc0 Rust guest toolchain needing the same curl fallback as r0vm would be a genuinely useful separate PR if you still want to open it.

danisharora099:
Rebased onto `master@9fcc376` — the doc additions above collided with #241, which rewrote the same regions. Both conflicts were content, not just line drift, so noting the resolutions:

- **README env table** (#245): #241 renamed `spel inspect` → `spel program-id` in the `SCAFFOLD_PROGRAM_ID` row. Kept master's wording; the two new rows sit above it.
- **DOGFOODING D6/L4** (#246): #241 had improved both blocks — D6 now uses two distinct accounts (`<account-id-a>` / `<account-id-b>`) with the `Public/` prefix, and L4 gained `export HOST_CC=cc HOST_CXX=c++`. Kept all of that; only the `export` line picked up the second var name.

---

**Unrelated, but you'll hit it:** #241's deploy-pacing tests are flaky on macOS, and it isn't caused by either branch — I reproduced it on a clean detached checkout of `origin/master@9fcc376`: run 1 failed, runs 2 and 3 passed, same tree.

Affected: `commands::deploy::tests::wait_for_block_past_returns_true_once_head_advances`, `wait_for_block_past_times_out_when_head_stalls`, and `deploy_aborts_remaining_programs_when_head_stalls` in `tests/cli.rs`. A different member of that set fails on each run, and each passes in isolation (3/3). Two failure shapes:

```
panicked at src/commands/deploy.rs:900:9:
head reached 6 > 5, pacing wait must succeed

panicked at ureq-2.12.1/src/response.rs:403:34:
failed to read exact buffer length from stream: Os { code: 22, kind: InvalidInput }
```

The `ureq` one is the same signature you saw against the local mock. Looks like the mock sequencer these tests stand up contends under parallel test execution. Linux CI is green, so it may be macOS-only — but it makes `cargo test` unreliable locally on a Mac, which is worth its own issue against `master` rather than being carried by either of these PRs.

danisharora099:
@weboko this is ready for another pass when you have a window — all three items from your review are addressed in the commits above, and it's been rebased onto `master@9fcc376` (the README env table collided with #241's `spel inspect` → `spel program-id` rename; resolution noted in the comment above).

- **§1** — `SCAFFOLD_TOPUP_SKIPPED` added, with tests covering both values. Also documented `SCAFFOLD_DEPLOY_SKIPPED`, which turned out to be undocumented outside DOGFOODING entirely.
- **§2** — you were right that it doesn't reproduce; both the README section and the PR description now stand on "my project owns funding" rather than the hard-fail. The mechanism is slightly different from your note — there's no runtime fallback in `resolve_wallet_address`, the address is seeded at `setup` time.
- **§3** — D7's self-fund hook now asserts the var, and the scenario exercises the `0` case too.

One residual I called out rather than papered over: the hook-env tests pass the bool in explicitly, so they'd survive an inverted step-4 branch condition. There's a test at the decision site now, but `run_pipeline_once` isn't covered end-to-end without a real toolchain — D7 is what genuinely exercises the polarity.

CI green, mergeable.

weboko:
## Re-review of the three new commits — `SCAFFOLD_TOPUP_SKIPPED` is a real improvement, the single-sourcing fix is the right shape, and D7 passes on the new base

Re-verified on a fresh Linux x86_64 container provisioned per `DOGFOODING.md` (Rust 1.94.1, Nix 2.35.1 with flakes, r0vm 3.0.5 matched to the pinned LEZ `cf3639d8`'s `risc0-zkvm = 3.0.5`, risc0 guest toolchain 1.91.1, risc0 C++ `v2024.1.5`, circuits v0.4.1, real `sequencer_service` + `wallet` built from the pin).

Branch already contains `master@9fcc376`, so unlike my last pass there is no merge to audit. On the branch: `cargo test` **722 green** (532 lib + 182 CLI + 3 testnode + 5 doc), `cargo fmt --check` clean.

Three new commits since `fd90c34`, all responding to the last round. Taking them in order of how much they change my read.

### `f8e035b` — `SCAFFOLD_TOPUP_SKIPPED` closes the actual gap

This is the substantive addition and it is the right one. Without it, a `topup = false` profile tells the *pipeline* to stand down but tells the *hook* nothing, so a self-funding hook has to either always claim (and tolerate an already-funded wallet) or read `scaffold.toml` itself to find out what scaffold did. Exporting the decision makes the hook's job a branch instead of a guess.

Two design choices here are both correct and worth naming:

- **Always set, never absent.** `0` when scaffold topped up, `1` when it did not. This matters more than it looks: had it been "set only when skipped", every hook would have to treat unset as "scaffold topped up", which is indistinguishable from "this scaffold is too old to tell you". The README says this explicitly (*"Hooks should branch on their `1`/`0` value, not on whether they exist"*) and there is a dedicated test for the `0` case, not just the `1` case.
- **Run-level, not per-program.** Step 4 has no per-program dimension, so there is no `SCAFFOLD_TOPUP_SKIPPED_<name>` — matching `SCAFFOLD_DEPLOY_SKIPPED`'s precedent, and covered by a multi-program test that asserts it still reaches the hook when the single-program shortcuts are absent.

Bonus: the README table now documents `SCAFFOLD_DEPLOY_SKIPPED`, which was shipped by #239 but never written down. Fixing an adjacent gap while you are in the file is the right instinct.

### `9c12edb` — this is the fix I actually wanted

The previous shape computed the skip decision in `run_pipeline_once` and passed a separately-derived value to the hook builder. `topup_step_skipped(&params.resolved)` now feeds both the `if` that skips step 4 and the value the hook reads, so the two cannot drift. The comment says exactly that, at the branch:

> The branch below and the value hooks read from `SCAFFOLD_TOPUP_SKIPPED` come from the same expression, so what the hook is told can't drift from what step 4 did.

And the justification for extracting a one-line function is the part I want to endorse, because "this is too trivial to extract" is the usual objection:

> `run_pipeline_once` needs a real sequencer to drive, so an inverted signal here would otherwise reach every hook unchallenged.

That is precisely right. `hook_receives_topup_skipped_*` pin how `build_hook_command` *renders* the flag, but both pass `topup_skipped` in by hand, so neither would notice a polarity inversion in the decision itself. `topup_step_skipped_follows_the_profile_flag` is the only test in the file that would fail if `!resolved.topup` became `resolved.topup`, and that bug would otherwise be invisible until a user's funding hook did the wrong thing against a live sequencer. A one-line function that buys a testable seam is worth it.

I also like `topup_skipped` being threaded as its own parameter rather than stuffed onto `DeployedPrograms`, with the reason recorded at the signature ("step 4's outcome has no per-program dimension, so it does not belong on a struct that describes deployed programs"). That is the kind of comment that stops the next person from "tidying" it into the struct.

### `35ecc62` — the attribution correction is right

The README now attributes the post-`--reset` re-seed to step 3 rather than to step 1, which matches `run.rs`: `reset_for_run` wipes and `reseed_after_wipe` re-seeds *within* step 3, after step 1's chained `setup` has already run. The previous wording implied the setup step would cover it, which would have been wrong for exactly the `--reset` case that most needs it.

The scoping is careful too — *"on a pin whose wallet config ships preconfigured accounts one is already in place by step 4"* correctly does not claim this holds for a v0.2.0 pin, which is #246's territory rather than this PR's.

### Config plumbing fits the existing pattern exactly

`topup` mirrors `deploy` end to end: `unwrap_or(true)` at both the inline and per-profile parse sites, `Default` hand-written (with the comment explaining why `derive(Default)` would be a silent behavior change for every existing project), emitted on serialize only when it deviates from `true`, and `has_inline` extended so `[run] topup = false` alone is enough to keep the section. Round-trip, default-not-serialized, inline, and per-profile cases all have tests. Nothing here invents a new convention.

### D7 on the new base — executed, not inferred

Ran the `topup = false` half of D7 against a real localnet (real `sequencer_service` from the pinned LEZ `cf3639d8`, 5 programs deployed, default wallet seeded and topped up) using this branch's binary and the exact profile the runbook now prescribes.

`run --profile self-fund`:

```
[1/6] Building...
[2/6] Building IDL...
[3/6] Ensuring localnet...
[4/6] Topup skipped (`topup = false` in the run profile; this project funds its own accounts — e.g. via faucet claims at runtime)
[5/6] Deploying programs...
  Succeeded: 5
  Failed: 0
[6/6] Running 1 post-deploy hook(s)...
===> post_deploy[1/1]: echo 'topup skipped:' $SCAFFOLD_TOPUP_SKIPPED
topup skipped: 1
<=== post_deploy[1/1] OK
```

Same hook without the profile:

```
[4/6] Topping up wallet...
[5/6] Deploy skipped (guest binaries + IDL + config + sequencer unchanged; …)
topup skipped: 0
```

Every one of the runbook's new D7 signals matches verbatim: the `[4/6] Topup skipped (…)` header replaces `[4/6] Topping up wallet...`, step 4's destination-address requirement is genuinely bypassed (nothing resolved an address), deploy and hooks still run, and `SCAFFOLD_TOPUP_SKIPPED` reports `1` then `0`. The "always set, never absent" contract therefore holds in a live pipeline and not only in the unit tests — which is the part that would have been easy to get wrong, since the `0` case is the one a hook author will actually rely on.

Incidentally the second run also exercised the deploy cache (`Deploy skipped … unchanged`), which is the other half of the `SCAFFOLD_DEPLOY_SKIPPED` row you added to the README — so that row is accurate for both of its stated causes.

### One thing worth a sentence in the README

Not a defect, and not introduced here — `resolve_profile` **selects**, it does not merge. So `[run] topup = false` combined with `--profile demo` (where `demo` omits `topup`) silently gives you `topup = true`, because the named profile is returned wholesale and never inherits inline values. The README's *"It works inline under `[run]` or per profile"* is accurate but reads like the two compose.

This trap already exists for `reset` and `deploy`, so it is a pre-existing sharp edge rather than something this PR creates — but this PR adds a third key with the same shape, and the failure mode here is the more expensive one: with `deploy` you notice immediately that nothing deployed, whereas an unexpectedly-*enabled* topup on a self-funding project just quietly claims from the faucet. A half-sentence in the profiles section ("a named profile is used as-is; it does not inherit the inline `[run]` values") would cover all three keys at once.

### Minor ergonomics note

`topup` is config-only, with no CLI counterpart — consistent with how #239 shipped `deploy`, so I am not asking you to change it here. But it is worth noticing that `reset` has `--reset` / `--no-reset` for one-off overrides while `deploy` and `topup` have neither, so the only way to try a single self-funding run is to edit `scaffold.toml`. If a third config-only step-skip shows up, that asymmetry is probably worth resolving in one go rather than per-key.

(Same category: `run --help`'s summary line still reads "Build, start localnet, top up wallet, deploy, and run post-deploy hook", which a `topup = false` profile makes partly untrue — as `deploy = false` already did. Not worth a separate fix, but if the summary is ever reworded, both belong in it.)

### Interaction with #246 — worth knowing before merging either

**#245 and #246 each merge cleanly into `master`, but conflict with each other.** One hunk, in `src/commands/run.rs`: the `hook_receives_wallet_home_dir_env` test, where #246 rewrites the hook to `printf` both wallet-home env names and this PR adds the `topup_skipped` argument to `run_post_deploy_hook`. The resolution is to take both sides — dual-name `printf` **and** the trailing `false`.

I verified the combination: merging both onto `master@9fcc376` and resolving that hunk that way gives **731 green, 0 failed**. Textual conflict only, no semantic interaction.

Also worth noting the two PRs are complements rather than overlaps: #246 fixes *why* the topup step had no destination address on a v0.2.0 pin, while this PR lets a project that funds itself opt out of that step entirely. Both are reasonable to land; neither makes the other redundant.

### Verdict

No blocking findings. The single-sourcing fix and the polarity test address the substantive concern from my last pass, and the hook contract is now both correct and documented.

---

danisharora099:
@weboko thanks for the re-review. Since there are no blocking findings, could you submit a formal approval when you get a moment?

weboko (approved):

PR #246, opened 2026-07-30
fix(setup): support the LEZ v0.2.0 wallet across setup/run/doctor (#240)

### LINKED ISSUE #240: Support LEZ v0.2.0 repo layout (wallet crate under lez/) in lgs setup
logos-scaffold 0.1.1's setup expects the pre-v0.2.0 logos-execution-zone repo layout; at v0.2.0 the wallet/sequencer crates moved under `lez/`, so `lgs setup` can no longer locate/build them from the pinned cache. In eth-lez-atomic-swaps we bridge this with a committed script (`scripts/scaffold-setup.sh`, see repo master, merged in PR #36) that builds and symlinks `wallet`/`sequencer` into the expected cache paths — happy to upstream its logic. Related follow-up: at v0.2.0 no client code reads LOGOS_BLOCKCHAIN_CIRCUITS anymore (builtin ELFs ship in the LEZ repo), so scaffold's `[circuits]` handling can likely be retired for v0.2.0+ pins.

danisharora099:
Fixes #240 — plain `lgs setup` (and the downstream `lgs run` pipeline) now works against a LEZ v0.2.0 pin (`a58fbce2`, crate tree under `lez/`) without the compat-symlink + reseed bridge script that eth-lez-atomic-swaps commits today (`scripts/scaffold-setup.sh`).

### 1. Wallet env var rename (`LEE_WALLET_HOME_DIR`)
The v0.2.0 wallet binary reads `LEE_WALLET_HOME_DIR`; scaffold set only the older `NSSA_WALLET_HOME_DIR`, so every wallet subprocess silently created and used `~/.lee/wallet` instead of the project wallet. All wallet invocations now go through a shared helper, `wallet_support::set_wallet_home_env`, which sets **both** names (`constants::WALLET_HOME_ENV_VARS`):

- `src/commands/wallet.rs` — list (plain + `--json`), passthrough, topup preflight/init/pinata (and the `--dry-run` env printout)
- `src/commands/deploy.rs` — both `deploy-program` invocations
- `src/commands/doctor.rs` — `check-health`, plus the `export …` remediation / next-step strings now export both names
- `src/commands/run.rs` — post-deploy hooks receive both vars
- `src/api/mod.rs` — doc comment for `wallet_passthrough`

The helper also passes the path as an `OsStr` (the pre-existing non-UTF8-safety at one call site now applies to all of them).

### 2. Default-wallet seeding without preconfigured accounts
v0.2.0's debug `wallet_config.json` ships no `initial_accounts`, so setup ended with *"could not seed default wallet automatically (no preconfigured public account found)"* and never wrote `default_address=` into `.scaffold/state/wallet.state` — which `cmd_wallet_topup_inner` hard-requires, breaking `lgs run`'s mandatory topup step.

`ensure_default_wallet_seeded` (shared by `lgs setup` and `lgs run --reset`'s `reseed_after_wipe`) now falls back to initializing the wallet storage when the config has no preconfigured public account: it runs `wallet account list` against the project wallet home (the wallet CLI creates a default Public/Private account pair on first use; password piped on stdin via the existing `wallet_password()` machinery), parses the first `Public/<base58>` entry (`wallet_support::first_public_address_in_listing`, base58-validated to 32 bytes), and registers it through the same `write_default_wallet_address` path `wallet default set` uses. A clear log line reports the seeded address; a missing wallet binary degrades to the pre-existing warning.

### 3. Nested repo layout probe (remainder after #234)
#234 already made the wallet/sequencer *config* paths layout-aware; the one production probe still hard-coded to the flat layout was doctor's standalone-support marker (`sequencer/service/Cargo.toml`). It now also probes `lez/sequencer/service/Cargo.toml`.

## Tests
`cargo test` (524 unit + 181 CLI + 3 testnode + doc tests, all green), `cargo fmt --check`, `cargo check --locked`. New coverage:
- `setup.rs`: storage-initialization fallback via a fake wallet script that mirrors the v0.2.0 first-use output and captures its env (locks in that the seeding subprocess sees both env var names); graceful-warning path when no wallet binary exists yet
- `wallet_support.rs`: `first_public_address_in_listing` (v0.2.0 listing shape, invalid-base58 rejection, Private-only output) and `set_wallet_home_env` sets every `WALLET_HOME_ENV_VARS` name
- `run.rs`: post-deploy hook env tests extended to assert both wallet-home vars (incl. the full env-contract test)

## Downstream verification (issue #240's acceptance case)
Throwaway clone of eth-lez-atomic-swaps @ `edd607c` (origin/master), LEZ build cache copied but with the bridge's `wallet`/`sequencer` compat symlinks **removed** from the checkout root, `.scaffold/wallet` and `.scaffold/state` absent, `~/.lee` absent, and this branch's `lgs` first on `PATH`. Plain `lgs setup` (not the bridge script):

```
$ /private/tmp/atomic-swaps-240/.scaffold/lez-cache/repos/lez/a58fbce2…/target/release/wallet account list
default wallet seeded by initializing wallet storage (config ships no preconfigured account)
  Address: Public/CwhyiVAviYvCS5pLZ6FL35hgGEUFxEgUVnAoW61VxZLr
  State file: /private/tmp/atomic-swaps-240/.scaffold/state/wallet.state
setup complete
EXIT=0
```

```
$ cat .scaffold/state/wallet.state
default_address=Public/CwhyiVAviYvCS5pLZ6FL35hgGEUFxEgUVnAoW61VxZLr
$ ls .scaffold/wallet
storage.json  wallet_config.json
```

No layout error, no "could not seed" warning. Wallet-env proof — `~/.lee` did not exist before the run and still doesn't after `lgs setup` + `lgs wallet list` (which lists the seeded project account):

```
$ lgs wallet list
/ Public/CwhyiVAviYvCS5pLZ6FL35hgGEUFxEgUVnAoW61VxZLr
/ Private/9a4xjZennFvCwyU1gyXpMuV7txdBoExBpU4a4WHN2meV
$ ls -la ~/.lee
ls: /Users/danisharora099/.lee: No such file or directory
```

And the pipeline step that motivated the seeding fix, end-to-end against a live localnet:

```
$ lgs localnet start   # localnet ready (sequencer pid=17823)
$ lgs wallet topup
wallet topup preflight: destination is uninitialized; running auth-transfer init
wallet topup complete
  Address: Public/CwhyiVAviYvCS5pLZ6FL35hgGEUFxEgUVnAoW61VxZLr
  Method: pinata faucet claim
  Network: local sequencer (http://127.0.0.1:3040)
```

Not run: the full `lgs run --profile demo` (~15+ min cold app build in the throwaway); the topup exercise above covers the step that previously failed. Follow-up noted in #240 and left out of scope here: retiring `[circuits]` handling for v0.2.0+ pins.


weboko:
## Review — `fix(setup): support the LEZ v0.2.0 wallet across setup/run/doctor (#240)`

Reviewed the full diff and ran the suite on this branch. The Rust is good — this is the right shape for the problem, and the three gaps in the description map cleanly onto the three fixes.

**Verification performed** (Linux x86_64, toolchain provisioned per `DOGFOODING.md`: r0vm 3.0.5 from the curl fallback, circuits v0.4.1, real `sequencer_service` built from the pinned LEZ):

- `cargo test` — **714 green** (524 lib, 181 CLI, 3 testnode, 5 doc), `cargo fmt --check` clean. Baseline `master` is 707; the +6 lib delta matches the new coverage exactly. CI green including both `render-and-build` matrix legs.
- `test-node doctor --json` → `ok: true`, all six checks pass against the v0.1.2 pin — i.e. **no regression on the flat layout** from the `check_standalone_support` probe change.
- I specifically checked the probe-widening for a regression risk and it's safe: `check_standalone_support` is a first-match-wins loop over `fs::read_to_string(...)` with the `Err` case silently skipped, so adding `lez/sequencer/service/Cargo.toml` can only ever *add* a hit on the nested layout. It cannot break a flat checkout where that path is absent. Good.
- **Live pipeline verification on a real localnet** with this branch's binary — `setup` → `localnet` → `topup` → real deploy of all 5 example programs → hooks. The env-var fix is observably working: on `master`/#245 the hook prints `lee wallet home: UNSET`, on this branch it prints the project wallet home. Both names also appear in the `--dry-run` printout, and `doctor` is **24 PASS / 0 WARN / 0 FAIL, "Ready"**:

  ```
  $ lgs wallet topup --dry-run
  NSSA_WALLET_HOME_DIR=/home/user/dogfood-runs/dogfood-default/.scaffold/wallet
  LEE_WALLET_HOME_DIR=/home/user/dogfood-runs/dogfood-default/.scaffold/wallet

  $ lgs run   # post-deploy hooks
  wallet home:     /home/user/dogfood-runs/dogfood-default/.scaffold/wallet
  lee wallet home: /home/user/dogfood-runs/dogfood-default/.scaffold/wallet
  ```

- I could **not** exercise a real v0.2.0 pin end-to-end here (`DEFAULT_LEZ` is still v0.1.2 `cf3639d8`, and standing up `a58fbce2` plus a matching spel pin is outside what this PR changes). The v0.2.0 evidence in the description is detailed and self-consistent, and I'm taking it at face value — noting it as unreproduced on my side.

Things I like: routing every wallet invocation through one `set_wallet_home_env` helper rather than sprinkling a second `.env(...)` at each call site is the right call, and it incidentally fixes the pre-existing `to_string_lossy` non-UTF8 corruption at five sites that previously only `wallet_list_for_project` got right. The `first_public_address_in_listing` base58-validation-as-parser-guard is a nice touch, and `seed_default_wallet_by_initializing` correctly lands *after* the wallet build in `setup_for_project`, so `wallet_binary.exists()` is meaningful.

The blocking issue is that the fix stops at the code and never reaches the documented surface.

---

### 1. The hook env contract is a declared public surface, and this PR changes it without updating it

`ADR.md:345` says so explicitly:

> ## Hook Env Contract is a Documented Public Surface
> Post-deploy hooks run via `sh -c` ... The env they see is **stable, documented in README, and validated by unit and integration tests**

This PR adds `LEE_WALLET_HOME_DIR` to that contract and updates the *tests* but neither the README nor the ADR. Concretely stale after this change:

- `README.md:157` — hook env list: `SEQUENCER_URL`, `NSSA_WALLET_HOME_DIR`, `SCAFFOLD_PROJECT_ROOT`, `SCAFFOLD_IDL_DIR`
- `README.md:253` — env var table row for `NSSA_WALLET_HOME_DIR`
- `FURPS.md:75` — same four-var contract, restated as a requirement
- `ADR.md:351` — same list

And `ADR.md:358` is now actively misleading, because it records a decision this PR supersedes:

> `NSSA_WALLET_HOME_DIR` keeps its upstream-wallet name rather than being renamed to a `SCAFFOLD_*` prefix: hooks that exec the wallet binary ... need the var under the name the binary's `WalletCore::from_env()` reads.

The reasoning is still right, but the conclusion is now "…and since v0.2.0 renamed it, scaffold sets *both* upstream names." That paragraph should say so, otherwise the next person to touch this re-derives it from scratch.

### 2. The shipped AI skills still tell users to export the variable that silently fails on v0.2.0

This is the part I'd most want fixed, because these files are materialized into **every generated project** (scenario `E3`) and are what an agent or a new user actually follows:

- `skills/lez-template/SKILL.md:101,150,189` — `export NSSA_WALLET_HOME_DIR="$(pwd)/.scaffold/wallet"`
- `skills/lez-framework-template/SKILL.md:177,232` — same
- `skills/lgs-cli/SKILL.md:133,144` — same, incl. the troubleshooting row for `wallet -- check-health` failures

On a v0.2.0 pin, following those instructions produces *exactly the bug this PR fixes*: the wallet silently uses `~/.lee/wallet` instead of the project wallet, and — per the description — does so without any error. A user hitting that will land on the skill's troubleshooting table, which will tell them to re-export the variable that isn't being read. Both names should be exported in each of these.

Same applies to the two runner exports in the runbook: `DOGFOODING.md:490` (D6) and `DOGFOODING.md:759` (L4).

### 3. `DOGFOODING.md` has no coverage for this at all

The PR adds support for a whole new LEZ generation and changes what `setup` does (it now *creates* wallet storage as a fallback), but the runbook is untouched. Given the maintenance note at the top of that file, I'd expect:

- **D7**, whose hook example at `:553` prints `$NSSA_WALLET_HOME_DIR` and whose Failure Signals at `:592` list the four-var contract, to name both vars — otherwise D7 passes while silently not covering the new var.
- **D1/D4** to gain the v0.2.0 seeding signal, so a dogfooder knows `default wallet seeded by initializing wallet storage (config ships no preconfigured account)` is a **pass** and not a degraded path. Right now the only documented outcome for a config without preconfigured accounts is the `warning: could not seed…` line this PR replaces.
- The **"Change → Scenario"** map (`:1655`+) to route wallet-env / wallet-seeding changes at all — there's currently no entry that would have told you to rerun anything for this diff.

### 4. `SECURITY.md:23` — `setup` is now a password-submitting command, and it *creates* key material

```
This override applies to scaffold commands that submit wallet password input (`wallet`, `deploy`, `doctor` checks).
```

`setup` now belongs in that list: `seed_default_wallet_by_initializing` pipes `wallet_password()` on stdin. More than a list update, though — the *nature* of the operation changed. Previously the deterministic password only ever **unlocked** storage that the debug config had pre-seeded. Now, on a v0.2.0 pin, scaffold **creates a brand-new keypair** whose password is the public constant `logos-scaffold-v0` (`constants.rs:59`), during the very first command a user runs, with the address written to `wallet.state` as the default topup destination.

That is defensible for a local dev tool and `SECURITY.md` already says not to use these wallets for real funds — but "scaffold generates your key material under a well-known password unless you set `LOGOS_SCAFFOLD_WALLET_PASSWORD` *before your first `setup`*" is a materially stronger statement than what's on the page today, and the ordering constraint is new and easy to get wrong. Worth a sentence.

---

### Minor / non-blocking

**`first_public_address_in_listing` scans all whitespace-delimited tokens, including prose.** The doc comment argues base58-validation makes prose safe, and your own test bakes in `note: Public/not-base58 is not an address` being skipped — but the guard is "is it a valid 32-byte base58 id", not "is it an account listing entry". Any future wallet build that mentions a *real* address in a banner, hint, or warning line before the listing would have that address silently adopted as the project default topup destination. Since entries have a known shape (`/ Public/<id>`), anchoring on it would be strictly tighter:

```rust
for line in output.lines() {
    let Some(rest) = line.trim().strip_prefix("/ ") else { continue };
    // …then the existing Public/ + normalize_address_ref check
}
```

Low likelihood, but the failure mode is silent and the blast radius is "funds go to the wrong account", so I'd rather over-constrain the parser.

**`run_with_stdin` writes the password once.** The fake wallet in the new test consumes no stdin, so the test wouldn't catch it if a first-use v0.2.0 wallet prompted twice (password + confirmation). Your downstream run proves it doesn't today — just noting the test doesn't pin that assumption.

---

The code change is correct and I'd approve it on the Rust alone. What's holding me back is that §1–§3 leave the repo in a state where every *documented* path still teaches the broken thing — which for a fix whose entire failure mode is "silent" is the part most likely to bite the next person.


---

weboko (commented):
## Review: `fix(setup): support the LEZ v0.2.0 wallet across setup/run/doctor (#240)`

Built and tested locally (Linux x86_64): `cargo fmt --check` clean and the full suite green — 524 lib + 181 CLI + 3 testnode + 5 doc, matching your description exactly. This is a well-scoped fix for a real and nasty failure mode, and the call-site coverage is genuinely complete — I grepped for stragglers and there is **no** remaining raw `.env("NSSA_WALLET_HOME_DIR", …)` anywhere in `src/`. Nice.

Two things I'd like addressed: one hardening fix in the new parser, and a docs sweep that I think is the bigger of the two.

### What checks out

- **Complete call-site conversion.** `wallet.rs` (list ×2, passthrough, topup preflight/init/pinata, `--dry-run` printout), `deploy.rs` (both `deploy-program` sites), `doctor.rs` (`check-health` + both remediation strings), `run.rs` (hook env), `api/mod.rs` (doc). Routing all of it through one `set_wallet_home_env` helper is the right call — this is exactly the kind of change that rots when it's done by hand at N sites.
- **The `OsStr` fix is a real improvement, not just a refactor.** Three of the converted sites previously did `.as_os_str().to_string_lossy().to_string()`, which corrupts a non-UTF8 wallet-home path. The helper takes `impl AsRef<OsStr>` and passes it through verbatim, so the one site that was already correct became the behavior everywhere. Worth calling out in the changelog if there is one.
- **Ordering in `setup_for_project` is correct.** `cargo build -p wallet` runs before `ensure_default_wallet_seeded`, so the `wallet_binary.exists()` guard passes on the normal path and the `Ok(None)` degradation is reserved for genuinely broken setups. The dedicated test for the missing-binary case is good.
- **No credential leakage.** `run_with_stdin` pipes the password on stdin and captures stdout/stderr rather than streaming them, and the echoed `$ …` line contains only argv. The seeded address is printed, which is public information. Fine.
- **`check_standalone_support` is additive.** I checked this specifically because adding an entry to a path array can flip an all-match into a failure — it's an any-match loop with `read_to_string` errors skipped, so the extra `lez/sequencer/service/Cargo.toml` probe cannot regress the flat layout.
- **The env-capture fake-wallet test is the right test.** Asserting that the *subprocess* saw both var names, rather than asserting on the helper in isolation, is what actually pins the bug this PR fixes.

### 1. `first_public_address_in_listing` should anchor to listing entries

The parser scans every whitespace-separated token in the output and accepts the first one that starts with `Public/` and base58-decodes to 32 bytes:

```rust
for token in output.split_whitespace() {
    let Some(rest) = token.strip_prefix("Public/") else { continue };
    …
}
```

The doc comment says entries render as `/ Public/<account-id>` and argues that base58 validation means "surrounding prose that merely mentions `Public/` cannot be mistaken for an address". That holds for *invalid* prose — which is what `first_public_address_in_listing_skips_invalid_account_ids` covers — but not for prose containing a *valid* address. I confirmed this against the branch by adding a temporary test:

```rust
let other = "8ZcHmiiJfuoSbPWkfWyHrLkQE1Sf3JzeuTVXwSCsK3XV";
let output = format!(
    "warning: previous default Public/{other} is unavailable\n\
     / Public/{ACCOUNT_ID}\n"
);
assert_eq!(first_public_address_in_listing(&output).as_deref(),
           Some(format!("Public/{ACCOUNT_ID}").as_str()));
```

which fails:

```
left:  Some("Public/8ZcHmiiJfuoSbPWkfWyHrLkQE1Sf3JzeuTVXwSCsK3XV")
right: Some("Public/6iArKUXxhUJqS7kCaPNhwMWt3ro71PDyBj7jwAyE2VQV")
```

The prose address wins over the real account entry. That value is then written straight to `.scaffold/state/wallet.state` as `default_address=`, which is the destination `lgs wallet topup` funds and the identity the rest of the pipeline uses — so a wrong parse here silently points the project's funding at an account the user never created, with no error anywhere.

This isn't only a synthetic case. Here is **real** `wallet account list` output I captured from a live localnet run against the pinned LEZ (`cf3639d8`):

```
Preconfigured Public/2RHZhw9h534Zr3eq2RGhQete2Hh667foECzXPmSkGni2
Preconfigured Public/CbgR6tj5kWx5oziiFptM7jMvrQeYY3Mzaao6ciuhSr2r
Preconfigured Private/9DGDXnrNo4QhUUb2F8WDuDrPESja3eYDkZG5HkzvAvMC
/ Public/2koS3jXbkvWY5QHMXmgrFLSy5jxfApaoYFvy4Xku9yvj
/ Private/E8wf4HmfmoUWrNWtT5cz69EDRNYP2Uz5hMDRxBZ7Tg9W
```

The `Preconfigured …` lines are **not** account entries, but they carry valid 32-byte base58 addresses. Feeding this verbatim to `first_public_address_in_listing` on your branch returns `Public/2RHZhw9h…` — a preconfigured address — rather than `Public/2koS3jXb…`, the account the wallet actually initialized into storage. So the "prose containing a valid address" shape is not hypothetical; it is what this wallet prints today.

To be precise about the blast radius, because I don't want to overstate it: that exact combination is **not currently reachable**, because when the config ships `initial_accounts` the earlier `first_public_wallet_address` branch intercepts and the fallback never runs. The fallback only fires when the config has no preconfigured accounts, and in that case the `Preconfigured` lines are absent. So there is no live bug today.

The objection is that the safety here is entirely incidental — it rests on two independent facts (which lines the wallet prints, and which branch runs first) staying in their current alignment, on a code path whose whole reason for existing is that a new LEZ version changed the wallet's behavior out from under scaffold. The objection is that the only thing standing between "wallet prints a diagnostic mentioning some other address" and "scaffold adopts it as the default wallet" is the accident that such diagnostics don't exist yet — and this is a value with financial meaning, parsed from output the wallet is free to change on any pin. Anchoring to the entry format the doc comment already specifies costs nothing and makes the guarantee structural:

```suggestion
pub(crate) fn first_public_address_in_listing(output: &str) -> Option<String> {
    for line in output.lines() {
        // Account entries render as `/ Public/<account-id>`; anchoring to the
        // entry marker keeps prose that merely mentions a (valid) address from
        // being adopted as the default wallet.
        let Some(entry) = line.trim_start().strip_prefix("/ ") else {
            continue;
        };
        let Some(rest) = entry.trim_start().strip_prefix("Public/") else {
            continue;
        };
        // Trim anything after the base58 run (e.g. trailing punctuation).
        let account_id: String = rest
            .chars()
            .take_while(char::is_ascii_alphanumeric)
            .collect();
        if let Ok(normalized) = normalize_address_ref(&format!("Public/{account_id}")) {
            return Some(normalized);
        }
    }

    None
}
```

The three existing tests still pass against this (they all use `/ `-prefixed entries), and the probe above then passes too. Worth adding it as a fourth test so the property is pinned rather than incidental.

### 2. Docs are stale across the board — including instructions users are told to run

This is the part I'd most like fixed. The PR changes the wallet-home env contract everywhere in code but touches **no** documentation, and `DOGFOODING.md` in particular is untouched despite this being new behavior in `setup`, `run`, and `doctor`. The stale references fall into three tiers:

**Tier 1 — literal commands that are now wrong on a v0.2.0 pin.** These are worst because they fail *silently* in exactly the way #240 describes: the export is a no-op, the wallet falls back to `~/.lee/wallet`, and the runner appears to work against the wrong wallet home.

- `DOGFOODING.md:490` — `export NSSA_WALLET_HOME_DIR="$PWD/.scaffold/wallet"` (D4)
- `DOGFOODING.md:773` — same export (D6)
- `skills/lez-template/SKILL.md:101,150,189` — same export, plus "if you hit the wallet `from_env()` panic … you forgot `export NSSA_WALLET_HOME_DIR=…`"
- `skills/lez-framework-template/SKILL.md:177,232`
- `skills/lgs-cli/SKILL.md:133` — the `wallet -- check-health` troubleshooting row

The skills files matter more than their location suggests: they're installed into **every generated project** under `.claude/skills/`, so this guidance ships to users and to agents operating in those projects. A v0.2.0 user following the documented remediation will be told to export a variable their wallet ignores.

**Tier 2 — contract descriptions that are now incomplete.**

- `README.md:157` — the prose list of hook env vars (`SEQUENCER_URL`, `NSSA_WALLET_HOME_DIR`, …); hooks now receive `LEE_WALLET_HOME_DIR` too.
- `README.md:253` — the env var table.
- `DOGFOODING.md:553` — the D7 hook example still echoes only `$NSSA_WALLET_HOME_DIR`.
- `DOGFOODING.md:605` — the env-contract failure signal ("missing any of `SEQUENCER_URL` / `NSSA_WALLET_HOME_DIR` / …"), which is the line a dogfooder checks against.
- `DOGFOODING.md:521,789,800` and `skills/lgs-cli/SKILL.md:101,144`.

**Tier 3 — the ADR, which explicitly owns this decision.** `ADR.md` has a section titled *"Hook Env Contract is a Documented Public Surface"* that enumerates the vars (`ADR.md:351`) and then justifies the naming (`ADR.md:358`):

> `NSSA_WALLET_HOME_DIR` keeps its upstream-wallet name rather than being renamed to a `SCAFFOLD_*` prefix: hooks that exec the wallet binary … need the var under the name the [wallet expects].

That rationale is precisely *why* a second name is now required — the recorded decision anticipated this and should be updated to say the contract is now "every name the pinned wallet may read", sourced from `WALLET_HOME_ENV_VARS`. Changing a surface the ADR calls public without touching the ADR is the thing I'd least like to see merged.

**On DOGFOODING specifically:** beyond fixing the stale lines, this PR adds a genuinely new, user-visible behavior — `setup` can now seed the default wallet by *initializing wallet storage*, printing `default wallet seeded by initializing wallet storage (config ships no preconfigured account)`. Nothing in D1/D4 tells a dogfooder that this line is an expected success signal rather than a warning, and D1's failure-signal list still says only that a missing default address is a regression. A short addition to D1's expected signals (both seeding paths are valid, distinguished by which line prints) would make the new path observable in the scenario that's supposed to cover it.

- `WALLET_HOME_ENV_VARS` is ordered `[NSSA, LEE]` and both are always set to the same value, so order is irrelevant today. If a future pin ever reads both with different precedence this becomes load-bearing — a one-line comment saying "both are set to the same path; order is not significant" would prevent someone from assuming it is.
- `seed_default_wallet_by_initializing` relies on `wallet account list` having the side effect of creating the account pair on first use. That's documented in the doc comment, which is good, but it does mean `setup` now mutates wallet state through a read-shaped command. If the wallet ever gains an explicit `wallet init`, prefer it.
- The doctor remediation string is now `export NSSA_WALLET_HOME_DIR=$(pwd)/.scaffold/wallet LEE_WALLET_HOME_DIR=$(pwd)/.scaffold/wallet`. Valid shell and correctly hoisted into the shared `STEP_EXPORT_WALLET_HOME` const — just noting it's long enough that the `--json` consumer sees a fairly unwieldy string.
- Scoping out `[circuits]` retirement for v0.2.0+ pins was the right call for this PR's size.

### Verification notes

I provisioned the full stack per DOGFOODING's "Provisioning the Real LEZ Sequencer Toolchain" (Nix 2.35.1, r0vm 3.0.5 matched to the LEZ tree's `risc0-zkvm` version, real `sequencer_service` built from the pinned checkout, `lgs setup` complete, localnet ready on `:3040`) and ran the D4 wallet scenario against your branch on a live localnet.

**The env-var fix is confirmed working.** `wallet topup --dry-run` on `master` prints one name; on this branch it prints both:

```
dry-run: wallet topup command will not be executed
NSSA_WALLET_HOME_DIR=/tmp/dogfood/dogfood-default/.scaffold/wallet
LEE_WALLET_HOME_DIR=/tmp/dogfood/dogfood-default/.scaffold/wallet
```

**No regression on a v0.1.x pin**, which is the main risk of setting an extra variable unconditionally. Against `cf3639d8`:
- `wallet list` — enumerates correctly from the project wallet home
- `wallet topup` — real faucet claim succeeded (`wallet topup complete`, pinata claim, local sequencer)
- `wallet -- check-health` — `✅All looks good!`

**Not exercised:** the storage-initialization seeding fallback itself. On this pin the config ships `initial_accounts`, so `setup` took the pre-existing branch and printed `default wallet seeded from preconfigured account` — the new path is unreachable without a v0.2.0 pin (`a58fbce2`). Your downstream evidence in the description covers that path, and the `~/.lee` absence proof is the right way to have shown the env half. I verified the parser behavior directly against real listing output instead, as above.

---

weboko:
## Re-review on the new base — the fix survives `#241` intact; all four doc findings still stand, anchors refreshed

No new commits since `aeb39dd`, so this is not a re-review of changed code. I'm posting because **the base moved after my 2026-07-31 pass**: `#241` (`9fcc376`) landed on `master` on 2026-07-31 21:39 UTC, ~17 hours later, and it is the one change most likely to interact with this PR — it added **+299 lines to `src/commands/deploy.rs`** and also touched `run.rs` and `doctor.rs`, all three of which this PR rewires. It also rewrote `DOGFOODING.md`, which my §3 anchors into.

### The interaction check that actually mattered

A clean auto-merge would not have told me much here. The invariant this PR establishes is *"every wallet subprocess gets both env var names"*, and that invariant is exactly the kind a large concurrent diff to `deploy.rs` can silently punch a hole in — a new `Command::new(&wallet.wallet_binary)` added by `#241` would compile fine, merge fine, and be wrong.

So I merged onto `master@9fcc376` and audited the result directly:

- **Every** wallet spawn in the merged tree routes through the helper — `deploy.rs:213`, `deploy.rs:553`, `doctor.rs:296`, `setup.rs:201`, and six sites in `wallet.rs`.
- **Zero** raw `.env("NSSA_WALLET_HOME_DIR", ...)` call sites remain; the only surviving mention of the literal is the `WALLET_HOME_ENV_VARS` definition in `constants.rs:66`.
- `#241` added deploy pacing and spel `program-id` handling but **no new wallet invocation**, so there is no gap. The centralize-into-one-helper approach is what made this cheap to verify — a second `.env(...)` sprinkled per call site would have required checking each one by hand.

`cargo test` on the merge result — **720 green, 0 failed** (530 lib, 182 CLI, 3 testnode, 5 doc).

### Live re-verification

Fresh Linux x86_64 container per `DOGFOODING.md` (Rust 1.94.1, Nix 2.35.1 flakes, r0vm 3.0.5 matched to the pinned LEZ `cf3639d8`'s `risc0-zkvm = 3.0.5`, guest toolchain 1.91.1, circuits v0.4.1, real sequencer + wallet from the pin), against a live localnet with 5 real programs deployed:

```
$ lgs wallet topup --dry-run          # this branch merged onto master
NSSA_WALLET_HOME_DIR=/home/user/dogfood-runs/dogfood-default/.scaffold/wallet
LEE_WALLET_HOME_DIR=/home/user/dogfood-runs/dogfood-default/.scaffold/wallet

$ lgs wallet topup --dry-run          # master binary, same project, for contrast
NSSA_WALLET_HOME_DIR=/home/user/dogfood-runs/dogfood-default/.scaffold/wallet
```

Post-deploy hooks see both names (`nssa:[...]` and `lee:[...]` both resolve to the project wallet home), and `doctor` is **24 PASS / 0 WARN / 0 FAIL, "Ready"** — so the widened `check_standalone_support` probe still causes no regression on the flat v0.1.2 layout.

As before, I could **not** exercise a real v0.2.0 pin end to end — `DEFAULT_LEZ` is still v0.1.2 `cf3639d8`, so `setup` took the preconfigured-account path (`default wallet seeded from preconfigured account`) and the new `seed_default_wallet_by_initializing` fallback never ran on my box. That path remains covered by your unit tests plus the downstream evidence in the description, not by my run.

---

### The four findings are unchanged — here are the refreshed anchors

All still reproduce verbatim on `master@9fcc376`. Only `DOGFOODING.md` moved (`#241` shifted it ~+26 lines); everything else is at the same line.

**§1 — hook env contract documented but not updated.** Unchanged anchors: `README.md:157` (hook env list), `README.md:253` (env var table), `FURPS.md:75`, `ADR.md:345/351`, and `ADR.md:358` — the paragraph that records the now-superseded "keeps its upstream-wallet name" decision. `FURPS.md:105` is worth adding to that list; it also names only `NSSA_WALLET_HOME_DIR`.

**§2 — shipped AI skills still teach the variable that silently fails on v0.2.0.** Unchanged: `skills/lez-template/SKILL.md:98,101,150,189`, `skills/lez-framework-template/SKILL.md:177,232`, `skills/lgs-cli/SKILL.md:101,133,144`. These materialize into every generated project, so on a v0.2.0 pin they reproduce the exact bug this PR fixes, and the troubleshooting row sends the user to re-export the variable that isn't read.

**§3 — `DOGFOODING.md` coverage, now at these lines:**

| what | was | now |
| --- | --- | --- |
| D6 runner export | `:490` | **`:514`** |
| D7 hook example printing `$NSSA_WALLET_HOME_DIR` | `:553` | **`:579`** |
| D7 Failure Signals (four-var contract) | `:592` | **`:618`** |
| L4 runner export | `:759` | **`:785`** |
| "Minimum Rerun Guidance" map | `:1655+` | **`:1684+`** |

Also `:547`, `:804`, `:815`, `:1639` name only `NSSA_WALLET_HOME_DIR` in prose. I ran D7 on the new base and confirmed the concrete consequence: the runbook's hook block prints `wallet home:` from `$NSSA_WALLET_HOME_DIR` only, so **D7 passes green while providing zero coverage of the variable this PR adds** — which is the failure mode worth fixing, given the whole bug is silent. The map at `:1684+` still has no entry that would route this diff anywhere.

**§4 — `SECURITY.md:23` unchanged**, still listing only `wallet`, `deploy`, `doctor`. `setup` now submits wallet password input and, on a v0.2.0 pin, *creates* key material under the public constant `logos-scaffold-v0` during the first command a user runs — with the ordering constraint that `LOGOS_SCAFFOLD_WALLET_PASSWORD` must be set *before* the first `setup`. Still my view that this is defensible for a local dev tool and still my view that it needs a sentence.

### The one nit I'd promote

Everything else I flagged as minor stays minor, but on re-reading I'd push harder on anchoring `first_public_address_in_listing` to the entry shape:

```rust
for line in output.lines() {
    let Some(rest) = line.trim().strip_prefix("/ ") else { continue };
    // …then the existing Public/ + normalize_address_ref check
}
```

The current guard answers "is this a valid 32-byte base58 id", not "is this an account-listing entry", so it scans banners and prose too. The reason I'd rather over-constrain: the address this picks becomes the default **topup destination** written to `wallet.state`, the adoption is silent, and the trigger is any future wallet build that prints a real address in a banner before the listing. Cheap to tighten now, unpleasant to debug later.

---

**Net:** the Rust is correct, survives `#241` with the invariant intact (verified per call site, not just by a clean merge), 720 green, and observably working on a live localnet. My position is unchanged from 2026-07-31: I'd approve on the code alone. What holds me is that §1–§3 leave every *documented* path — README, ADR, FURPS, the runbook, and the skills shipped into every generated project — still teaching the thing that breaks, for a bug whose entire signature is that it fails silently.

---

danisharora099:
Thanks for both passes — the re-anchored line numbers made this straightforward to work through. Pushed `faccf82` + `fcd80db` + `11a577e`.

### §1 — hook env contract

- `README.md` — the `run` prose clause and a `LEE_WALLET_HOME_DIR` row in the env table, both now saying which pin reads which name.
- `ADR.md:351` — contract bullet lists both.
- `ADR.md:358-362` — **rewritten, not appended.** You were right that the paragraph records a superseded decision. The keep-the-upstream-name reasoning survives; the conclusion is now "…and since v0.2.0 renamed it upstream, scaffold sets both names, to the same path." I also added an explicit rule for the next rename: add the new name, keep the old, never swap.
- `FURPS.md:75` and `:105`.

### §2 — the shipped skills

All three files: `lez-template` (98/101/150/189), `lez-framework-template` (177/232), `lgs-cli` (101/133/144, plus a new troubleshooting row for the "commands succeed but the wallet looks empty" shape, which is what the silent fallback actually looks like from the user's side).

### §3 — DOGFOODING

D6 (491, 522), D7 (555, 594), L4 (761, 777, 788), T4 (1612), plus a new rerun-guidance bullet routing `WALLET_HOME_ENV_VARS` / `set_wallet_home_env` / `ensure_default_wallet_seeded` / LEZ-pin-across-v0.2.0 changes at D1, D4, D6, D7. D1 and D4 now quote both seeding lines verbatim from `setup.rs` and state that **both are a PASS** — only `warning: could not seed default wallet automatically` is a failure.

### §4 — SECURITY.md

`setup` added to the list, plus a paragraph on the part that actually changed: on a v0.2.0 pin the deterministic password now guards **freshly created** key material, and `LOGOS_SCAFFOLD_WALLET_PASSWORD` has to be exported before the first `setup` (or first `run`, which chains it). Remediation is `run --reset`, which I verified goes `reset_for_run` → `cmd_localnet_reset(reset_wallet=true)` → `reseed_after_wipe`.

### The promoted nit — and one thing I'd like your read on

Took it: `first_public_address_in_listing` now anchors on the `/ ` entry prefix, with a new test proving a real address in a banner line is no longer adopted (it fails against the old token scanner).

But I added a **fallback** you didn't ask for, and I think it's load-bearing. Requiring the marker converts a hypothetical wrong-address risk into a hypothetical **no-address** risk — and the failure mode of the latter is exactly #240, `wallet.state` with no `default_address=`. Our own fake wallet in `tests/cli.rs:3706-3707`, written to imitate real `wallet account list` output, prints two valid public accounts and **one has no `/ ` marker**. So the scan now prefers `/ `-prefixed entries and falls back to the old behaviour when none yields a usable address; it can never return `None` where the old code returned `Some`. There's a test pinning which one wins on fixture-shaped input so the choice is visible rather than silent.

**What I can't establish from this repo: whether the `/ ` marker actually appears on a real v0.2.0 `wallet account list`.** You had a real pin in front of you — if you still do, could you confirm? The fallback bounds the risk but doesn't answer the question.

### Two smaller things

- **"Every wallet subprocess" needed scoping.** `doctor.rs:278` spawns `wallet --version` with no `set_wallet_home_env` — every other wallet `Command` in `src/` has one on the next line. It reads no wallet state so it's harmless, but the ADR and the `WALLET_HOME_ENV_VARS` doc comment both claimed "every", so I scoped the claim rather than changing behaviour. Say the word if you'd rather have the call added.
- **A D4 gap I did not fill.** Given the SECURITY change, the higher-value check is "export the override before the first `setup` on a v0.2.0 pin, then confirm the seeded wallet opens with it" — but I can't verify the observable signal without a real v0.2.0 pin, and writing the steps blind would be guesswork. Flagging rather than inventing it.

`cargo test` 714 → 715 (527 lib + 181 CLI + 3 testnode + 5 doc), `cargo fmt --check` clean, `cargo test template::skills` 12 passed including the byte-identity assertion. Full `rg NSSA_WALLET_HOME_DIR` audit: every remaining hit is either paired with `LEE_WALLET_HOME_DIR` on the same or adjacent line, or is the `WALLET_HOME_ENV_VARS` definition itself.

Still on v0.1.2 here, so as with your runs the v0.2.0 seeding path is covered by unit tests rather than by a live pin.

danisharora099:
Rebased onto `master@9fcc376` — the doc additions above collided with #241, which rewrote the same regions. Both conflicts were content, not just line drift, so noting the resolutions:

- **README env table** (#245): #241 renamed `spel inspect` → `spel program-id` in the `SCAFFOLD_PROGRAM_ID` row. Kept master's wording; the two new rows sit above it.
- **DOGFOODING D6/L4** (#246): #241 had improved both blocks — D6 now uses two distinct accounts (`<account-id-a>` / `<account-id-b>`) with the `Public/` prefix, and L4 gained `export HOST_CC=cc HOST_CXX=c++`. Kept all of that; only the `export` line picked up the second var name.

---

**Unrelated, but you'll hit it:** #241's deploy-pacing tests are flaky on macOS, and it isn't caused by either branch — I reproduced it on a clean detached checkout of `origin/master@9fcc376`: run 1 failed, runs 2 and 3 passed, same tree.

Affected: `commands::deploy::tests::wait_for_block_past_returns_true_once_head_advances`, `wait_for_block_past_times_out_when_head_stalls`, and `deploy_aborts_remaining_programs_when_head_stalls` in `tests/cli.rs`. A different member of that set fails on each run, and each passes in isolation (3/3). Two failure shapes:

```
panicked at src/commands/deploy.rs:900:9:
head reached 6 > 5, pacing wait must succeed

panicked at ureq-2.12.1/src/response.rs:403:34:
failed to read exact buffer length from stream: Os { code: 22, kind: InvalidInput }
```

The `ureq` one is the same signature you saw against the local mock. Looks like the mock sequencer these tests stand up contends under parallel test execution. Linux CI is green, so it may be macOS-only — but it makes `cargo test` unreliable locally on a Mac, which is worth its own issue against `master` rather than being carried by either of these PRs.

danisharora099:
@weboko ready for another pass when you have time — all four findings plus the promoted parser nit are in the commits above, and it's rebased onto `master@9fcc376` (D6/L4 collided with #241; I kept its improved blocks and only reapplied the second var name — details in the comment above). CI green including both `render-and-build` legs.

**One question I'd genuinely value your answer on, since you had a real pin in front of you and I don't:** does the `/ ` prefix actually appear on a real v0.2.0 `wallet account list`?

I took the anchoring nit, but added a fallback you didn't ask for, because requiring the marker turns a hypothetical wrong-address risk into a hypothetical *no*-address risk — and that failure mode is #240 itself. Our own fake wallet in `tests/cli.rs` prints two valid public accounts and one has no marker, so the seeded default silently changed against that fixture. The scan now prefers marked entries and falls back to the old behaviour otherwise, so it can never return `None` where the old code returned `Some`. But the fallback bounds the risk rather than answering the question.

weboko:
## Re-review of the three docs commits — all four of my findings are addressed, the listing-scan doc now matches the shipped code, and the security disclosure is the right one

Re-verified on a fresh Linux x86_64 container provisioned per `DOGFOODING.md` (Rust 1.94.1, Nix 2.35.1 with flakes, r0vm 3.0.5 matched to the pinned LEZ `cf3639d8`'s `risc0-zkvm = 3.0.5`, risc0 guest toolchain 1.91.1, risc0 C++ `v2024.1.5`).

Branch already contains `master@9fcc376` (`git merge origin/master` → *Already up to date*), so there is no merge drift to audit this time. On the branch: `cargo test` **723 green** (533 lib + 182 CLI + 3 testnode + 5 doc), `cargo fmt --check` clean.

The code is unchanged since `d5e8929`; `f601053`/`691e811`/`212c4dc` are docs. I re-audited the code anyway, because two of the doc claims are only true if the code says what the docs now say it says.

### The invariant still holds, and it is now greppable

The property this PR establishes — *every wallet subprocess gets both env var names* — is intact on the current head:

- Zero raw `.env("NSSA_WALLET_HOME_DIR", …)` call sites remain anywhere in `src/`. The only surviving mention of the bare literal outside `constants.rs:67` is a test assertion in `doctor.rs:522`.
- Every `Command::new(&wallet.wallet_binary)` / `Command::new(wallet_binary)` site is immediately followed by `set_wallet_home_env`: `wallet.rs` 57/58, 153/154, 174/175, 239/240, 247/248, 255/256; `deploy.rs` 212/213, 552/553; `doctor.rs` 295/296; `setup.rs` 200/201.
- `doctor.rs:278` (`wallet --version`) is the one site without it, which is exactly the deliberate exception `constants.rs:65-66` documents. Good that the exception is written down at the constant rather than at the call site — that is where someone adding a third name will look.
- `run.rs`'s hook builder iterates `WALLET_HOME_ENV_VARS` directly rather than calling the helper. That is fine and stays correct under a future rename since both paths read the same const, though `set_wallet_home_env(&mut cmd, &wallet_home)` would type-check unchanged there (it takes `impl AsRef<OsStr>`) and would make the invariant enforceable by grepping for a single function name. Purely a nit.
- `reseed_after_wipe` (`run.rs:521`) passes `lez.join(WALLET_BIN_REL_PATH)`, matching what `setup_for_project` passes, so `run --reset` and a fresh `setup` seed through identical primitives.

### `212c4dc` — the listing-scan doc now matches the code exactly

This was worth checking character by character, because it is the commit that claims to reconcile docs with code. The runbook now says seeding

> adopts the first `Public/` account on a `/ `-prefixed listing line, ignoring `Public/` addresses printed anywhere else (a banner or status line); only if no `/ `-prefixed line yields a usable `Public/` address does it fall back to the first `Public/` token anywhere in the output.

`first_public_address_in_listing` (`wallet_support.rs:149-159`) is precisely that, in that order: `lines().find_map(strip_prefix("/ ") → public_address_at_start)`, then `if marked.is_some() { return marked }`, then `output.split_whitespace().find_map(public_address_at_start)`. The two-tier structure and the "only if" are both real. The base58 validation is real too — `public_address_at_start` funnels through `normalize_address_ref`, so a candidate that decodes to anything other than 32 bytes returns `None` and the scan moves on rather than writing a malformed default address.

One consequence worth being aware of rather than fixing: `take_while(char::is_ascii_alphanumeric)` will over-consume if a `Public/<id>` is immediately followed by alphanumerics with no separator. That fails the 32-byte check and falls through to the next candidate, so it degrades to "no address found" rather than to a wrong address — the safe direction.

### The security disclosure is the right one, and it is in the right places

This is the part of the PR I care most about, and I think it is now handled correctly.

The fallback path does something materially different from the preconfigured path: it **creates fresh key material**, guarded by a deterministic password that defaults to a constant. The old docs described `setup` as seeding "from preconfigured public accounts", which would have left a reader assuming the password only ever guarded storage that shipped pre-seeded. `SECURITY.md` now says it plainly:

> The deterministic password therefore guards freshly created key material, not just storage the debug config pre-seeded

and pairs it with the actionable instruction (export **before** the first `setup`/`run`; if storage already exists under the default, `run --reset` to wipe and re-seed). `README.md`, `templates/default/README.md` and the `setup` bullet all carry the same framing. That is the correct call — the alternative of quietly reusing the default password without saying so is exactly the kind of thing that ends up in a postmortem.

Two supporting properties I verified rather than assumed, since the fallback changes what lives in the wallet home:

1. **The password never reaches the process table.** `run_with_stdin` (`process.rs:585`) pipes it to the child's stdin; `should_echo()` renders only the command line, not stdin, so `-q`/verbose modes cannot leak it either.
2. **The new key material stays out of support bundles.** The fallback means `.scaffold/wallet/` now gains a real `storage.json`, where previously it often held only `wallet_config.json`. `report.rs:159` still excludes `.scaffold/wallet/**` by policy ("wallet key material is never included") and `report.rs:163` excludes `.scaffold/state/wallet.state`. `report`'s own wallet probe is `--version` only, with no wallet-home env attached, so it cannot enumerate accounts into the archive. The `report` hygiene story survives this change intact — worth stating explicitly because it would have been an easy thing to break silently.

The new `D4` failure signal is a good operational companion to this: *scaffold itself only ever puts `wallet_config.json` into `.scaffold/wallet`, so if `ls .scaffold/wallet` still shows that file alone after a wallet command that should have created storage, the wallet CLI wrote it somewhere else.* That gives a dogfooder a positive check for a failure whose defining property is that it is silent and exits 0.

### `691e811` — the asymmetry is the useful part

The two half-exports failing differently is the detail that makes this documentation instead of a warning, and it is now stated consistently in `D6`, `L4`, both template skills, and the `lgs-cli` troubleshooting table:

- v0.2.0 + only `NSSA_*` → falls back to `~/.lee/wallet`, **no error**, runner sees an empty wallet.
- pre-v0.2.0 + only `LEE_*` → no wallet home at all, loud failure in `WalletCore::from_env()`.

"Only the first shape is silent — treat an empty-looking wallet as the signal to check" is the sentence that will actually save someone an afternoon.

### My §3 finding is resolved

The `Change → Scenario` map now routes this class of change:

> Changes to the wallet-home env contract (`WALLET_HOME_ENV_VARS`, `set_wallet_home_env`, or any new upstream rename), to default-wallet seeding (`ensure_default_wallet_seeded` and its storage-initialization fallback), or to the LEZ pin across the v0.2.0 boundary: rerun `D1`, `D4`, `D6`, and `D7`

Naming the symbols rather than describing the area is the right choice — it is checkable against a diff.

### I exercised the new seeding path against the real wallet, and it does what you say

The unit test drives a fake `sh` script. That pins the parsing and the env contract, but not the thing that actually has to be true: that the **real** LEZ wallet, run against a config with no `initial_accounts`, creates storage and prints something this parser can read. So I reproduced both halves on a real pin.

Setup: second generated project, pinned LEZ `cf3639d8`, with `.scaffold/wallet/wallet_config.json` pre-placed as that pin's debug config with `initial_accounts` deleted — i.e. a v0.2.0-shaped config on a v0.1.x pin. (`prepare_wallet_home` only copies the config when absent, so the stripped file survives `setup`.)

**Before — `master@9fcc376`, reproducing #240 exactly:**

```
warning: could not seed default wallet automatically (no preconfigured public account found)
setup complete
$ cat .scaffold/state/wallet.state   → <ABSENT>
$ ls .scaffold/wallet                → wallet_config.json
```

No `default_address=`, which is precisely the state that makes `cmd_wallet_topup_inner` — and therefore `lgs run` — fail.

**After — this branch, same project, same stripped config:**

```
$ .../target/release/wallet account list
default wallet seeded by initializing wallet storage (config ships no preconfigured account)
  Address: Public/EkA2CTdiiPw6b2syzMxBZcZ33n4W5RCbWH3DDsonWy6w
  State file: /home/user/dogfood-runs/dogfood-246-fallback/.scaffold/state/wallet.state
setup complete

$ cat .scaffold/state/wallet.state
default_address=Public/EkA2CTdiiPw6b2syzMxBZcZ33n4W5RCbWH3DDsonWy6w
$ ls .scaffold/wallet
storage.json  wallet_config.json
$ ls ~/.lee → does not exist
```

Exactly the acceptance shape from #240, produced by a plain `lgs setup` with no bridge script.

### The two-tier scan is load-bearing — here is the proof

This is the part I would not have been confident about from the unit test, because the fake wallet's output is *constructed* to have the marked line. Here is what the real wallet printed in that project:

```
Preconfigured Public/6iArKUXxhUJqS7kCaPNhwMWt3ro71PDyBj7jwAyE2VQV
Preconfigured Public/7wHg9sbJwc6h3NP1S9bekfAzB8CHifEcxKswCKUt3YQo
Preconfigured Private/5ya25h4Xc9GAmrGB2WrTEnEWtQKJwRwQx3Xfo2tucNcE
Preconfigured Private/E8HwiTyQe4H9HK7icTvn95HQMnzx49mP9A2ddtMLpNaN
/ Public/EkA2CTdiiPw6b2syzMxBZcZ33n4W5RCbWH3DDsonWy6w
/ Private/4vgGxgZ5vkP99trgfVDLaF7o5NdHjnJUh26BGRUqCmRD
```

Two things fall out of this that are worth recording in the PR:

1. **A naive "first `Public/` token" scan would have picked the wrong account.** It would have adopted `6iArKUXxh…`, four lines above the one you actually want. The seeded address is `EkA2CTdi…` — the `/ `-prefixed entry, i.e. the account the wallet just created in *its own persistent storage*. So the marked-line tier is not defensive garnish; on real output it is the difference between a correct default and a wrong one. Good call making it tier one rather than a fallback.
2. **Even with `initial_accounts` removed from the config, the wallet still prints `Preconfigured` entries** (from its own built-ins — note they are different IDs than the ones the unmodified config produces). So "the config ships no preconfigured account" does **not** imply "the listing contains no preconfigured account". Your code handles this correctly; only the docs undersell it, which is my one suggestion below.

### Docs nit that this turned up

`DOGFOODING.md` describes the skipped candidates as *"`Public/` addresses printed anywhere else (a banner or status line)"*. On the real wallet they are neither — they are four `Preconfigured Public/…` / `Preconfigured Private/…` entries, which a reader could easily assume *are* the listing. Since the whole point of tier one is to skip exactly those, I would name them:

> …ignoring `Public/` addresses on lines that are not `/ `-prefixed — notably the wallet's own `Preconfigured …` entries, which it prints above the stored accounts even when the config ships no `initial_accounts`.

That also makes the fallback tier's risk legible: if the wallet ever stops `/ `-prefixing stored accounts, the fallback will silently start adopting a preconfigured address.

### The rest of the affected scenarios, run live

Full `D1` on the pinned LEZ with a real sequencer, in a separate project with the **unmodified** config, to confirm the preconfigured path is untouched:

- `setup` → `default wallet seeded from preconfigured account` / `Address: Public/CbgR6tj5…` — the pre-existing branch, unchanged.
- `localnet start` → `localnet ready (sequencer pid=19408)`; `build` → `Building guest methods...`; `deploy` → **5 succeeded, 0 failed**, paced one program per block; `wallet topup` (no address, using the seeded default) → `wallet topup complete`; `wallet -- check-health` → `✅All looks good!`.
- `~/.lee` does not exist after any of it.

`D4` / `D7` slices for the env contract, with this branch's binary:

```
$ lgs wallet topup --dry-run
dry-run: wallet topup command will not be executed
NSSA_WALLET_HOME_DIR=/home/user/dogfood-runs/dogfood-default/.scaffold/wallet
LEE_WALLET_HOME_DIR=/home/user/dogfood-runs/dogfood-default/.scaffold/wallet

$ lgs run --post-deploy 'echo "NSSA=[$NSSA_WALLET_HOME_DIR]"; echo "LEE=[$LEE_WALLET_HOME_DIR]"'
NSSA=[/home/user/dogfood-runs/dogfood-default/.scaffold/wallet]
LEE=[/home/user/dogfood-runs/dogfood-default/.scaffold/wallet]
```

Both names, same path, on both a wallet subprocess and a post-deploy hook — which is the D7 failure signal you added ("must both be set and must print the same path"), satisfied.

### Two things I would still change

Neither blocks.

1. **`check_standalone_support`'s nested probe is untested.** `doctor_checks.rs:299` adds `lez/sequencer/service/Cargo.toml` to the probe list. I confirmed the addition is safe — the loop returns `Pass` on the *first* file containing `standalone`, so adding a path is purely additive and cannot break the flat layout. But it is the only production change in this PR with no test, and it is the remainder of #234's layout work, i.e. the exact thing that was missed once already. A two-case test (flat marker found, nested marker found) would close it.

2. **Pre-existing, not yours — a misplaced doc comment in `run.rs`.** The block documenting `reseed_after_wipe` ("Re-seed the project's default wallet after `cmd_localnet_reset` wiped it…") sits immediately above `fn watch_loop` (`run.rs:298-303`), while `reseed_after_wipe` itself is at `run.rs:521` with no doc comment. It is on `master@9fcc376` already, so this PR did not introduce it — flagging it because this PR is the one touching `reseed_after_wipe`'s signature, so it is a cheap drive-by if you are in the file.

### Note on merge order with #245

Worth knowing before either is merged: **#246 and #245 each merge cleanly into `master`, but conflict with each other.** One hunk, in `src/commands/run.rs` — the `hook_receives_wallet_home_dir_env` test, where this PR rewrites the hook to `printf` both env var names and #245 adds a `topup_skipped` argument to `run_post_deploy_hook`. The resolution is to take both (dual-name `printf` **and** the trailing `false`); the failure mode to avoid is resolving toward #245's side and silently dropping the `LEE_WALLET_HOME_DIR` assertion, which would leave this PR's central invariant untested at the hook boundary.

I verified the combination works: merging both onto `master@9fcc376` and resolving that hunk by taking both sides gives **731 green, 0 failed**, including `hook_receives_wallet_home_dir_env` and `set_wallet_home_env_sets_every_wallet_home_var`. So this is textual only — no semantic interaction between the two PRs.

---

danisharora099:
@weboko thanks for the re-review. Since the remaining suggestions are non-blocking, could you submit a formal approval when you get a moment?

weboko:
@danisharora099 these a low hanging things and I think AI is perfect for addressing them in no time

weboko (approved):

PR #247, opened 2026-08-10
fix(deploy): drain the pacing stub socket before close

danisharora099:
This PR closes the pre-release issues found while running and reviewing the D-series and Basecamp dogfooding scenarios.

## 1. Platform-fragile pacing test stub

`commands::deploy::tests::read_head_with_retry_recovers_after_transient_failure`, added in #241, failed intermittently on macOS — **3 failures in 12 local runs**.

The product code is fine. The bug was in the scripted HTTP server: a single read did not guarantee the complete request had arrived, so dropping a socket with unread inbound bytes could send RST and discard response bytes already queued for the client.

The deploy and wallet-support stubs now share one test-only helper that reads through the declared HTTP request body before writing a response. This removes the RST window deterministically and gives the initial read a timeout. The helper documents its intentional `Content-Length` framing assumption: the current `ureq` callers send fixed JSON bodies, while `Transfer-Encoding: chunked` is not decoded. The failure reproduced on macOS but not under the Linux CI conditions tested during review, so this is platform hardening rather than a demonstrated release-CI blocker.

## 2. Basecamp guidance synchronized with the current interface

The dogfooding runbook and embedded compatibility document now list the eleven real `basecamp` verbs in `--help` order:

`setup`, `modules`, `install`, `launch`, `develop`, `build`, `build-portable`, `run`, `doctor`, `paths`, `docs`.

The shipped `basecamp` and `lgs-cli` agent skills now cover the same command surface and teach the current top-level `[modules.<name>]` schema. Runtime remediation messages and current-behavior comments use that schema too. Intentional references to the pre-0.2.0 `[basecamp.modules.<name>]` layout remain in migration documentation, migration tests, and adversarial fixtures.

Regression coverage now derives the Basecamp verb set from clap, requires both shipped skills to cover every live subcommand, and compares the embedded `basecamp docs` canonical verb list exactly and in order. Missing, reordered, and phantom verbs are rejected; applicable shipped guidance also cannot reintroduce the legacy schema outside the intentional migration note.

## 3. Guest artifact paths

Current guidance consistently identifies `target/riscv-guest/...` as the default template output. `methods/target/...` remains documented and tested as a supported sub-crate compatibility layout, matching deploy's two search roots.

## Not included

`Cargo.toml` remains on `0.1.1`. The release version is deliberately left to the release owner.

## Verification

- Full suite: **544 + 182 + 3 + 5 passed, 0 failed**.
- `cargo fmt --check` clean.
- `cargo clippy --all-targets --all-features` completed with only pre-existing warnings and none on changed lines.
- `read_head_with_retry_recovers_after_transient_failure`: **200 consecutive runs passed**.
- Built `basecamp --help` emits exactly the documented eleven verbs in order.
- Built `basecamp docs` emits the current command list and `[modules.<name>]` schema.
- D1 was previously executed end to end against this branch's parent: `new` → `create` → `setup` → `localnet start` → `build` → `deploy` → `wallet topup` → `wallet -- check-health`; all commands exited 0 and all 5 programs deployed with zero failures.


weboko:
## Review — both changes verified; the socket fix is correct and the verb list now matches `--help` exactly

Reviewed on a fresh Linux x86_64 container provisioned per `DOGFOODING.md` (Rust 1.94.1, Nix 2.35.1 with flakes, r0vm 3.0.5 matched to the pinned LEZ `cf3639d8`'s `risc0-zkvm = 3.0.5`, risc0 guest toolchain 1.91.1, risc0 C++ toolchain `v2024.1.5`). Branch is based directly on current `master` (`9fcc376`), no merge needed.

### 1. The pacing-stub fix — mechanism confirmed, and it can't regress the client

`cargo test` on the branch: **714 green** (524 lib + 182 CLI + 3 testnode + 5 doc), zero failures. `cargo fmt --check` clean.

I could not reproduce the flake on Linux — **0 failures in 40 runs** of `read_head_with_retry_recovers_after_transient_failure` on `master@9fcc376`, and **0 failures in 60 runs** of the whole `commands::deploy::tests::` module on this branch. That is consistent with your report rather than in tension with it: you measured 3/12 on macOS, and whether a close with unread inbound data emits RST or FIN is exactly the kind of thing that differs between the two stacks. On loopback Linux the `ureq` request lands in the single 4096-byte `read` essentially every time, so there is nothing left unread to trigger the reset. The mechanism you describe is real regardless of which host observes it, and the fix is the right one.

The thing I did want to rule out is the drain becoming a *new* source of flakiness, since it now holds each connection open for up to 100 ms:

- **It cannot swallow a subsequent request.** `rpc_get_last_block_id` (`wallet_support.rs:407`) builds a **fresh `ureq::Agent` per call**, so there is no connection pooling across polls and no keep-alive reuse for the drain to consume. Had the agent been shared, the second scripted request could have arrived on the stream the server was draining and been discarded — worth stating explicitly, because it makes this fix quietly dependent on that per-call construction.
- **It cannot hang.** The loop exits on `Ok(0)` (EOF) and on the `Err` from the 100 ms read timeout, so it is bounded either way.
- **It costs nothing in practice.** 60 runs of the deploy module took 13 s total, ~0.21 s per run — the client closes promptly, so the drain hits EOF rather than the timeout.

The comment explaining the RST-vs-FIN reasoning is the right length and will save the next person the packet-level detour.

### 2. The B1 verb list — matches exactly

Verified against the binary rather than the source. `basecamp --help` on the freshly built branch lists, in order:

```
setup, modules, install, launch, develop, build, build-portable, run, doctor, paths, docs
```

which is the new line character-for-character, including ordering. The old line was missing `develop`, `build` and `run` and had `paths` in a different position, so this closes the contradiction with B5/B6 that you flagged.

### On the version bump you left out

Agreed with leaving it, and agreed with calling it out. `publish-crate.yml`'s tag-match check firing is the most likely explanation for `v0.2.0` being tagged in git but absent from crates.io. Worth noting for whoever owns the release: this branch does not unblock that on its own.

### Verdict

No changes requested. Test-only plus a doc correction, no production code touched, both claims independently verified. This is a clean pre-release pickup.

---

weboko:
## One more stale runbook line, found while dogfooding this branch — `D1`'s guest-artifact path

Not a request to expand this PR's scope; flagging it here because this PR is the one correcting stale `DOGFOODING.md` text, so it may be cheapest to fold in.

I ran `D1` end to end on Linux x86_64 against this branch's parent (`master@9fcc376`) with the full real toolchain — real `sequencer_service` from the pinned LEZ `cf3639d8`, r0vm 3.0.5, risc0 guest toolchain 1.91.1 — and everything passed: `localnet ready (sequencer pid=…)`, `Building guest methods...`, deploy **5 succeeded / 0 failed** with the documented one-program-per-block pacing, `wallet topup` off the seeded default, `wallet -- check-health` → `✅All looks good!`.

One documented success signal did not match. `D1` says `build`:

> also prints `Building guest methods...` and produces a `methods/target/.../release` artifact

The first half is right; the second is not. There is no `methods/target` directory at all after a successful build — `risc0-build` puts the guest ELFs under the **project** target dir:

```
$ ls -d methods/target
→ does not exist

$ ls target/riscv-guest/example_program_deployment_methods/example_program_deployment_programs/riscv32im-risc0-zkvm-elf/release/*.bin
hello_world.bin
hello_world_with_authorization.bin
hello_world_with_move_function.bin
simple_tail_call.bin
tail_call_with_pda.bin
```

`deploy` agrees — it submitted from exactly those paths:

```
$ .../wallet deploy-program /home/user/dogfood-runs/dogfood-default/target/riscv-guest/example_program_deployment_methods/example_program_deployment_programs/riscv32im-risc0-zkvm-elf/release/hello_world.bin
OK  hello_world submitted
```

So the product is fine and the runbook is stale, the same shape of problem as the `basecamp` verb list you already fixed here: a dogfooder checking the documented signal finds nothing at the named path and has to decide whether they are looking at a regression. Suggested replacement:

> …also prints `Building guest methods...` and produces guest ELFs under `target/riscv-guest/<methods-crate>/<guest-crate>/riscv32im-risc0-zkvm-elf/release/*.bin` — the same paths `deploy` submits from.

Anchoring it to "the paths `deploy` submits from" also makes the line self-checking against the deploy output right below it.

---

danisharora099:
@weboko agreed—the D1 guest-artifact path is stale. I’ll fold that correction into this PR and ping you for approval once it is pushed.

weboko:
ok, lmk

src/commands/deploy.rs
```diff
@@ -887,6 +887,17 @@ mod tests {
...
                 let _ = stream.write_all(payload.as_bytes());
+                let _ = stream.flush();
+                // The single `read` above can return before the client has
+                // finished sending its request body. Closing a socket that
+                // still has unread inbound data makes the kernel send RST
+                // instead of FIN, and the RST discards the response we just
+                // wrote — the client then sees a truncated read rather than
+                // the scripted answer. Drain to EOF (bounded by a short
+                // timeout) so every connection closes gracefully.
+                let _ = stream.set_read_timeout(Some(Duration::from_millis(100)));
```
weboko: Non-blocking, but worth considering: the 100 ms bound makes this a strong mitigation rather than a guarantee. If a client is slow to close — exactly the loaded-CI condition where this flake bites — the drain gives up, the socket is dropped with inbound data still queued, and you are back to the RST you are fixing here.

Raising the bound is close to free. In the happy path the loop exits on `Ok(0)` as soon as the client closes, which happens promptly because `rpc_get_last_block_id` builds and drops a fresh `ureq::Agent` per call, so the pooled connection dies with it. The timeout only ever elapses in the pathological case, and the enclosing tests already carry a 10 s deadline of their own.

```suggestion
                let _ = stream.set_read_timeout(Some(Duration::from_secs(1)));
```

Optional companion: a `stream.shutdown(std::net::Shutdown::Write)` right after the `flush()` sends FIN immediately, so the client sees a complete, properly terminated response before the drain even begins, instead of only when the stub finally drops the socket.

---

DOGFOODING.md
```diff
@@ -1041,7 +1041,7 @@ ls .scaffold/basecamp/profiles
 
 ### Expected Success Signals
 
-- `basecamp --help` lists `setup`, `modules`, `install`, `launch`, `paths`, `build-portable`, `doctor`, and `docs`.
+- `basecamp --help` lists `setup`, `modules`, `install`, `launch`, `develop`, `build`, `build-portable`, `run`, `doctor`, `paths`, and `docs`.
```
weboko, line 1062: This line is now correct, but the same staleness you are fixing here exists in `docs/basecamp-module-requirements.md:3` — and it is the more user-visible copy, because `basecamp docs` prints it verbatim:

> This is the contract between a module project and `logos-scaffold basecamp {setup,install,launch,reset,build-portable}`.

That list is wrong in both directions. It advertises **`reset`, which is not a verb at all**:

```
$ logos-scaffold basecamp reset --help
error: unrecognized subcommand 'reset'
```

and it omits `modules`, `develop`, `build`, `run`, `doctor`, `paths`, and `docs`.

Your rationale for the B1 fix applies with more force here: a stale runbook makes someone file a spurious regression, but a doc that names a command which errors out sends them down a dead end from the tool's own help output. Since B1 asserts on `basecamp docs` output two lines below this one, the two are worth keeping in sync.

Entirely your call whether to fold it into this PR or split it — flagging it because it is the same defect class and one line to fix.

---

weboko (commented):
## Review — verified hands-on, recommend merge

Reviewed the diff and reproduced the verification independently on Linux (Ubuntu 24.04, Rust 1.94.1). Both changes are correct and the claims in the description hold up.

### Test-suite verification

| Check | Result |
| --- | --- |
| `cargo fmt --check` | clean |
| Full suite | **524 + 182 + 3 + 5 passed, 0 failed** — matches the description exactly |
| `read_head_with_retry_recovers_after_transient_failure` × 30 on this branch | **0 failures** |

### On the flake itself

I could not reproduce the failure on the base commit (`9fcc376`) on Linux: **0 failures in 30 runs**, and still 0 in a further 40 runs with all cores saturated by busy-loops to skew the scheduling. That is not evidence against the fix — it is consistent with the diagnosis. Whether a close with unread inbound data emits RST or FIN, and how aggressively a queued RST discards already-delivered receive-buffer data, is stack-specific; Darwin is materially more eager here than Linux loopback. Your 3-in-12 → 0-in-30 on macOS is the signal that matters. Recording the negative result so nobody later reads a green Linux run as proof the flake was imaginary.

The diagnosis is sound on the merits regardless of reproduction: `read` returning once does not mean the request is fully drained, and the RST-discards-pending-data behaviour is exactly what produces the two symptoms you describe (the `ureq` "failed to read exact buffer length" panic and the silent 10s timeout).

One thing I checked specifically, because it would have broken the helper's one-connection-per-response model: `rpc_get_last_block_id` builds a **fresh `ureq::Agent` per call** (`wallet_support.rs:407`), so there is no pooled-connection reuse across polls, and holding the socket open ~100ms longer cannot cause the client to reuse a connection the stub has already retired. The change is safe on that axis.

Scope is right: the change is entirely inside `#[cfg(test)]`, so there is no product-behaviour risk, and your read that this should land before the release is correct — `publish-crate.yml`'s `verify` job runs `cargo test --locked`.

### Non-blocking notes

Two comments inline: one on the drain timeout bound, one unrelated finding on the same class of staleness you fixed in B1.

### B1 doc fix — verified accurate

Built the binary and diffed the runbook line against real output. `basecamp --help` lists exactly the eleven verbs, **in the order the new line states them**:

```
setup, modules, install, launch, develop, build, build-portable, run, doctor, paths, docs
```

The old eight-verb list was stale as described. Also confirmed the surrounding B1 signals that do not need a Nix basecamp build: `init` → `basecamp docs` prints the compatibility rules, and running `basecamp doctor` outside a project fails with the documented "must be run inside a logos-scaffold project" message.

---

weboko:
## Follow-up: ran D1 end to end, and found a third stale expected-signal in the same runbook

Provisioned a real LEZ toolchain on Linux (Nix + risc0 rust/cpp + source-built `sequencer_service` and `wallet` at pin `cf3639d8`) and executed **D1 end to end** against this branch. All steps exit 0 and every documented signal matched — `new`/`create`, `setup` (wallet seeded), `localnet start` ("localnet ready"), `build`, `deploy` (**5 succeeded / 0 failed**), `wallet topup` (preflight → `auth-transfer init` → `pinata claim`), `wallet -- check-health` ("✅All looks good!").

Relevant to this PR specifically: the deploy leg exercised the real pacing path your test covers, and it behaved as documented —

```
Waiting for a new block past 24 before the next deployment (one program per block; bedrock inscription size cap)...
```

five times, zero failures. So the production code the flaky test guards is healthy on a live sequencer; the flake really was confined to the stub.

### The finding: `DOGFOODING.md:237` is stale in the same way B1 was

D1's `build` expected-signal says:

> …also prints `Building guest methods...` and produces a **`methods/target/.../release`** artifact.

`Building guest methods...` is right. The path is not — and `methods/target/` **does not exist at all** after a successful build:

```
$ ls methods/
Cargo.toml  build.rs  guest  src
```

risc0-build 3.x writes the guest ELF into the workspace `target/riscv-guest/` tree instead:

```
$ file target/riscv-guest/example_program_deployment_methods/\
example_program_deployment_programs/riscv32im-risc0-zkvm-elf/release/hello_world_with_move_function
ELF 32-bit LSB executable, UCB RISC-V, soft-float ABI, version 1 (SYSV), statically linked
```

This is the same defect class you fixed in B1, and arguably worse in effect: a dogfooder checking D1 looks for a directory that a *correct* build never creates, so the natural conclusion is that the guest build silently produced nothing. The B1 list at least described commands that exist.

Suggested replacement for that clause:

> …also prints `Building guest methods...` and produces a guest ELF under `target/riscv-guest/<methods-pkg>/<guest-pkg>/riscv32im-risc0-zkvm-elf/release/` (risc0-build writes into the workspace `target/` tree, not `methods/target/`).

Same call as before on whether to fold it in here or split it — flagging it because you are already in this file fixing exactly this, and because I only found it by running the step rather than reading it.

---

danisharora099:
Addressed all three remaining review points:

- reduced the test stub timeout to 1 second and half-closed the response before draining the request
- corrected the Basecamp command list
- corrected the D1 build output path

Also rebased onto current `master`. Both CI runs are green and all three commits are verified. @weboko please approve when you get a chance.

src/commands/deploy.rs
```diff
@@ -877,6 +877,22 @@ mod tests {
...
+                // still has unread inbound data makes the kernel send RST
+                // instead of FIN, and the RST discards the response we just
+                // wrote — the client then sees a truncated read rather than
+                // the scripted answer. Drain to EOF with a one-second ceiling;
+                // the happy path exits as soon as the client closes, while
+                // loaded CI gets enough time to avoid turning this back into
+                // a reset.
+                let _ = stream.set_read_timeout(Some(Duration::from_secs(1)));
+                let mut sink = [0_u8; 1024];
+                while matches!(stream.read(&mut sink), Ok(n) if n > 0) {}
```
weboko: The fix is correct and the flake it targets is real. But the helper this one documents itself as mirroring already solves the same problem, and solves it structurally rather than by compensation.

`wallet_support::tests::drain_http_request` (`src/commands/wallet_support.rs:568`) reads the request **to completion before writing the response**, using `http_request_complete` (`:594`) to detect the end from `\r\n\r\n` plus `content-length`. Because the request is fully consumed, there is never unread inbound data on the socket at close, so the RST window this PR is closing never opens in the first place.

This PR keeps the single partial `read` on line 866 and compensates *after* the response with a bounded post-hoc drain. That works, but it is the weaker shape in three ways:

1. **Exit condition is a race, not a fact.** The drain stops on EOF or a 1s timeout. `drain_http_request` stops when the request is *known complete* — deterministic, and independent of how the client schedules its writes.
2. **It parks the accept loop.** The stub serves connections sequentially (`for response in responses` → `accept` → handle), so the drain of connection N delays `accept` of N+1. Raising the bound 100ms → 1s raises the worst case from 0.3s to 3s against the tests' own 10s deadline. Happy path still exits immediately on EOF, so this is a margin question, not a live failure — but the margin got 10× thinner.
3. **The first `read` on line 866 has no timeout at all.** A client that connects and never sends blocks this thread forever. `drain_http_request` sets a 2s read timeout up front, so reusing it fixes that too.

There's also a consistency cost: the doc comment on line 851 still says this "Mirrors the single-shot helper in `wallet_support::tests`", but after this change it diverges from that helper in exactly the aspect the comment is pointing at. Two different answers to one problem in the same crate's test code is the thing that bites the next person.

Suggest lifting `drain_http_request` / `http_request_complete` to a shared test helper — `pub(crate)` on the `wallet_support::tests` items, or a small `test_support` module — and calling it before the write instead of the read/compensate pair:

```rust
                drain_http_request(&mut stream);
```

Non-blocking: what's here does fix the flake, and I verified it does (0 failures in 50 runs). This is about which of the two shapes the codebase keeps.

---

weboko (commented):
Re-reviewed after the three new commits. **Both points from my previous review are addressed**, and I verified the doc claims against a real binary rather than by reading:

- Drain bound raised 100ms → 1s, and the optional `shutdown(Shutdown::Write)` companion was taken. ✅
- `docs/basecamp-module-requirements.md:3` rewritten — the phantom `reset` verb is gone and all eleven real verbs are listed. ✅

## What I verified

Built this branch and checked the documentation changes against actual output, since this doc is embedded via `include_str!` (`src/commands/basecamp.rs:143`) and printed verbatim by `basecamp docs`:

- `basecamp --help` emits exactly the eleven verbs, in exactly the order both docs now claim: `setup, modules, install, launch, develop, build, build-portable, run, doctor, paths, docs`. Matches `BasecampSubcommand` in `src/cli.rs:1034-1080`. ✅
- `basecamp docs` prints the corrected contract line. ✅
- The D1 guest-path correction is accurate: `GUEST_BIN_SEARCH_ROOTS` (`src/commands/deploy.rs:28`) is `["target/riscv-guest", "methods/target"]`, and the `target/riscv-guest/<methods-crate>/<guest-crate>/riscv32im-risc0-zkvm-elf/release` shape matches the layout test at `deploy.rs:975`. "the same paths `deploy` submits from" is literally true. ✅
- Full suite green on a fresh toolchain: 543 + 182 + 3 + 5, `cargo fmt --check` clean, no new clippy warnings. ✅
- The target test, 50 consecutive runs on this branch: **0 failures**. ✅

## One calibration on the release-risk framing

The PR argues this should land before the release because `publish-crate.yml`'s `verify` job runs `cargo test --locked`, "so this can fail a release run on a bad roll." I could not reproduce that risk on Linux, and every workflow in `.github/workflows/` is `runs-on: ubuntu-latest`.

On `master` (d403a95), unpatched, the test passed **116/116**: 50 runs idle, 30 with all 4 cores saturated, and 36 more as 12-way concurrent processes. Your repro was macOS, which is consistent with the mechanism — on loopback Linux the ureq request almost certainly lands in a single segment, so the one `read(4096)` consumes it whole and there is no unread data to turn the close into an RST. Splitting headers and body across writes is what opens the window, and that is platform- and timing-dependent.

So: the fix is sound hardening and I'd still take it, but the "this can fail the release" argument is weaker than stated for a Linux-only CI matrix. Worth landing on its merits (correct test, no longer platform-fragile) rather than as a release blocker.

Confirmed separately that the "Not included" note is accurate — `Cargo.toml` is still `version = "0.1.1"`.

One non-blocking code-reuse comment inline.

---

danisharora099:
@weboko final pass is ready on verified signed commit `ba55fb8`.

- All three review threads are addressed and resolved: the stubs share the complete-request drain, both canonical verb lists match the eleven-command CLI surface, and D1 names the real workspace guest-artifact path.
- A proactive consistency audit also updated the shipped Basecamp/lgs-cli skills and runtime remediation hints to the current `[modules.<name>]` schema, while preserving intentional migration-history references. Regression coverage now guards the schema and all eleven verbs.
- Verification: `cargo fmt --check`; full suite **544 + 182 + 3 + 5**; target recovery test **200/200**; both validation jobs and both template E2E variants green.

No version bump or production socket behavior change. Please approve when convenient.

weboko (commented):
Re-reviewed after the three commits pushed since my last pass (`5082de2`, `b3b2d20`, `ba55fb8`). My three earlier comments are all resolved, and I verified the claims rather than taking the PR description on trust. One new gap, non-blocking.

## Earlier comments — resolved

1. **100 ms drain bound was a mitigation, not a guarantee.** Now moot: the deploy stub no longer compensates with a bounded read, it reuses the structural complete-request drain (`drain_http_request`, 2 s safety timeout rather than a 100 ms budget). This is the structural fix, not a bigger number.
2. **`docs/basecamp-module-requirements.md:3` advertised `reset` and omitted seven verbs.** Fixed. `grep -rn "basecamp reset" docs/ skills/ DOGFOODING.md README.md` → no hits, and `basecamp reset --help` still errors as `unrecognized subcommand`, confirming it never existed.
3. **The wallet-support helper already solved this structurally.** Done — extracted to `src/commands/test_support.rs` and shared by both stubs instead of duplicated.

## Verification

Provisioned a clean box (Nix 2.35.1 + flakes, risc0 r0vm 3.0.5 / guest Rust 1.91.1 / C++ 13.2.0, Rust 1.94.1) and ran the scenarios this PR touches.

| Check | Result |
|---|---|
| Full suite | 544 + 182 + 3 + 5 passed, 0 failed — matches the PR body exactly |
| `read_head_with_retry_recovers_after_transient_failure` | **200/200 passed** |
| `cargo fmt --check` | clean |
| `cargo clippy --all-targets --all-features` | 56 warnings, **zero on changed lines** (all pre-existing) |
| `basecamp --help` | exactly the eleven documented verbs, in the documented order |
| `basecamp docs` | byte-identical to `docs/basecamp-module-requirements.md` |
| Legacy schema in shipped binary | 0 occurrences of `[basecamp.modules]`, 27 of `[modules]` |

**D1, end to end, real LEZ sequencer:** `new` → `create` → `setup` (wallet seeded) → `localnet start` (ready) → `build` → `deploy` → `wallet topup` → `wallet -- check-health` (`✅All looks good!`). Deploy submitted **5/5, zero failures**, correctly paced one program per block.

The guest-artifact-path correction is right, and the text it replaced was wrong. `build` produced all five `.bin` files under:

```
```

`methods/target/` is **absent** in the default template, and `deploy` submitted from those exact `target/riscv-guest/...` paths — so "the same paths `deploy` submits from" is accurate.

**E3** passes fully: all three project flavors get the four skills under `.claude/skills/` + `.cursor/rules/`, `AGENTS.md` present, `.mdc` frontmatter carries `description:`/`alwaysApply: false` with no `name:` leak, skills byte-identical to canonical source, re-`init` idempotent (identical shasums), nothing leaked into `.gitignore`.

**DOGFOODING.md** is up to date for this change — B1's verb list now matches the CLI, and B1–B6 already cover `develop` / `build` / `run`.

## One gap

The new regression test guards two of the three shipped surfaces — see the inline comment on `src/template/skills.rs`. Worth closing since it is the file comment 2 above was actually about, but I would not hold the PR on it.

---

src/template/skills.rs
```diff
@@ -213,6 +213,50 @@ mod tests {
...
+                    .chars()
+                    .next()
+                    .is_some_and(|c| c.is_ascii_alphanumeric() || c == '-')
+            })
+        }
+
+        let skills = [
+            ("basecamp", include_str!("../../skills/basecamp/SKILL.md")),
+            ("lgs-cli", include_str!("../../skills/lgs-cli/SKILL.md")),
+        ];
```
weboko, line 231: **Regression coverage misses the surface that motivated the fix.** Non-blocking, but cheap to close.

This test guards `skills/basecamp/SKILL.md` and `skills/lgs-cli/SKILL.md`. The verb list also ships in `docs/basecamp-module-requirements.md`, which `basecamp docs` prints verbatim via `include_str!` (`src/commands/basecamp.rs:143`) — the *more* user-visible copy, and the one that carried the phantom `reset` verb this PR removes.

That third copy is unguarded. I verified it by reintroducing exactly the drift this PR fixes:

```
$ sed -i '3s/`launch`, `develop`, `build`, `build-portable`, `run`, `doctor`/`launch`, `build`, `build-portable`, `doctor`/' \
    docs/basecamp-module-requirements.md
$ cargo test --release
test result: ok. 544 passed; 0 failed
test result: ok. 182 passed; 0 failed
test result: ok. 3 passed; 0 failed
test result: ok. 5 passed; 0 failed
```

Dropping `develop` and `run` from the shipped `basecamp docs` output passes the entire suite.

Second, `verbs` is a hardcoded literal, so it only catches a *dropped* verb — adding a twelfth subcommand leaves every shipped surface silently stale. Deriving from clap fixes both, and `cli_command()` (`src/cli.rs:1593`) already exposes what's needed.

I prototyped this and confirmed it (a) passes clean, (b) fails with `basecamp docs is missing the \`develop\` command` on the drift above, and (c) still fails with `lgs-cli is missing the \`basecamp paths\` command` when a skill drops a verb:

```suggestion
        let skills = [
            ("basecamp", include_str!("../../skills/basecamp/SKILL.md")),
            ("lgs-cli", include_str!("../../skills/lgs-cli/SKILL.md")),
        ];
        // `basecamp docs` ships this file verbatim, so it is a third surface
        // that can drift out of sync with the CLI.
        let embedded_docs = include_str!("../../docs/basecamp-module-requirements.md");

        // Derive the verb set from clap so that adding a subcommand fails here
        // until every shipped surface documents it.
        let command = crate::cli::cli_command();
        let basecamp = command
            .get_subcommands()
            .find(|c| c.get_name() == "basecamp")
            .expect("basecamp subcommand");
        let verbs: Vec<&str> = basecamp.get_subcommands().map(|c| c.get_name()).collect();
```

Then keep the existing `[modules]`/verb loop over `skills`, and add for the embedded doc — note it lists verbs as a backticked prose list, not as `basecamp <verb>` invocations, so `contains_command` does not apply:

```rust
// The embedded doc still names the pre-0.2.0 layout in its migration
// note, so it only carries the positive schema assertion.
assert!(
    embedded_docs.contains("[modules"),
    "basecamp docs must teach the current top-level [modules] schema"
);
for verb in &verbs {
    assert!(
        embedded_docs.contains(&format!("`{verb}`")),
        "basecamp docs is missing the `{verb}` command"
    );
}
```

---

src/commands/test_support.rs
```diff
@@ -0,0 +1,47 @@
+use std::io::Read;
+use std::net::TcpStream;
+use std::time::Duration;
+
+/// Read one HTTP request through its declared body before a test stub responds.
+///
+/// Consuming the complete request prevents the stub from closing a socket with
+/// unread inbound data, which can turn the close into an RST and discard a
+/// response that was already written.
```
weboko, line 9: Minor, non-blocking: this treats a missing `Content-Length` as "body length 0", so a request using `Transfer-Encoding: chunked` is declared complete as soon as the headers land. The body then stays unread, the socket is dropped with inbound data queued, and the RST window this helper exists to close reopens — silently, as a flake rather than a clear failure.

Not reachable today: both stubs are driven by `ureq` with a fixed JSON body, which always sets `Content-Length`. It is a latent trap for whoever changes the client next, so a one-line note on the assumption would carry its weight:

```suggestion
/// Read one HTTP request through its declared body before a test stub responds.
///
/// Consuming the complete request prevents the stub from closing a socket with
/// unread inbound data, which can turn the close into an RST and discard a
/// response that was already written.
///
/// Assumes a `Content-Length`-delimited request (what `ureq` sends for a fixed
/// body). `Transfer-Encoding: chunked` is not decoded — a chunked request would
/// be treated as header-only and its body left unread.
```

---

src/template/skills.rs
danisharora099: Addressed in verified signed commit `875c5d9`: the test now derives Basecamp verbs from clap and checks the embedded `basecamp docs` source alongside both shipped skills, so both missing verbs and newly added subcommands cause a regression failure.

src/commands/test_support.rs
danisharora099: Addressed in verified signed commit `875c5d9`: the helper now documents that it supports the `Content-Length` framing used by the current fixed-body `ureq` callers and does not decode `Transfer-Encoding: chunked`.

danisharora099:
@weboko both non-blocking follow-ups are closed on verified signed commit `875c5d9`:

- Basecamp regression coverage now derives verbs from clap and checks the embedded `basecamp docs` source plus both shipped skills.
- The shared HTTP drain helper now documents its intentional `Content-Length` framing boundary and lack of chunked decoding.

Local verification is green: `cargo fmt --check`, focused regression, full suite **544 + 182 + 3 + 5**, changed-file clippy audit, and socket recovery **200/200**. Both threads are resolved; no production behavior or release version changed. Please approve if satisfied.

src/template/skills.rs
```diff
@@ -213,6 +213,65 @@ mod tests {
...
+        assert!(
+            embedded_docs.contains("[modules"),
+            "basecamp docs must teach the current top-level [modules] schema"
+        );
+        for verb in &verbs {
+            assert!(
+                embedded_docs.contains(&format!("`{verb}`")),
+                "basecamp docs is missing the `{verb}` command"
+            );
+        }
```
weboko: **The two surfaces this test guards are not guarded equally — and the weaker half is the one that motivated the fix.** Non-blocking, but the fix is small and I verified it.

The skills half is strong. `contains_command` (`:218`) pins an actual command mention and correctly rejects prefix collisions. I mutation-tested it: renaming `basecamp develop` in `skills/basecamp/SKILL.md` fails with *"basecamp is missing the `basecamp develop` command"*, and reintroducing `[basecamp.modules]` fails with *"basecamp must teach the current top-level [modules] schema"*. Exactly right.

The `embedded_docs` half is close to vacuous, and that is the copy `basecamp docs` prints verbatim — the one that carried the phantom `reset` verb.

**1. The verb assertion (`:267-272`) matches incidental prose.** `` `launch` `` occurs 18 times in the document, `` `build-portable` `` 12, `` `install` `` 5 — almost all in body text, not in the canonical list. So the assertion passes even when the verb is deleted from the list. Verified on this branch: I removed `install`, `launch` and `build-portable` from the line 3 list — the exact sentence that was stale — and the test still passed.

**2. The schema assertion (`:263-266`) is satisfied by the migration note alone.** `contains("[modules")` is true because line 39 says *"the section has moved to the top-level `[modules.<name>]` namespace"*. Verified: I reverted every other `[modules.` in the document back to `[basecamp.modules.` (7 occurrences, leaving line 39 untouched) and the test still passed — a document that teaches the legacy schema everywhere it matters is green.

Anchoring the verb check to the canonical list line and excluding the migration note from the schema check closes both. I ran this against the clean tree (passes) and against both mutations above (fails with the intended message):

```suggestion
        // The migration note intentionally names the pre-0.2.0 layout, so the
        // schema assertion is made against the rest of the document.
        let schema_body = embedded_docs
            .lines()
            .filter(|line| !line.contains("The pre-0.2.0 schema used"))
            .collect::<Vec<_>>()
            .join("\n");
        assert!(
            schema_body.contains("[modules.") && !schema_body.contains("[basecamp.modules"),
            "basecamp docs must teach the current top-level [modules] schema outside its migration note"
        );
        // The canonical verb list is the sentence the document opens with —
        // the copy that carried the stale verb set — so assert against that
        // line rather than the file at large.
        let verb_list = embedded_docs
            .lines()
            .find(|line| line.contains("`logos-scaffold basecamp` provides"))
            .expect("basecamp docs must carry the canonical verb list");
        for verb in &verbs {
            assert!(
                verb_list.contains(&format!("`{verb}`")),
                "basecamp docs verb list is missing `{verb}`"
            );
        }
```

---

src/commands/test_support.rs
```diff
@@ -0,0 +1,51 @@
...
+            }
+            Err(err)
+                if matches!(
+                    err.kind(),
+                    std::io::ErrorKind::WouldBlock | std::io::ErrorKind::TimedOut
+                ) =>
+            {
+                break
+            }
+            Err(err) => panic!("read request: {err}"),
```
weboko, line 36: Minor / optional. `ErrorKind::Interrupted` falls through to this `panic!`, but EINTR is a *retry* condition, not a failure — `Read::read` on a `TcpStream` surfaces it rather than restarting the syscall (unlike `read_exact` / `read_to_end`, which retry internally).

It is unlikely to fire here — nothing in the test binary installs a signal handler — so this is defensive rather than a bug I can demonstrate. Noting it only because this helper exists specifically to remove a nondeterministic failure, so a residual "rare spurious panic" arm is worth spending one line on:

```suggestion
            Err(err) if err.kind() == std::io::ErrorKind::Interrupted => continue,
            Err(err) => panic!("read request: {err}"),
```

For what it's worth, the panic itself is the right call and is properly observable — `handle.join().expect("server thread")` at `deploy.rs:892` and `:923` propagates it instead of swallowing it in a detached thread.

---

weboko (commented):
Re-reviewed at `875c5d9`. Both comments I left on the previous round are genuinely closed, and I verified the fixes rather than taking the summary at its word.

**Environment.** Provisioned per the runbook's own directives — Nix 2.35.2, Rust 1.94.1, r0vm 3.0.5 matching the LEZ `Cargo.lock` pin, risc0 guest Rust 1.91.1, risc0 C++ 2024.01.05. `rzup` fails exactly as `DOGFOODING.md:16` predicts (`InvalidCertificate(UnknownIssuer)` — this sandbox intercepts TLS on *all* egress, direct connections included), and the documented `curl` fallback worked verbatim. That section earned its keep.

**What I verified**

- **The flake fix holds.** 200/200 consecutive runs of `read_head_with_retry_recovers_after_transient_failure`, driving the test binary directly. Matches your 200-run claim.
- **Full suite green**: 544 lib + 182 cli + 3 testnode_api + 5 doc-tests, 0 failed. `cargo fmt --check` clean. `cargo clippy --all-targets --all-features` reports nothing on the new files.
- **The `#1` review comment is closed.** `basecamp --help` emits exactly the eleven documented verbs in the documented order, and `docs/basecamp-module-requirements.md:3` now carries that same list. I diffed `basecamp docs` output against the file — byte-identical, so the `include_str!` path is really the surface being fixed. All B1 signals present (`env_file`, `runtime_dir`, `log_file`, `[repos.basecamp.attr]`).
- **The `#2` review comment is closed** — the `Content-Length` framing assumption is now documented on the helper.
- **The product code is untouched.** The whole `deploy.rs` diff is inside `#[cfg(test)]`, and `test_support` is gated `#[cfg(test)]` at the `mod` declaration, so none of this reaches a release binary. Nothing here touches key material, wallet state, or the `report` sanitisation path.
- **Guest-path claims are consistent everywhere.** `GUEST_BIN_SEARCH_ROOTS` is `["target/riscv-guest", "methods/target"]`, and `DOGFOODING.md`, `FURPS.md`, `ADR.md`, `constants.rs` and `skills/lez-template/SKILL.md` now all describe those same two roots with the workspace layout as the default. No stale copy left.
- **`DOGFOODING.md` is up to date** for what this PR changes: the B1 verb list and the D1 artefact-path signal are both corrected, and `skills/lgs-cli/SKILL.md`'s `B1–B6` claim matches the six `## B` sections actually in the runbook.

**What I'd still like changed**

One non-blocking finding, on the new regression test itself — the `embedded_docs` assertions are far weaker than the skills assertions beside them, and I confirmed by mutation that the document can regress in exactly the way this PR is fixing while the test stays green. Details and a verified replacement inline. Plus one optional one-liner on the drain helper.

Neither is a merge blocker. The flake fix is correct and demonstrated, and the guidance sync is real.

---

weboko:
Follow-up: I've now run the D-series end to end against a real sequencer, rather than only the unit suite. Both of this PR's doc claims hold empirically.

**`build` → guest artefact paths.** The corrected `DOGFOODING.md:237` / `skills/lez-template/SKILL.md:72` claim is exactly right. On a freshly generated default-template project the five guests land at:

```
target/riscv-guest/example_program_deployment_methods/example_program_deployment_programs/riscv32im-risc0-zkvm-elf/release/*.bin
```

matching the documented `target/riscv-guest/<methods-crate>/<guest-crate>/riscv32im-risc0-zkvm-elf/release/` shape. `methods/target/` **does not exist** in this project — so the line master still carries was not merely imprecise, it named a path that is never produced by the default template. Sizes are ~370 KiB each, consistent with the pacing rationale in the same bullet.

**`deploy` submits from those same paths.** Real `localnet start` → `deploy` against the pinned LEZ sequencer, 5/5 succeeded, 0 failed. Every submission line resolves through `target/riscv-guest/...`, confirming the "the same paths `deploy` submits from" half of the claim:

```
OK  hello_world submitted
Waiting for a new block past 1 before the next deployment (one program per block; bedrock inscription size cap)...
OK  hello_world_with_authorization submitted
Waiting for a new block past 2 ...
...
Summary:
  Succeeded: 5
  Failed: 0
```

Pacing engaged between every pair of submissions and the sequencer survived the full run — which is the live counterpart of the stub-driven test this PR de-flakes. `doctor` afterwards: **24 PASS, 0 WARN, 0 FAIL — "Doctor status: Ready"**.

So the D1 signals you rewrote are accurate as written, and the runbook is up to date for what this PR changes. My inline comments above are the only outstanding items, and neither blocks merge.

---

src/template/skills.rs
danisharora099: Addressed in verified signed commit `6d148fe`. The embedded-doc guard now excludes the intentional migration-note line from the schema assertion and rejects legacy schema anywhere else. The canonical opening verb sentence is parsed and compared exactly, in order, with clap—so missing verbs, reordered verbs, and extra phantom verbs such as `reset` all fail. Mutation checks confirmed failures for a missing `install`, an added `reset`, and legacy schema outside the migration note.

src/commands/test_support.rs
danisharora099: Addressed in verified signed commit `6d148fe`: `ErrorKind::Interrupted` now retries the read loop instead of panicking.

danisharora099:
@weboko the two new non-blocking threads are closed on verified signed commit `6d148fe`.

The embedded-doc regression guard now compares the canonical opening verb list exactly and in order with clap, rejects legacy schema outside the intentional migration note, and was mutation-checked against a missing verb, a phantom `reset`, and legacy-schema drift. The socket helper also retries `EINTR`.

Verification remains green locally: formatting, focused regression, full suite **544 + 182 + 3 + 5**, changed-file clippy audit, and socket recovery **200/200**. Both threads are resolved; please approve if satisfied.

weboko:
## Re-review of `6d148fe` — both open threads closed; verified end-to-end against a real sequencer

I re-ran this from a clean container: Nix, the risc0 3.0.5 toolchain (r0vm + guest Rust), and a real LEZ sequencer built from the pinned commit. Everything in the PR description reproduces. One new non-blocking finding at the bottom.

### The two threads I left on `875c5d9` are properly closed

**`src/template/skills.rs` — the guard is now real, not a tautology.** I mutation-tested it seven ways; every one fails the test:

| Mutation | Result |
| --- | --- |
| Drop `run` from the docs canonical list | ✅ caught (`verb list must match the clap command surface exactly`) |
| Reorder `doctor` / `paths` in the docs list | ✅ caught |
| Add a phantom `teleport` verb to the docs list | ✅ caught |
| Rename `basecamp develop` out of `skills/basecamp/SKILL.md` | ✅ caught (`basecamp is missing the basecamp develop command`) |
| Rename `basecamp doctor` out of `skills/lgs-cli/SKILL.md` | ✅ caught |
| Reintroduce `[basecamp.modules.foo]` into a skill | ✅ caught |
| Add a **new clap subcommand** and wire it, leaving docs/skills untouched | ✅ caught |

That last one is the case I actually cared about — deriving the verb set from clap means the CLI can no longer grow a subcommand without the shipped guidance failing CI. The exact-and-ordered comparison against the embedded docs closes the "incidental mention in later prose" hole I raised. Confirmed the guard covers exactly the surfaces that need it: `skills/basecamp` and `skills/lgs-cli` are the only shipped skills that mention basecamp at all (`lez-template` / `lez-framework-template` have zero mentions), and the sole surviving `[basecamp.modules` in the tree is the intentional migration note at `docs/basecamp-module-requirements.md:39`.

**`src/commands/test_support.rs` — `ErrorKind::Interrupted` now `continue`s** instead of falling through to `panic!`. Correct.

### Verification

- **Full suite: 544 + 182 + 3 + 5 passed, 0 failed.** Matches your numbers exactly.
- `cargo fmt --check` clean. `cargo clippy --all-targets --all-features`: zero warnings on any changed line (the 56 lib warnings are all pre-existing).
- `read_head_with_retry_recovers_after_transient_failure`: **200/200 passed.**
- `basecamp --help` emits exactly the eleven documented verbs, in the documented order. `basecamp docs` matches at runtime too.
- **D1 end-to-end against a real sequencer** (not a stub): `new` → `create` → `setup` → `localnet start` → `build` → `deploy` → `wallet topup` → `wallet -- check-health`. All exited 0. `setup` seeded the default wallet from a preconfigured account; **5/5 programs deployed, 0 failures**, correctly paced one per block (`Waiting for a new block past 16…19`), ~52s total — right at the documented ~15s-per-additional-program budget. `wallet topup` executed a real `auth-transfer init` + `pinata claim`; `check-health` returned `✅All looks good!`.

### The doc corrections are right, and the text they replaced was wrong

I checked these rather than taking them on faith:

- **Guest artifact paths.** After `build`, the five `.bin` files are at `target/riscv-guest/example_program_deployment_methods/example_program_deployment_programs/riscv32im-risc0-zkvm-elf/release/` and **`methods/target/` does not exist at all**. The old D1 line ("produces a `methods/target/.../release` artifact") was factually wrong for the default template; your replacement matches reality, and `deploy` visibly submitted from exactly those paths. The ELFs are 371–377 KiB, which independently corroborates the "~370 KiB" figure in the pacing rationale.
- **Two search roots.** `GUEST_BIN_SEARCH_ROOTS = ["target/riscv-guest", "methods/target"]`, workspace layout first — so the FURPS #4 and `METHODS_DIR` rewordings are accurate, and both roots have test coverage.
- **`[basecamp.dependencies]`.** Confirmed it is parsed *nowhere* in `config.rs`; the old comment pointed users at a config surface that never existed. The replacement ("capture an explicit `[modules.<name>]` entry") is the mechanism that actually works — `resolve_manifest_dependencies` skips names already in `captured` regardless of role.
- The runbook's own rzup caveat held up: `rzup install r0vm 3.0.5` failed with exactly the documented `InvalidCertificate(UnknownIssuer)`, and the curl fallback worked as written.

### Security / architecture

No production surface. `test_support` is `#[cfg(test)]`-gated at the `mod` declaration, so the drain helper cannot be reached from a release build, and it only ever operates on already-accepted loopback sockets. The security-relevant function in this diff — `normalize_and_validate_module_name`, the section-header injection guard — has **only its error-message string changed; the charset validation is untouched**, and its test assertion was tightened from `||` to `&&` in the process. Extracting the shared helper into `commands::test_support` rather than duplicating it in two `mod tests` fits the existing module layout.

---

### One new finding — non-blocking: README is a fourth guidance surface with the same drift

This PR synchronizes three surfaces (DOGFOODING B1, the embedded compatibility doc, and the two shipped skills), but `README.md` — the most user-facing one — still carries a stale basecamp verb list. It is missing **four of the eleven** verbs, in both places:

- the usage block at `README.md:125–131` lists only `setup`, `modules`, `install`, `launch`, `develop`, `build-portable`, `doctor`
- the prose at `README.md:159–165` documents the same seven

Missing from both: **`build`**, **`run`**, **`paths`**, **`docs`**. (`README.md:130` also omits `build-portable`'s `--module NAME` flag, which the skill documents.)

That is precisely the drift class this PR set out to eliminate, and a reader who starts at the README today cannot discover `basecamp build` or `basecamp run` at all. Since the new test already derives the verb set from clap, extending the `skills` array to include the README's command block would close it with a couple of lines — the same `contains_command` helper works unmodified.

Relatedly, **`DOGFOODING.md` itself remains unguarded**: no test reads it, so the B1 verb list you just corrected by hand can silently go stale again the same way it did before. Worth folding into the same guard.

Neither blocks merge — both are pre-existing staleness this PR simply doesn't reach, and the three surfaces it *does* claim are now correct and enforced. I'd take them as a short follow-up.

---

weboko (approved):

PR #249, opened 2026-08-10
fix: address the non-blocking review comments left open by #244, #245, #246

weboko:
Follow-up to #246, #245 and #244. Merge those three first — this branch is stacked on all three and only makes sense on top of them.

Direct reviewer demand: every item here is a comment I left on #244, #245 or #246 that was flagged non-blocking and therefore left open when those PRs were approved. Nothing is invented — each change traces to a specific review comment, linked below.

Non-blocking is not the same as not worth doing, and the three most substantive items here (the untested `check_standalone_support` probe, the missing depth assertion on the basecamp module root, and the ADR that still says the pipeline has no skip knobs) are all cases where the repo's own written contract or its test suite would stay green through a real regression.

## User Story

Reviewing #244/#245/#246 surfaced twelve findings that were correct but not worth holding three approved PRs for. Left alone they decay into the usual failure mode: the next person reads `ADR.md` and believes the pipeline has no skip knobs, or edits `basecamp_xdg_subpath` and ships a basecamp launch that loads zero project modules with all 738 tests passing.

Surfaced by: the review passes on #244 (`B3`), #245 (`D7`) and #246 (`D1`, `D4`, `D6`, `D7`). Tried first: leaving them as review comments — which is exactly how they ended up hanging.

## Change Summary

Addresses the twelve open non-blocking findings across the three PRs: adds the three missing tests, moves one misplaced doc comment, routes one env-set through the shared helper, hoists one inline env-var list to `constants.rs`, and reconciles five documentation surfaces (ADR, README, FURPS, DOGFOODING, `run --help`) with what the code now does. No behavior change — the only non-test, non-doc edits are a helper call that sets the same two variables it set before, and a constant moved between files.

### From #246 — [first review](https://github.com/logos-co/scaffold/pull/246#pullrequestreview-4825560614), [re-review](https://github.com/logos-co/scaffold/pull/246#issuecomment-5242620934)

| Finding | Change |
| --- | --- |
| "`check_standalone_support`'s nested probe is untested … the only production change in this PR with no test, and it is the remainder of #234's layout work, i.e. the exact thing that was missed once already" | Three tests in `doctor_checks.rs`: flat (`sequencer/service/Cargo.toml`) passes, nested (`lez/sequencer/service/Cargo.toml`) passes, and a checkout carrying all four probed files *without* the marker still fails — so the two passing cases cannot pass for the wrong reason |
| "`run.rs`'s hook builder iterates `WALLET_HOME_ENV_VARS` directly … `set_wallet_home_env` would type-check unchanged there and would make the invariant enforceable by grepping for a single function name" | `build_hook_command` now calls `set_wallet_home_env`. Both spellings set the same two vars to the same path; this makes "every wallet-home consumer gets every name" a one-symbol grep |
| "`DOGFOODING.md` describes the skipped candidates as `Public/` addresses printed anywhere else (a banner or status line). On the real wallet they are neither — they are four `Preconfigured Public/…` entries, which a reader could easily assume *are* the listing" | Reworded in `DOGFOODING.md` D1 and in the `first_public_address_in_listing` doc comment, both naming `Preconfigured` explicitly and recording that the wallet prints them even when the config ships no `initial_accounts`. Your point that this "also makes the fallback tier's risk legible" is carried too: D1 now says a seeded address matching a `Preconfigured` line rather than a `/ ` one is worth reporting |
| "a misplaced doc comment in `run.rs` … flagging it because this PR is the one touching `reseed_after_wipe`'s signature, so it is a cheap drive-by if you are in the file" | Doc comment moved onto `reseed_after_wipe`; `watch_loop` got its own |
| "`WALLET_HOME_ENV_VARS` is ordered `[NSSA, LEE]` and both are always set to the same value, so order is irrelevant today … a one-line comment would prevent someone from assuming it is" | Recorded at the constant, with the condition that would make it stop being true |

The `Preconfigured` correction is also pinned by a test, built from the real `wallet account list` output in your re-review — the case where a first-`Public/`-token scan adopts `6iArKUXx…`, four lines above the account the wallet actually created. It fails against a single-tier scanner.

### From #245 — [first review](https://github.com/logos-co/scaffold/pull/245#pullrequestreview-4825558410), [re-review](https://github.com/logos-co/scaffold/pull/245#issuecomment-5242610276)

| Finding | Change |
| --- | --- |
| "`ADR.md:342-343` … *Step ordering is fixed; if a new step is needed, it joins the chain rather than offering a knob.* `deploy = false` was the first knob; `topup = false` is the second … someone should reconcile it, ideally here since this is the change that turns a one-off into a pattern" | ADR rewritten rather than appended. Ordering is still fixed; what changed is that a step may expose a boolean skip. Records the boundary — the project can legitimately own that concern, never "scaffold has a bug here" — and the three constraints that keep the pipeline shape visible (step keeps its slot and prints a skipped header; default is not-skipped via a hand-rolled `Default`; the decision is exported to hooks). A step whose skip cannot satisfy all three does not get a knob |
| "`resolve_profile` **selects**, it does not merge … the README's *It works inline under `[run]` or per profile* is accurate but reads like the two compose … a half-sentence would cover all three keys at once" | Stated once for all profile keys in the README, with the `topup = false` + `--profile demo` example. Already pinned by `resolve_profile_default_profile_fully_shadows_inline` in `config.rs`, never documented |
| "three docs need the same one-line follow-through, since the hook env is treated as a public surface: README's prose list, DOGFOODING's env-contract failure signal, ADR's bullet list" | The README env *table* got the two vars during #245; these three enumerations did not. All now list `SCAFFOLD_TOPUP_SKIPPED` / `SCAFFOLD_DEPLOY_SKIPPED`, plus FURPS. The DOGFOODING one is the line a dogfooder checks against, and until now it passed with either var missing — it now also flags a value that disagrees with the step header printed in the same run |
| "`topup = false` combined with the default `deploy = true` on an unfunded wallet will fail at step 5 rather than step 4 … may be worth one clause noting that the project is then responsible for funding *before* deploy" | README clause. Skipping topup moves the responsibility, not the requirement — and a project funding from a `post_deploy` hook generally wants `deploy = false` too, since that hook runs after deploy |
| "`run --help`'s summary line still reads *Build, start localnet, top up wallet, deploy, and run post-deploy hook*, which a `topup = false` profile makes partly untrue — as `deploy = false` already did … if the summary is ever reworded, both belong in it" | Reworded, both in it. Kept under 100 chars so clap does not wrap it; the assertion in `tests/cli.rs` moved with it |

Not done, deliberately: `topup` and `deploy` still have no `--topup` / `--no-topup` CLI counterpart. You raised the `reset`-has-flags asymmetry but said it is consistent with how #239 shipped `deploy` and that resolving it is worth doing in one go if a third config-only skip appears — so it stays a whole-family decision rather than a third one-off here.

Also not done: a CLI-level test for the step-4 skip branch (your §3). `run_pipeline_once` needs a real sequencer, so a `tests/cli.rs` test cannot reach the branch; `topup_step_skipped_follows_the_profile_flag` plus D7 remain the coverage, as you and the author both concluded.

### From #244 — [first review](https://github.com/logos-co/scaffold/pull/244#pullrequestreview-4825186093), [re-review](https://github.com/logos-co/scaffold/pull/244#issuecomment-5242365151)

| Finding | Change |
| --- | --- |
| "a unit test asserting the default value **ends in** `Logos/LogosBasecamp` and equals the dir `install` writes into would keep a future `basecamp_xdg_subpath` change from silently breaking 0.2.x while all six existing tests stay green" | Added. This is the one I would most want kept: 0.2.x consumes `LOGOS_USER_DIR` as the base and appends `modules/` and `plugins/` itself, so a value one level off loads zero project modules with the suite fully green. The new test pins the exported value against `profile_modules_and_plugins` — the one place the install layout lives — for both keys, and asserts the portable XDG subpath suffix. Until now that relationship was pinned only by a log excerpt in a PR description |
| "`for key in [\"LOGOS_DATA_DIR\", \"LOGOS_USER_DIR\"]` hardcodes the list inline … a `BASECAMP_MODULE_ROOT_ENV_VARS` constant would match the house style and give the next rename one place to edit … #246 is carrying exactly that shape as `constants::WALLET_HOME_ENV_VARS`" | Hoisted, sitting next to `WALLET_HOME_ENV_VARS`. Both now carry the same note on the next rename: add the new name, keep the old, never swap |

The author offered to add the depth assertion to #244 directly; it is here instead so #244 can merge as approved.

## Verification

`cargo test` **738 green, 0 failed** (548 lib + 182 CLI + 3 testnode + 5 doc) on the three branches merged together. `cargo fmt --check` clean. `cargo clippy --all-targets` — 60 warnings before, 60 after, byte-identical set.

Baseline for that number: the three PR branches merged onto `master@9fcc376` give 733, so the +5 delta is exactly the new tests (3 × `check_standalone_support`, 1 × basecamp depth, 1 × `Preconfigured` listing). Each of the three commits compiles clean on its own with no unused-item warnings.

Each new test was checked to fail against the code it guards:

- the `Preconfigured` listing test fails against a single-tier `split_whitespace` scanner, returning `6iArKUXx…`;
- `standalone_support_found_in_nested_layout` fails with the `lez/sequencer/service/Cargo.toml` probe removed;
- the basecamp depth test fails if the exported default gains or loses a path component.

Merge-conflict resolution, since the three PRs conflict with each other: the one hunk is `hook_receives_wallet_home_dir_env` in `src/commands/run.rs`, resolved by taking both sides — #246's dual-name `printf` **and** #245's trailing `false` — exactly as called out on both PRs. Resolving toward #245's side alone would silently drop the `LEE_WALLET_HOME_DIR` assertion.

**DOGFOODING scenarios rerun: none.** This branch changes no runtime behavior, so there is nothing for a scenario to observe that the three parent PRs' own runs did not already cover — and the scenarios that would be routed here (`D1`, `D4`, `D6`, `D7`, `B3`) need a real LEZ toolchain and, for `B3`, a macOS portable stack. The changes to those runbook entries are corrections to expected-signal text, verified by reading them against the code they describe. Concretely unverified on my side, same as on #244 and #246: the macOS portable launch path (`cfg!(target_os = "macos")`, unreachable on Linux) and the LEZ v0.2.0 seeding path (`DEFAULT_LEZ` is still v0.1.2).

The "Change → Scenario" map needs no new entry — wallet-home env contract changes already route to `D1`/`D4`/`D6`/`D7`, `run` step-skip and post-deploy env changes to `D7`, and `basecamp launch` env export to `B3`.

## Checklist

- [X] CI is green
- [x] Relevant DOGFOODING scenarios rerun (listed above) — none apply; reasoning above
- [x] Docs updated (`README.md`, `DOGFOODING.md`) if user-facing behavior changed


src/commands/basecamp.rs
```diff
@@ -3940,6 +3941,49 @@ mod tests {
...
+                Some(exported.as_path()),
+                "{key} must be the parent of the modules dir install writes into"
+            );
+            assert_eq!(
+                plugins_dir.parent(),
+                Some(exported.as_path()),
+                "{key} must be the parent of the plugins dir install writes into"
+            );
+            assert!(
+                exported.ends_with(BASECAMP_XDG_APP_SUBPATH_PORTABLE),
```
weboko: This assertion — and in fact all three in this test — cannot fail for the regression the test is named after.

`exported` is produced by `set_absolute_basecamp_data_dirs` → `basecamp_xdg_subpath` → `BASECAMP_XDG_APP_SUBPATH_PORTABLE`. `modules_dir`/`plugins_dir` come from `profile_modules_and_plugins`, which calls the *same* `basecamp_xdg_subpath`. And this `ends_with` compares against the constant itself. So every side of every assertion moves in lockstep with the value being pinned.

I verified this rather than reasoning about it. Changing the constant to the one-level-off value your doc comment describes:

```rust
pub(crate) const BASECAMP_XDG_APP_SUBPATH_PORTABLE: &str = "Logos/LogosBasecamp/WRONG_DEPTH";
```

leaves **all 738 tests green, this one included** — which is exactly the "loads zero project modules with the suite fully green" outcome the test exists to prevent.

What the test *does* catch is a depth change in `set_absolute_module_root_var` alone: I mutated it in both directions (dropping `.join(basecamp_xdg_subpath(…))`, and appending an extra `.join("modules")`) and it failed both times. So it genuinely pins this call site against `profile_modules_and_plugins` — that part is real and worth keeping. It just does not pin the shared value, which is the half that is an external contract with basecamp 0.2.x's `LogosBasecampPaths.h::baseDirectory()` rather than something scaffold gets to choose.

The fix is to state the contract as a literal, so the expectation cannot follow the code:

```suggestion
            assert!(
                exported.ends_with("Logos/LogosBasecamp"),
                "{key} must end in Logos/LogosBasecamp (basecamp 0.2.x's own \
                 baseDirectory()); changing this is an upstream-contract change, \
                 got: {}",
                exported.display()
            );
```

Verified both ways: with this literal in place the `WRONG_DEPTH` mutation fails the test with a clear message, and it passes unchanged against the real constant.

Worth a matching tweak to the doc comment above — as written it promises coverage of a `basecamp_xdg_subpath` change, which only becomes true with this edit.

---

src/commands/run.rs
```diff
@@ -809,10 +812,11 @@ fn build_hook_command(
...
     // Both wallet home names (old NSSA_*, v0.2.0 LEE_*) so hook-spawned
-    // wallet CLIs target the project wallet regardless of the LEZ pin.
-    for name in WALLET_HOME_ENV_VARS {
-        cmd.env(name, &wallet_home);
-    }
+    // wallet CLIs target the project wallet regardless of the LEZ pin. Routed
+    // through the same helper every wallet subprocess uses rather than
+    // iterating `WALLET_HOME_ENV_VARS` here, so the "every wallet-home consumer
+    // gets every name" invariant is greppable by one function name.
+    set_wallet_home_env(&mut cmd, &wallet_home);
```
weboko, line 819: Minor, and adjacent rather than in this hunk — but this is the PR that makes it worth raising.

The two vars set just above (`SCAFFOLD_DEPLOY_SKIPPED` / `SCAFFOLD_TOPUP_SKIPPED`) are documented by this PR as always-on public contract in four places at once: README's prose list, FURPS §3, the ADR bullet list, and DOGFOODING's D7 failure signal — where an unset one is now explicitly called out as *itself* the regression.

Meanwhile `hook_receives_full_env_contract_in_one_invocation` (`run.rs:1200`) opens with:

> Integration-style assertion: every documented always-on env var reaches the hook in a single shell invocation, in the same form `cmd_run` would produce.

and then covers only `SEQUENCER_URL`, the two wallet-home names, `SCAFFOLD_PROJECT_ROOT`, and `SCAFFOLD_IDL_DIR`. The two skip flags are tested elsewhere (`hook_receives_topup_skipped_env_for_multiprogram_run` and friends), so the behaviour is not unguarded — but the one test whose stated job is to hold the contract *as a whole* has drifted from the contract this PR is writing down. That is the same failure mode the rest of this branch is fixing, one level up.

They do reach that code path; I confirmed by instrumenting the test, which captured:

```
SCAFFOLD_TOPUP_SKIPPED=0
SCAFFOLD_DEPLOY_SKIPPED=0
```

So it is two echo lines plus two assertions, and no production change.

---

weboko (commented):
## Review — verified hands-on, one finding worth fixing before merge

Reviewed the full diff and re-ran the verification independently on Linux (Ubuntu 24.04, Rust 1.94.1), on `7fabccf` rebased context — the branch is now a single commit on top of `master@d403a95`, since #244/#245/#246 have all landed. Every number in the description reproduces.

### Reproduced

| Claim | Result |
| --- | --- |
| `cargo test` 738 green | **548 + 182 + 3 + 5 = 738 passed, 0 failed** |
| `cargo fmt --check` clean | clean |
| clippy "60 before, 60 after, byte-identical set" | **60 → 60, sets diff identical** (isolated `CARGO_TARGET_DIR` per branch) |
| `run --help` summary under 100 chars | 96 |
| Merge resolution keeps both sides | confirmed — `hook_receives_wallet_home_dir_env` asserts both names *and* carries the trailing `false` |

I also independently mutation-tested the three "each new test was checked to fail against the code it guards" claims, and all three hold:

- **Removing the `lez/sequencer/service/Cargo.toml` probe** → exactly 1 failure, `standalone_support_found_in_nested_layout`.
- **Collapsing `first_public_address_in_listing` to a single-tier `split_whitespace` scan** → the new test fails returning `Public/6iArKUXx…`, the exact value you predicted, along with 2 pre-existing tests.
- **Changing the exported default's depth in `set_absolute_module_root_var`** (both losing and gaining a component) → the new depth test fails in both directions.

The docs check out against the code too: the "always set, never absent" claim for `SCAFFOLD_TOPUP_SKIPPED` / `SCAFFOLD_DEPLOY_SKIPPED` matches the unconditional `.env(…)` calls in `build_hook_command`; the README's new profile-shadowing paragraph matches `resolve_profile_default_profile_fully_shadows_inline` ("inline reset must not bleed into resolved profile"); and the ADR's "hand-rolled `Default` rather than `derive(Default)`" matches `impl Default for RunProfile` in `model.rs:373`. The `set_wallet_home_env` swap in `build_hook_command` is exactly behaviour-preserving — same two names, same `OsStr`-valued path.

Worth stating plainly because the description undersells it: the `Preconfigured` fix is **security-relevant**, not just cosmetic. The entries it now skips are the wallet binary's own built-in accounts — the listing prints `Preconfigured Private/…` right next to them, so those keys are shared by every developer on the same LEZ pin. A first-token scan adopting one would make it the project's default topup destination, i.e. faucet funds sent to an account whose private key ships in the binary. Localnet-only, but that is the right reason to keep the `/ ` tier load-bearing and pinned by a test.

### One finding — the depth test does not catch the regression it names

Detailed inline on `src/commands/basecamp.rs`. Short version: the test is described (in the PR body and in its own doc comment) as guarding against "a `basecamp_xdg_subpath` change that moved the shared value to the wrong depth, which would load zero project modules while the suite stayed green." I changed `BASECAMP_XDG_APP_SUBPATH_PORTABLE` to `"Logos/LogosBasecamp/WRONG_DEPTH"` — precisely that regression — and **all 738 tests stayed green, this one included.** Every assertion in it derives both sides from `basecamp_xdg_subpath`, so they move in lockstep with the value they are meant to pin.

The test does have real value — it caught both of my `set_absolute_module_root_var` depth mutations, so it genuinely pins that call site against `profile_modules_and_plugins`. It just does not pin the constant, which is the part that is an external contract with basecamp's `LogosBasecampPaths.h`. One-line fix in the inline comment, verified to catch the mutation and to pass on the correct value.

Given this is the test you said you would "most want kept", it seems worth making it hold the line it claims.

### Second, minor

Also inline: `hook_receives_full_env_contract_in_one_invocation` describes itself as asserting "every documented always-on env var … in a single shell invocation", but does not include the two vars this PR documents as always-on across four surfaces. Two lines.

### DOGFOODING.md

Agreed that no new scenario is warranted — this branch changes no runtime behaviour, and the two edits (D1's `Preconfigured` wording, D7's env-contract failure signal) are corrections to expected-signal text on existing scenarios rather than new surface. I checked the D7 signal is now actually falsifiable: previously it passed with either var missing; it now names both and calls out disagreement with the step header. I verified the underlying behaviour directly rather than by reading — the hook tests spawn real `sh -c` processes through `build_hook_command`, and the two vars arrive as `SCAFFOLD_TOPUP_SKIPPED=0` / `SCAFFOLD_DEPLOY_SKIPPED=0`.

Not verified on my side, same as you flagged: the macOS portable launch path and the LEZ v0.2.0 seeding path are unreachable on Linux without a real toolchain.

---

weboko:
## Follow-up: D7 was runnable after all — executed end to end, all signals matched

You recorded "DOGFOODING scenarios rerun: **none**", on the reasoning that the routed scenarios need a real LEZ toolchain. I provisioned one on Linux (Nix + risc0 rust/cpp toolchains + a source-built LEZ `sequencer_service` and `wallet` at the pinned `cf3639d8`) and ran **D7 in full against this branch's binary**. Everything passed. Posting the evidence since it converts your largest unverified block into a verified one.

For the record on provisioning, since the runbook's agent directives call it out: `rzup` fails here with `InvalidCertificate(UnknownIssuer)` because its GitHub client uses pinned built-in roots and this sandbox TLS-intercepts egress — no env var fixes it. The curl-with-`$GH_TOKEN` fallback works; I installed the components into the `~/.risc0` layout by hand and `rzup show` picks them up.

### D7 results

Bare `run` printed `[1/5]`…`[5/5]`, reused the running localnet (`localnet already running (sequencer pid=24738)`), 5 succeeded / 0 failed.

With the runbook's `[run].post_deploy` block, every documented env var arrived non-empty, `cwd` was the project root:

```
sequencer: http://127.0.0.1:3040
idl:                    <project>/idl
project root:           <project>
cwd:                    <project>
wallet home:            <project>/.scaffold/wallet
wallet home (v0.2.0):   <project>/.scaffold/wallet
topup skipped: 0
deploy skipped: 1
```

**The two wallet-home names printed the same path** — the live confirmation of the invariant your `set_wallet_home_env` swap in `build_hook_command` exists to protect.

Skip profiles, both exactly as documented:

| Run | Header | Hook sees |
| --- | --- | --- |
| `--profile self-deploy` | ``[5/6] Deploy skipped (`deploy = false` in the run profile; …)`` | `deploy skipped: 1` |
| `--profile self-fund` | ``[4/6] Topup skipped (`topup = false` in the run profile; …)`` | `topup skipped: 1` |
| no profile, same hook | `[4/6] Topping up wallet...` | `topup skipped: 0` |

So the "always set, never absent" claim you added to D7 holds against a real run, and the value never disagreed with the step header.

Two things this incidentally verified that were previously only pinned by unit tests:

- **The ADR's three constraints.** Step numbering stayed `/6` in both skip runs — the skipped step kept its slot and nothing was renumbered, exactly as the rewritten ADR says.
- **The README's new profile-shadowing paragraph.** Both profiles ran **1** hook, not the 9 configured inline — `resolve_profile` selecting rather than merging, observed live rather than via `resolve_profile_default_profile_fully_shadows_inline`.

Flag overrides also matched: `--post-deploy` ran only the override, `--no-post-deploy` dropped the step count to 5, and the two together failed at clap parse with exit 1.

### One observation, not a defect

`SCAFFOLD_DEPLOY_SKIPPED=1` is also what a **dedup**-skipped deploy reports, not just `deploy = false`. That is self-consistent — it agrees with the `[5/6] Deploy skipped (guest binaries + IDL + config + sequencer unchanged…)` header, so it never trips D7's new disagreement signal — and the ADR's framing ("whether scaffold acted on this run") covers it. But FURPS's "reporting whether the run's skippable steps ran" sits next to the `topup`/`deploy` knobs and could be read as knob-specific. If you touch that line again, "whether the step ran, for any reason" would remove the ambiguity.

### Still genuinely unverified

Your two disclosures stand and I could not close them: the macOS portable launch path is `cfg!(target_os = "macos")`, and the v0.2.0 seeding path is unreachable while `DEFAULT_LEZ` is v0.1.2 — my `setup` took the other branch and printed `default wallet seeded from preconfigured account`, so the `Preconfigured` scan this PR pins is still covered only by its unit test.

---

ADR.md
```diff
@@ -351,6 +370,16 @@ integration tests:
...
   `SCAFFOLD_PROJECT_ROOT` / `SCAFFOLD_IDL_DIR` — pipeline state. The two
   wallet-home vars always carry the same path.
+- `SCAFFOLD_TOPUP_SKIPPED` / `SCAFFOLD_DEPLOY_SKIPPED` — run-level outcome
+  of the two skippable steps, `1` or `0`. Always set, never absent: a hook
+  that had to treat "unset" as "scaffold did it" could not distinguish that
+  from an older scaffold that cannot report. Each is produced by the same
+  expression that executes or skips its step, so what a hook is told cannot
+  drift from what ran. Neither has a per-program form — both describe the
+  run, not a program.
+- `SCAFFOLD_DEPLOY_SKIPPED_<name>` — per-program deploy outcome, alongside
```
danisharora099: These two bullets contradict each other as written: the first ends with "Neither has a per-program form — both describe the run, not a program", and this one documents `SCAFFOLD_DEPLOY_SKIPPED_<name>` as a "per-program deploy outcome". The code says neither is quite right — `build_hook_command` sets every `SCAFFOLD_DEPLOY_SKIPPED_<name>` from the same run-level `run_deploy_skipped` (its own comment: "deploy-skip state is run-level (it's the same for every program in this invocation)"), so the suffixed form exists, but it is not an independent per-program outcome.

Suggest scoping the "no per-program form" sentence to `SCAFFOLD_TOPUP_SKIPPED` only, and describing `SCAFFOLD_DEPLOY_SKIPPED_<name>` as the run-level value duplicated under each program's suffix so the per-program var family stays parallel — not an independent per-program signal. One-line fix, and this section drifting from the code is exactly what this PR exists to prevent.

src/commands/basecamp.rs
danisharora099: This assertion can't fail independently of the constant it's meant to pin: `exported` is built from `BASECAMP_XDG_APP_SUBPATH_PORTABLE` via `basecamp_xdg_subpath`, and `profile_modules_and_plugins` co-moves through the same function — so a wrong edit to the constant keeps all three assertions green. The two parent assertions above do the real work (exporter-vs-installer divergence, the "zero modules loaded, suite green" failure mode), and those are the ones I most wanted — but the "ends in `Logos/LogosBasecamp`" half of the original ask only bites if this asserts the literal string rather than the constant.

Same caveat for the PR-body claim "fails if the exported default gains or loses a path component": true for edits to either join chain, not for changes routed through the shared constant/subpath function. Non-blocking either way.

danisharora099 (approved):
Verified against the code rather than the PR text: the three `check_standalone_support` tests hit exactly the four probed files and the negative case can't pass for the wrong reason; `set_wallet_home_env` sets exactly `WALLET_HOME_ENV_VARS`, so the `build_hook_command` routing is behavior-identical; `RunProfile`'s hand-rolled `Default` matches the new ADR text; `resolve_profile_default_profile_fully_shadows_inline` backs the new README paragraph; both skip vars come from the same expressions that gate steps 4/5 and `total_steps` doesn't shrink; the help string is 96 chars with the `tests/cli.rs` assertion moved in lockstep; and the `hook_receives_wallet_home_dir_env` resolution kept both the dual-name printf and the trailing `false`.

Two comments inline, both non-blocking — the ADR one is a one-line wording fix I'd take in this PR since reconciling that exact section with the code is what the PR is for; the test one can ride along or follow up. Nothing here holds the release.

DOGFOODING.md
```diff
@@ -1035,9 +1036,9 @@ Validate that a module project can fetch the pinned basecamp + `lgpm` binaries,
...
...ifically**. This is a stricter requirement than the rest of this runbook and the usual reason a `B`-series run stalls in an agent container, so check it before installing anything. Nix resolves `github:` flake inputs over `https://api.github.com/repos/…/commits/HEAD` and `https://github.com/…/archive/<rev>.tar.gz`; a proxy that allowlists GitHub per repository answers `403` on both, and the basecamp closure pulls ~25 repos across `logos-co`, `NixOS/nixpkgs` and `oxalica/rust-overlay`. Neither `git clone` working nor `nix --version` working proves this — probe it directly with `curl -sS -o /dev/null -w '%{http_code}\n' https://api.github.com/repos/NixOS/nixpkgs/commits/HEAD` (expect `200`) before starting. Rewriting the project's own input to `git+https://` does not help: the transitive inputs are already locked as `github:` inside each dependency's own `flake.lock`. If the probe fails, `B1`–`B6` are out of reach in that environment and the honest result is to record the blocker — the scaffold-side surface that needs no Nix (`basecamp --help`, `basecamp docs`, the missing-Nix hint, `basecamp doctor`, out-of-project errors) is still worth exercising and reporting as partial coverage.
```
weboko, line 1039: I ran this precondition in exactly the environment it describes, and the diagnosis is right but the conclusion is too pessimistic — I got `B`-series flake resolution working here.

The probe reproduces perfectly. Through the org proxy, both surfaces you name answer 403:

```
curl -sS -o /dev/null -w '%{http_code}\n' https://api.github.com/repos/NixOS/nixpkgs/commits/HEAD   → 403
curl -sS -o /dev/null -w '%{http_code}\n' -L https://github.com/NixOS/nixpkgs/archive/master.tar.gz → 403
```

and `nix flake metadata github:NixOS/nixpkgs/nixos-unstable` fails with the allowlist's own body ("GitHub access to this repository is not enabled for this session"). So the paragraph correctly identifies what breaks and correctly rules out the `git+https://` rewrite.

But "`B1`–`B6` are out of reach in that environment" does not follow. The 403 is the *intercepting proxy*, not the network — direct egress to `api.github.com` is open here, and the same probe bypassing the proxy returns `200`. Two changes make Nix resolve `github:` inputs:

```bash
# 1. authenticate Nix to GitHub (bypassing the proxy drops you onto a shared egress IP,
#    so unauthenticated requests then hit GitHub's own rate limit — a different 403)
printf 'experimental-features = nix-command flakes\naccess-tokens = github.com=%s\n' "$GH_TOKEN" \
  > ~/.config/nix/nix.conf && chmod 600 ~/.config/nix/nix.conf

# 2. run nix off the intercepting proxy
env -u HTTPS_PROXY -u https_proxy -u HTTP_PROXY -u http_proxy \
  nix flake metadata github:NixOS/nixpkgs/nixos-unstable
```

Result in this container:

```
Resolved URL:  github:NixOS/nixpkgs/nixos-unstable
Locked URL:    github:NixOS/nixpkgs/0e251e24a4f24e036a084b6b4b2d2491af4167f4?narHash=sha256-yNJd40f…
Path:          /nix/store/31w94yhfpllma9jnlgavjkzsfs548ijx-source
```

— resolved *and* unpacked into the store, which is the operation the paragraph says is unavailable. Since the closure is `github:` inputs across three orgs, this is the mechanism that unblocks all of them, not a one-repo special case. (I did not carry it through a full `basecamp setup`, so I can't promise `B1`–`B6` go green end-to-end — but "the probe failed, therefore stop" is no longer the right instruction.)

Two concrete asks:

1. The probe as written runs through the proxy, so it answers 403 in a container where the B-series is in fact reachable. Worth probing both, since the difference is exactly the diagnosis: `curl -sS -o /dev/null -w '%{http_code}\n' --noproxy '*' https://api.github.com/repos/NixOS/nixpkgs/commits/HEAD`. A 403-proxied/200-direct split means "bypass the proxy", not "give up".
2. Replace "the honest result is to record the blocker" with the two-step workaround above, keeping the fallback (partial scaffold-side coverage) for the case where *both* probes 403.

This is the one place in the diff where I'd push back on substance rather than wording: the paragraph is load-bearing advice that tells the next dogfooder to stop, and the stop is avoidable. Everything else about it — the `~25`-repo closure, the transitive-lock explanation, "neither `git clone` nor `nix --version` proves it" — is accurate and worth keeping.

---

DOGFOODING.md
```diff
@@ -341,32 +341,33 @@ export EXAMPLE_PROGRAMS_BUILD_DIR="$PWD/target/riscv-guest/example_program_deplo
 
 Use a known default-template program name such as `hello_world`. If the generated project exposes a different set of programs in `methods/guest/src/bin`, record the discovered list.
 
-`--json` only produces structured JSON output when combined with `--program-path`. On the discovery-based path (`deploy` or `deploy <name>`), the `--json` flag is accepted but silently ignored. This scenario validates that distinction.
+Both deploy paths honor `--json`, with a different shape each. `--program-path --json` prints one flat object for the single submitted program; the discovery-based path (`deploy` or `deploy <name>`) wraps one such object per attempted program in `{"deploys":[…]}`. Either way `--json` is pure JSON on stdout — no command echoes, no summary block. This scenario validates that distinction. (Both shapes omit absent fields rather than emitting `null`: today the pinned LEZ exposes no transaction receipt for a deploy, so `tx` is always absent and consumers test `has("tx")` rather than branching on a guaranteed-null key. `program_id` is present whenever the vendored `spel` binary resolved it, and a failed entry carries `error` in its place.)
```
weboko, line 344: Verified this rewrite against a real sequencer — every claim in it holds, including the two easy ones to get wrong (`tx` absent rather than null, and pure JSON with no echoes on the discovery path). Good correction; the old "silently accepted but ignored" text was wrong on both paths' behalf.

One gap left: the **D3 row in the Scenario Index (line 167)** was not updated with this body, and now contradicts it:

```
| D3 | `default` | Advanced | Deploy path variations and machine-readable single-program submission | `deploy [program-name]`, `deploy --program-path`, `deploy --program-path --json` |
```

The Evidence section below now asks for "one `{"deploys":[…]}` output from the discovery path", but the index still says *single-program* and its command surface lists only `deploy --program-path --json` — neither `deploy --json` nor `deploy <name> --json` appears. A dogfooder scoping the run from the index runs a strict subset of what D3 now requires, which is the same doc-drift failure mode this commit exists to close.

---

src/commands/doctor.rs
```diff
@@ -511,17 +517,28 @@ fn derive_next_steps(rows: &[CheckRow]) -> Vec<String> {
...
             || remediation.contains("NSSA_WALLET_HOME_DIR")
         {
             include_wallet_home = true;
+            mapped = true;
+        }
+        // Two rows can share one remediation (e.g. a missing tool reported by
+        // more than one check); emit it once.
+        if !mapped && !unmapped.iter().any(|seen| seen == remediation) {
+            unmapped.push(remediation.to_string());
         }
```
weboko, line 542: Nit, non-blocking — the dedup is against `unmapped` only, while the trailing `STEP_DOCTOR` is appended unconditionally at line 557. So an unmapped remediation that itself ends in ``…, then `logos-scaffold doctor` `` would render the re-run twice.

No current remediation does this — I checked all of them. The two that embed `STEP_DOCTOR` are both mapped and therefore never reach `unmapped`:

- `"Run \`logos-scaffold localnet start\` (…), then \`logos-scaffold doctor\`"` → mapped via `STEP_LOCALNET_START`
- `"Verify wallet config and run \`{STEP_EXPORT_WALLET_HOME}\`, then \`logos-scaffold doctor\`"` → mapped via `STEP_EXPORT_WALLET_HOME`

But that is a property of today's strings, not of the code — and ``…, then `logos-scaffold doctor` `` is clearly the house phrasing for remediations, so the next one added for a non-canonical check lands in exactly this hole. A one-line guard pins it:

```suggestion
        if !mapped
            && !remediation.contains(STEP_DOCTOR)
            && !unmapped.iter().any(|seen| seen == remediation)
        {
            unmapped.push(remediation.to_string());
        }
```

Your `next_steps_keep_canonical_order_and_dedupe_unmapped_remediations` test already exercises a remediation containing `STEP_DOCTOR`, but it reaches the mapped path — so it would stay green either way and doesn't cover this.

---

weboko (commented):
## Review + hands-on dogfooding of the two new commits

Reviewed `f0d7c9a`+`ff8374c` (the commits pushed after my 2026-08-11 pass; `7fabccf` was covered then and is unchanged). Provisioned a real LEZ sequencer and ran the scenarios this diff actually routes to, rather than reading the runbook against the code.

**Environment:** Nix 2.35.2, Rust 1.94.1, r0vm 3.0.5 + risc0 Rust toolchain 1.91.1 provisioned via the "Provisioning the Real LEZ Sequencer Toolchain" section. That section is accurate — the documented rzup `InvalidCertificate(UnknownIssuer)` failure reproduced exactly, the `curl` fallback worked verbatim, and both discovered versions matched what it predicts (`risc0-zkvm 3.0.5`, toolchain `r0.1.91.1`). `test-node prepare` → `setup` → `build` → `localnet start` all clean against LEZ `cf3639d8`.

**Suite:** 742 green (552 lib + 182 CLI + 3 testnode + 5 doc), `cargo fmt --check` clean, clippy 62 warnings.

### What I verified working

**1. The doctor next-steps fix is real and the before/after is exactly as described.** Built `origin/master` (`7a68b64`) alongside this branch and ran both against the same un-set-up project:

```
master        PR
------        --
Next steps:   Next steps:
- logos-scaffold doctor
              - run `logos-scaffold basecamp setup`
              - logos-scaffold doctor
```

The `doctor --json` contract shows the same thing (`Install \`nix\`` now reaches the list). All-pass still yields `next_steps: []` (24 PASS on the provisioned project), so the no-op path holds.

**2. The `deploy --json` correction is right, and the old docs were actively wrong.** Against a real sequencer, all four shapes, five programs:

```
deploy --program-path … --json  → {"program":"hello_world","program_id":"0a44769e…","status":"submitted"}
deploy hello_world --json       → {"deploys":[{…}]}
deploy --json                   → {"deploys":[ …5 entries… ]}     (1 line, exit 0)
```

Every documented detail holds: single-line pure JSON on stdout with no echoes or summary block (the `EchoGuard::suppress` + piped subprocess stdio is what earns that claim), `tx` **absent** rather than null, `program_id` present on each entry. I also forced the failure branch by removing one `.bin`:

```
exit=1, object still emitted, failed entry carries "error":"missing program binary" and no program_id
```

— matching the new `skills/lgs-cli` wording exactly. Correcting `lez-template`'s "deploy one program at a time if you need JSON" is the highest-value line in this commit; it was steering agents away from a feature that has existed since #139.

**3. The `#245`/`#246` items from `7fabccf` still hold at runtime.** `run` post-deploy hooks receive `NSSA_WALLET_HOME_DIR` and `LEE_WALLET_HOME_DIR` both set to the same path (the `set_wallet_home_env` routing), and `SCAFFOLD_TOPUP_SKIPPED`/`SCAFFOLD_DEPLOY_SKIPPED` are `0`/`0` on the default profile and `1`/`1` under `--profile self-fund`, with the step keeping its slot and printing a skipped header. `run --help` is 96 chars and does not wrap.

I specifically checked whether `SCAFFOLD_DEPLOY_SKIPPED=1` on a *cache* skip contradicts the docs — it does not: README:258 already says "either `deploy = false`, or the deploy cache found … unchanged". Accurate.

### Findings

Three things to fix, one of which I'd treat as blocking. Inline comments on the two DOGFOODING lines and the doctor branch; the rest here because they sit outside the diff.

---

**A. (blocking) The PR description no longer describes the PR.** The body's Change Summary still says:

> No behavior change — the only non-test, non-doc edits are a helper call that sets the same two variables it set before, and a constant moved between files.

`ff8374c` changes `derive_next_steps`, which changes both `doctor`'s printed output and the `next_steps` array in `doctor --json`. That is a user-visible behavior change to a documented JSON surface. The Verification section is stale in the same direction — it says `cargo test` **738 green** (now 741/742) and, more importantly:

> **DOGFOODING scenarios rerun: none.** This branch changes no runtime behavior…

while `ff8374c`'s own commit message reports running D1–D7, L1–L4, E1–E3, A1 and T1–T4. The checklist above it is ticked against the old body.

This repo treats the PR body as the record — it is what #244/#245/#246 were reviewed against, and it is what the next reader will believe. Please reconcile it with the two new commits before merge. That is the only thing I'd hold this for; the code is fine.

Worth deciding at the same time whether the doctor fix and the `--json` doc drift belong in a PR titled "address the non-blocking review comments left open by #244, #245, #246" — they came from a different activity (a dogfooding sweep), and neither traces to a comment on those three PRs. Splitting is optional, retitling is cheap.

**B. `DOGFOODING.md:167` — the D3 index row still contradicts the D3 body this commit rewrote.**

```
| D3 | `default` | Advanced | Deploy path variations and machine-readable single-program submission | `deploy [program-name]`, `deploy --program-path`, `deploy --program-path --json` |
```

The body now instructs the dogfooder to capture a `{"deploys":[…]}` output from the discovery path, but the index still frames D3 as *single-program* submission and its command surface omits `deploy --json` / `deploy <name> --json` — the two commands the Evidence section now requires. A dogfooder who scopes the run from the index (which is what the index is for) runs the wrong set. This is the same drift the commit exists to remove, one table above the fix. Suggested:

```
| D3 | `default` | Advanced | Deploy path variations and machine-readable submission on both paths | `deploy [program-name]`, `deploy --program-path`, `deploy --program-path --json`, `deploy [program-name] --json` |
```

**C. `DOGFOODING.md:294`/`304` — D2's signal is the exact weakness this commit fixed elsewhere.** D2 currently says:

> - `doctor` returns actionable next steps rather than only raw failures.
> - If `doctor` omits next steps or machine-readable output becomes malformed, treat that as a DX regression.

Both of those **passed on master** while `basecamp doctor` printed nothing but `logos-scaffold doctor`. The section was not omitted and the JSON was not malformed — it was just circular, which is precisely the bug. This is the same shape as the `SCAFFOLD_*_SKIPPED` line you strengthened in D7 ("the line a dogfooder checks against, which until now passed with either var missing"); D2 deserves the same treatment now that there is code to guard. Something like:

> - If `next_steps` consists solely of `logos-scaffold doctor` while some WARN/FAIL row carries a remediation, that is a regression — the re-run step is a trailer, never the whole list.

Since D2 is the scenario that routes here, and the "Change → Scenario" map sends doctor changes to it, this is the assertion that would have caught the bug.

---

Nits, no action needed unless you're in the file: the `unmapped` dedup is against `unmapped` only, so a future remediation that is unmapped *and* embeds `logos-scaffold doctor` would render it twice (inline comment has the detail — no current remediation does this, I checked all of them). And `next_steps` now mixes runnable commands with prose (`Install \`nix\``, `Set [circuits].version to …`); that is a net improvement, but `skills/lgs-cli`'s `doctor --json` line is the place to say so if any consumer treats the array as commands.

Not raised as a finding: the B-series blocker. See the inline comment — I reached a different result than the commit did, and it changes the advice rather than the code.

---

PR #253, opened 2026-08-19
docs: rebuild the public surface around install and first run

weboko:
### Description

After so many iterations, especially using AI `README.md` file became one big pile of stuff.
I want to improve it, drop it into smaller files and make experience reading it pleasant. 

### User Story

A developer arriving at the repo could not get from landing to a running project. The README opened with links to FURPS.md and ADR.md, internal engineering documents. Install sat at line 55, behind a seven-item prerequisites wall, and read cargo install --path ., which assumes a clone the README never mentions. The first runnable command was at line 177 of 437. Measured against Foundry, Anchor, Stellar CLI, Hardhat, and dfx, tagline-to-install runs 12–26 lines; this repo was 51, and tagline-to-first-command was 174 against their 15–25.

Separately, the crates.io page carried a description saying "LSSA", a term with zero remaining occurrences in the README, and a repository URL (logos-co/logos-scaffold) that returns 403.

### Change Summary

Restructures the README around install and first run, moves reference material into docs/, and fixes the two broken Cargo.toml metadata fields. README goes 437 → 280 lines. Nothing is deleted: the command dump and Command Semantics move to [docs/commands.md](https://github.com/logos-co/scaffold/blob/weboko/public-readme/docs/commands.md), run configuration to [docs/configuration.md](https://github.com/logos-co/scaffold/blob/weboko/public-readme/docs/configuration.md), and FURPS/ADR/DOGFOODING pointers to a Project Documents table in CONTRIBUTING.md. Adds AGENTS.md, CODE_OF_CONDUCT.md, issue templates, and a social preview brief.

It also carries a one-line `cargo fmt` fix to `src/process.rs`. That is not documentation, but `cargo fmt --check` is CI's first step and has been failing on master since the setsid daemonize change, so this branch inherited a red run. Happy to split it out if you would rather fix master directly.

### Verification

- `cargo fmt --check`, `cargo check`, `cargo test` pass locally on the merged branch (769 tests).
- Every command signature in `docs/commands.md` re-checked against `--help` from a build of this branch. That turned up the missing `--json` on `localnet logs` / `wallet list` / `wallet topup`, the missing `--timeout-sec` on `test-node start` / `run`, and four basecamp subcommands the page did not list at all (`build`, `run`, `paths`, `docs`).
[...]

PR #256, opened 2026-08-20
feat(basecamp): move the default pin set to Basecamp 0.2.3

weboko:
## User Story

Setting `[repos.basecamp].pin` to 0.2.3 by hand is not enough. Five things break, all silently:

- Scaffold's `lgpm` pin is 24 commits behind the rev basecamp 0.2.3 locks. Basecamp embeds that same library, so the CLI writing the profile and the app reading it disagree.
- That `lgpm` validates content hashes on install. The default `delivery_module` pin predates them and is rejected.
- 0.2.x wraps the dev binary (`bin/LogosBasecamp` execs `bin/.LogosBasecamp`), so `launch` stopped killing the previous instance and left two on one profile.
- 0.2.x resolves its data dir from `AppDataLocation` unless `LOGOS_USER_DIR` is set. Scaffold only set it on the macOS portable stack, so dev-stack profiles collapse onto one tree.
- The bundled-module skip list is the v0.1.1 set: `counter`, `counter_qml`, `webview_app` are gone, `package_downloader` is new.

Surfaced by B1-B6.

## Change Summary

Bumps the three pins that move as a set (basecamp `0.2.3`, `lgpm` to the rev that release locks, `delivery_module` v0.2.0), corrects the bundled-module list, and fixes what the bump exposes: process-identity matching through the wrapper, `LOGOS_USER_DIR` on every host and stack, an install-failure hint for rejected packages, `module_data/` and `logs/` in `basecamp paths`, and a `doctor` row that warns on a split pin pair. 0.1.x stays supported and existing projects keep their own pins until they edit them.

Two follow-up commits fix defects found while verifying the first against a real basecamp build:

- **`resolve_basecamp_binary` probe order broke 0.1.x.** The pin bump reordered the probe to try `bin/LogosBasecamp` first, on the reasoning that it is the entry point on both generations. It is not. Neither generation wraps its Qt binary in place (`wrapQtApps` is skipped so the process name stays `LogosBasecamp` for the macOS Dock), so a `/bin/sh` launcher is what exports `QT_PLUGIN_PATH` / `QML2_IMPORT_PATH` / `LD_LIBRARY_PATH` — and the generations name it differently: 0.1.x's launcher is `bin/logos-basecamp` (its `bin/LogosBasecamp` is the raw ELF), while 0.2.x has no `bin/logos-basecamp` and makes `bin/LogosBasecamp` the launcher over a hidden `bin/.LogosBasecamp`. Confirmed against a built `#app`: the raw binary dies at exec with `libQt6RemoteObjects.so.6: cannot open shared object file`, the launcher reaches `Logos Core started successfully!`. `bin/logos-basecamp` now goes first — it exists only on 0.1.x, so it can never shadow a 0.2.x build. The same generation split left a matching hole in the kill path: 0.1.x's indirection is a *rename*, not a dot-prefix, so `pid_comm_matches` never matched and `launch` silently skipped `kill_process_tree` on every 0.1.x project.
- **`describe_pin_set` warned on an unwritten `[repos.lgpm]`.** `setup` resolves a blank pin to the scaffold default, so a project carrying only `[repos.basecamp]` has two *effective* pins that are exactly the matched defaults. Doctor reported a split that did not exist and prescribed a `setup` re-run that would change nothing.
- **`lgpm_install_hint` was too broad.** It keyed the "predates content hashes" rebuild advice on `Package validation failed`, which is liblgx's banner for *every* structural rejection (the same prefix carries `Manifest: 'name' field is empty`). A malformed package was answered with "rebuild with logos-module-builder 0.2.x", which fixes nothing and buries lgpm's own message. Now matches `content hashes` alone.

## Verification

- 769 tests green (was 742), `cargo fmt --check` clean. New tests cover each behaviour change.
- CI is green on `77ac115`: `validate`, `render-and-build (default)` and `render-and-build (lez-framework)` all pass.
- Pins re-checked against upstream: basecamp `aa237766` is tag `0.2.3`; `202af6fa` is exactly the `logos-package-manager` rev that release's `flake.lock` locks as its root input; `delivery_module` `3258cdb0` is tag `v0.2.0`. The bundled-module list matches `installedDev` in the release's `flake.nix`, and the four base-dir children match `LogosBasecampPaths.h` at that tag.
- The 0.1.x launcher regression above was caught and fixed by running a real built `#app` output, not a stub.
- B1-B6 remain **partial coverage**. Two independent blockers, both recorded in the runbook: a proxy that 403s Nix's GitHub fetchers, and — on a host where that access *is* available — evaluating the 0.2.3 flake (~10k lock nodes, up from ~250 at v0.1.1) exceeding a small container's RAM and dying as a bare `SIGKILL` with no error text.
- Adds **`B7`**, a basecamp-app-free pin-set scenario to run when `B1` is out of reach: it builds both `lgpm` attrs, checks the `.lgx` carries `hashes.root`, runs a real `lgpm install` with scaffold's exact flags, drives both hint paths as negative tests, and checks the resolved entry point is a Qt launcher rather than a raw binary. Every step was executed while validating this PR.
- Still needs a Nix host with enough RAM: `nix build .#app` output layout for 0.2.x, two-profile launch, and macOS `attr = "app"` vs `bin-macos-app`.

## Checklist

- [x] CI is green
- [x] Relevant DOGFOODING scenarios rerun (B1-B6 partial as noted above; B7 added and run in full)
- [x] Docs updated (`README.md`, `DOGFOODING.md`, `docs/basecamp-module-requirements.md`, both skills, `FURPS.md`, new ADR)
- [ ] I have read [CONTRIBUTING.md](/CONTRIBUTING.md) and this PR fits within the weekly contribution cap


weboko (commented):
### Review: move the default pin set to Basecamp 0.2.3

Strong PR — the "pins move as a set" framing is captured well in the new ADR, and the docs (README, DOGFOODING B1–B6, requirements doc, both skills, FURPS) are updated consistently with the code. `cargo test --lib` → 574 pass (173 basecamp).

**Verified on my side (reads + the one code-unverifiable claim):**
- `BASECAMP_PREINSTALLED_MODULES` — I read `logos-basecamp`'s `flake.nix` at the pinned rev `aa237766…`. Its `installedDev` list is exactly five entries (package-manager-module, package-downloader-module, capability-module, main-ui plugin, package-manager-ui plugin), which map to your `capability_module, main_ui, package_downloader, package_manager, package_manager_ui`. The stale `counter`/`counter_qml`/`webview_app`/`basecamp_main_ui` entries are correctly gone. ✅
- **Wrapper process-identity matching:** `basecamp_comm_candidates` returning both the dotted and undotted `comm` twin (truncated to 15) and `pid_comm_matches` accepting any candidate correctly handles 0.2.x's `bin/LogosBasecamp` → `bin/.LogosBasecamp` wrapper, so relaunch kills the prior instance. Fail-closed behavior preserved.
- **`LOGOS_USER_DIR` unconditional / `LOGOS_DATA_DIR` macOS-portable-only** matches the documented AppDataLocation rationale.
- **`lgpm_install_hint`** and the **`describe_pin_set` doctor row** (warn only on a split default pair) are sound and well tested.

**Coverage gap I could not close (same as your "partial coverage" note):** this environment 403s Nix's GitHub fetchers for every repo except `logos-co/scaffold` (Nix pulls `logos-basecamp` + nixpkgs + rust-overlay over `archive/<rev>.tar.gz`, all denied), so I could not run a real `nix build .#app`, a real `lgpm` install (incl. the `Missing content hashes` rejection path), a two-profile launch, or the macOS attr variants. **Recommendation:** these "still needs a Nix host" items should get one real B1–B6 pass on a Nix host (ideally Linux + macoS with both `attr = "app"` and `bin-macos-app`) before merge — the pin correctness is the kind of thing only a real install/launch fully proves.

Code and docs are in great shape; my only ask before merge is that real-Nix confirmation.

---

weboko (commented):
### Re-review of `94401d2` — the real-Nix `B1`–`B5` pass I asked for is now done, and it is green

My previous review left exactly one thing open: *"these 'still needs a Nix host' items should get one real `B1`–`B6` pass … before merge."* I got a Nix host working this round and ran it. **Everything the PR claims about the 0.2.3 pin set holds up against the real build.** No blocking issues from me.

**Environment:** Ubuntu 24.04 x86_64, Nix 2.35.2 (flakes), Rust 1.94.1, r0vm 3.0.5. Cold `nix` store, basecamp's own binary cache (`cache.nix.logos.co/public`) added to `/etc/nix/nix.conf` per your new B1 precondition — cold `basecamp setup` took **5m21s** with it. Module project: `logos-module-builder@464a75d9` `templates/minimal-module` (0.2.x tooling).

<details><summary><b>How I got Nix past the per-repo GitHub allowlist</b> (worth folding into B1 — see finding 3)</summary>

The B1 precondition is right that `api.github.com` answers `403` through the sandbox's intercepting proxy, but "out of reach" turned out to be recoverable: **direct egress bypassing the proxy is open.** Concretely — run the daemon with the proxy vars unset and `no_proxy='*'`, pass `no_proxy='*'` to the client, and put `access-tokens = github.com=$GH_TOKEN` in `nix.conf`. `github:` inputs, `cache.nixos.org` and `cache.nix.logos.co` all resolve after that. All ~10k lock nodes fetched fine.
</details>

---

### Verified against the real 0.2.3 build

**`BASECAMP_PREINSTALLED_MODULES` — now confirmed from `$out`, not just from `flake.nix`.** This is the claim I could only read last time:

```
$ ls app-result/modules        $ ls app-result/plugins
capability_module              main_ui
package_downloader             package_manager_ui
package_manager
```

Five names, byte-for-byte your list. The stale `counter` / `counter_qml` / `webview_app` / `basecamp_main_ui` entries are correctly gone, and `main_ui` is indeed the bundled name (`basecamp_main_ui` never would have matched). Confirmed at runtime too — the launched app loads them straight out of the nix store (`…/logos-basecamp-0.2.3/modules/capability_module/…`), so a freshly seeded profile's empty `modules/` is correct, exactly as your new B1 signal says.

**The wrapper/`comm` fix — reproduced the failure it prevents.** `bin/LogosBasecamp` in the 0.2.3 dev build is a `/bin/sh` wrapper ending in `exec "$BINDIR/.LogosBasecamp" "$@"`, so the live process reports `.LogosBasecamp`. I ran the same relaunch twice, once per binary, against the same 0.2.3 profile:

| relaunch `alice` while an instance is live | result |
|---|---|
| **this PR** | `kill: (12938)`, `kill: (13340…13343)` → prior instance and its three `logos_host_qt` children gone; exactly one `.LogosBasecamp` after |
| **`master`** | no kill attempted → **two `.LogosBasecamp` processes**, six `logos_host_qt` children, both bound to `alice`'s profile tree and `/tmp/lgs-<hash>-alice` sockets |

Silent double-instance on every relaunch, precisely as your doc comment predicts. `basecamp_comm_candidates` fixes it.

**`LOGOS_USER_DIR` unconditional — the Linux dev stack needed it too.** Two simultaneous `launch`es, one process environment each:

```
pid 18119  LOGOS_PROFILE=alice  XDG_RUNTIME_DIR=/tmp/lgs-642ad05a-alice
           LOGOS_USER_DIR=…/profiles/alice/xdg-data/Logos/LogosBasecampDev
pid 18132  LOGOS_PROFILE=bob    XDG_RUNTIME_DIR=/tmp/lgs-642ad05a-bob
           LOGOS_USER_DIR=…/profiles/bob/xdg-data/Logos/LogosBasecampDev
```

`LOGOS_DATA_DIR` correctly absent (not macOS portable) — the narrow gate does what the constant's doc says. Runtime dirs `0700`, distinct, outside the tree; `find .scaffold -type s` empty; per-profile sockets isolated.

**The rest of the series:**

- **`lgpm` pin / content hashes.** Real `lgpm install` of a real 0.2.x-built `.lgx` into both profiles: `done` / `Installed to: …` with only `Warning: Package is unsigned` — the documented `warn` policy running validation and passing. The ADR's writer-and-reader-are-one-library argument is borne out end to end.
- **`basecamp pin set` doctor row.** `PASS | basecamp pin set | basecamp aa237766baf6 / lgpm 202af6fa0f0f — scaffold defaults`. Post-`setup` doctor is the 4 PASS rows plus this one; post-`install` it is `6 PASS, 0 WARN, 0 FAIL` — including the normalized-flake-ref drift check passing on `path:.#lgx` vs the discovered absolute ref.
- **New `paths` fields.** `module_root`, `module_data_dir`, `app_logs_dir` all resolve under `…/Logos/LogosBasecampDev`, and the running app really does populate `module_data/<module>/<id>/` and `logs/basecamp_<ts>.log` there. `B4` scrub removes them along with a planted marker.
- **`B2` capture** byte-identical on re-run; **`B5`** `build` defaults to `all` and emits both `01-minimal.lgx` artefacts.

---

### Findings

**1. `B1`'s idempotency signal no longer matches what a second `setup` prints.** *(docs, non-blocking — but it is this PR's pin that changes it)*

B1 still says: *"Second `basecamp setup` is idempotent: pin unchanged → **no rebuild reported**, exit 0."* Observed on this head:

```
$ time lgs basecamp setup      # second run, pin unchanged
building basecamp (.#app)
  ✓ building basecamp (.#app) (1m14s)
  ✓ building lgpm (…#cli) (0.9s)
setup complete
real 1m16.647s
```

`cmd_basecamp_setup` has no "pin unchanged → skip" path — it always re-invokes `nix build` — so this was always *reported*; what this PR changes is the cost. At the v0.1.1 pin the lock was **253 nodes**; at `aa237766` it is **9991**, so re-evaluation went from near-instant to ~75s. A dogfooder following B1 will read "no rebuild reported" and file the `✓ building basecamp (.#app) (1m14s)` line as a regression. Since the PR already re-opens the whole B series for a pin change, suggest either restating the signal (e.g. *"a second `setup` re-evaluates the flake and re-links — expect ~1 minute at the 0.2.3 lock and no derivation rebuilds — exit 0"*) or making `setup` skip the build when `basecamp.state` already records the same pin and both out-links resolve.

**2. `FURPS.md` Supportability #1 is inaccurate.** *(pre-existing, not caused by this PR — flagging because you edit that block)*

> `logos-scaffold doctor` gains a basecamp section when `.scaffold/basecamp/` exists

It does not. `push_basecamp_doctor_rows` is called only from `basecamp doctor` (`src/commands/basecamp.rs:3212`); the top-level `doctor` in this fully-provisioned basecamp project emitted no basecamp row at all. Either wire it up or reword the requirement.

**3. Suggest recording the proxy bypass in `B1`.** The current text tells an agent to record the blocker and stop at partial coverage — which is what my last review did. The recipe in the collapsed section above turns that into a full pass in the same class of container, and it is the difference between this review and the last one. Worth two sentences in the precondition.

---

### Not covered (unchanged from the PR's own scope note)

- **macOS**, either attr — no host available. `set_module_root_env`'s macOS branches remain unit-test-only, and the `LOGOS_DATA_DIR` / `bin-macos-app` path is still unexercised live.
- **`B3` UI-level p2p.** I verified two-instance isolation headless (`QT_QPA_PLATFORM=offscreen`); no window was driven, so no chat/delivery round-trip.
- **`B6`** is not applicable to `minimal-module` — it exposes no `apps.<system>.default`, so `basecamp run` fails inside `nix run` (`unable to execute …/bin/logos-minimal-module`). Precondition unmet rather than a defect, though a scaffold-side hint for "this module has no standalone app" would beat the raw nix error.

### Recommendation

The pin-set thesis, the bundled-module list, the `comm` fix, the `LOGOS_USER_DIR` widening, the `lgpm`/hash coupling and the new doctor/paths surface are all confirmed against a real 0.2.3 build. My previous round's only blocking ask is satisfied. Finding 1 is a one-line docs fix I'd like to see before merge; findings 2–3 are optional.

Note the red `validate` check on this head is `master`'s pre-existing `cargo fmt` violation at `src/process.rs:632` (from #34), not anything in this PR — #253 carries the fix, so landing that first turns this green.

---

Issue #118, opened 2026-05-08
Hardcoded port-3040 sites beyond #40 (deploy.rs DEFAULT_SEQUENCER_ADDR + 2 substring-match callsites)


wozos:
## Problem

#40 covers `doctor`'s port-check label hardcoding `127.0.0.1:3040`. After that fix, three additional sites still hardcode the same value:

1. **`src/commands/deploy.rs:27`** — `const DEFAULT_SEQUENCER_ADDR: &str = "http://127.0.0.1:3040";`. Used as the fallback when `wallet_config.json` lacks `sequencer_addr`. A project that configured `[localnet].port = 14321` but seeded its wallet config without `sequencer_addr` (or before the port edit) will have `deploy` preflight against the wrong port.
2. **`src/commands/wallet_support.rs:226-241`** (`is_connectivity_failure`) — substring needles include the literal `"127.0.0.1:3040"` and `"localhost:3040"`. For nonstandard-port projects the connectivity heuristic loses one of its signals (it usually still fires via `connection refused` etc., but the literals are an inappropriate coupling to a default that lives elsewhere).
3. **`src/commands/doctor.rs`** — the wallet-config sequencer-address check (separate from #40's port-check check) substring-detects `127.0.0.1:3040` / `localhost:3040` against the raw `wallet_config.json` text instead of parsing the JSON and inspecting the `sequencer_addr` field. Brittle to whitespace / formatting.

## Why it matters

#40 already establishes the port-config-respect invariant. Half-fixing it across only one site keeps users on nonstandard ports stranded on `deploy` and `doctor` paths.

## Proposed solution

- Replace `DEFAULT_SEQUENCER_ADDR` constant with `format!("http://127.0.0.1:{}", project.config.localnet.port)` at call sites in `deploy.rs`. Or, better, derive the addr once in `load_wallet_runtime` from `project.config.localnet.port` when `wallet_config.json` doesn't supply it.
- Drop the `127.0.0.1:3040` / `localhost:3040` literals from `is_connectivity_failure`. The remaining keywords (`connection refused`, `tcp connect error`, `network is unreachable`, etc.) cover the actual failure modes; the address literals are noise.
- In doctor's wallet-config check, parse `wallet_config.json` with `serde_json` and inspect `sequencer_addr` explicitly.

## Source

`code` lens — sweep across `deploy.rs`, `wallet_support.rs`, `doctor.rs`.

## Traceability

- Related: #40
- Run ID: `20260508-181437`
[...]

PR #257, opened 2026-08-21
fix(wallet,deploy): corroborate sequencer connectivity via RPC ErrorKind, not text matching

erhnysr:
Wallet-reported timeouts against a blackholed sequencer (SYN dropped, not refused) were never classified as connectivity failures, because ureq's connect-timeout prose matched none of is_connectivity_failure's text needles. Classifies ErrorKind::ConnectionFailed | Dns structurally in map_ureq_error instead, fixing both the new RPC probe and deploy's existing preflight_sequencer_reachability in one place.

Adds sequencer_connectivity_failure with a two-tier match: fast-path on address + connect phrase, fallback via rpc_get_last_block_id probe
map_ureq_error classifies on ErrorKind rather than scraping transport prose
DOGFOODING.md: reachability notes in D3/D4, original T4 bullet restored
Tests: fake-probe unit suite, real refused-port control, real blackhole (10.255.255.1) regression

Links #118. All 566 lib tests pass, cargo fmt --check clean.

---

### Follow-up (non-blocking, out of scope for this PR)

`preflight_sequencer_reachability`'s warn-and-continue arm interpolates the responder's HTTP body through `one_line()`, which strips newlines but does **not** truncate or escape it. A process squatting on the configured port can therefore write unbounded text into the warning. This PR already moves that warning off `--json` stdout onto stderr (so it can no longer corrupt the JSON stream or be mistaken for forged output), which removes the material harm; what remains is only untidy, length-unbounded text on stderr. A length cap on the interpolated body would close it — deferred here because the cap threshold is a judgment call worth deciding deliberately rather than guessing.


weboko (commented):
Reviewed on a full local environment (Nix 2.35.2, Rust 1.94.1, r0vm 3.0.5 + risc0 guest toolchains, real LEZ sequencer built from the pinned `cf3639d8`). `cargo test --lib` is green on this head: **563 passed, 0 failed**.

The diagnosis is right and the direction is right: keying off the resolved sequencer URL rather than a hard-coded `:3040` is the correct fix, requiring a co-signal matches what `is_localnet_connectivity_failure` already does in `doctor`, and dropping the two address needles from `is_connectivity_failure` is safe (its only remaining caller is `map_ureq_error`, which is fed `ureq::Error::Transport` strings that always carry a transport token).

Three things I verified by running the classifier directly, one of which I think is blocking.

**1. The confirmation-timeout guard does not hold (blocking).** The doc comment promises the exclusion survives "even if that message is later reworded to mention a connection or the URL". It does not — `is_connectivity_failure` is evaluated *first* and returns early, so the guard below it never runs, and that needle list contains two very generic tokens (`"http error"`, `"error sending request"`). Measured on this branch:

```
text = "transaction not found in preconfigured amount of blocks (http error contacting http://127.0.0.1:3040)"
is_confirmation_timeout_failure = true
sequencer_connectivity_failure  = true   <-- guard defeated
```

This matters at the call site: in `wallet.rs` the connectivity branch (L380) pre-empts the dedicated confirmation-timeout branch (L394), so a transaction that *did* reach the sequencer but did not settle reports `sequencer appears unavailable at … Run 'logos-scaffold localnet start'` and bails — the exact misdirection this PR sets out to remove — and the `"status":"pending"` JSON carrying the tx id is lost with it. Swapping the two blocks fixes it and costs nothing.

The existing collision-guard test passes only because `"connection timed out"` happens not to be in the transport list; it pins the guard's *existence* but not its *position*. Worth adding the `http error` variant so the ordering is pinned too.

**2. The `localhost` alias is no longer matched (worth fixing).** The PR cites `doctor`'s implementation as prior art, but that one deliberately accepts **both** `127.0.0.1:{port}` and `localhost:{port}`. The new function matches only the single resolved string, and `sequencer_addr` comes from `wallet_config.json` when that file sets it — so config and wallet output can legitimately disagree on the alias:

```
addr = "http://127.0.0.1:3040"
text = "Error: connection timed out contacting localhost:3040"
sequencer_connectivity_failure = false   <-- old hard-coded needle caught this
```

That is a narrow regression against the pre-change behaviour on the default port, which the PR description says is not intended.

**3. Substring matching is unanchored (nit).** `sequencer_connectivity_failure("… connection timed out contacting http://127.0.0.1:30401", "http://127.0.0.1:3040")` returns `true`, because `127.0.0.1:3040` is a prefix of `127.0.0.1:30401`. `doctor` shares this shape so it is consistent, but the port is right there and matching on a delimiter boundary is cheap. Related: `host_port` keeps any path, so a `wallet_config.json` with `sequencer_addr = "http://127.0.0.1:3040/rpc"` yields the needle `127.0.0.1:3040/rpc`, which wallet output naming only the authority will never contain.

**Docs / DOGFOODING.** No `DOGFOODING.md` update needed — this adds no command surface. Optionally, the T4 note at L1694 ("`wallet topup` failing with a sequencer-unreachable hint means the test-node is not on the wallet's expected port") is precisely the diagnostic this PR makes trustworthy on a custom port, and a half-line there would pay off.

**CI.** `cargo fmt --check` is red on this branch at `src/process.rs:632`, but that is inherited: `src/process.rs` is byte-identical on `origin/master` and this head, so master has been red since the setsid change. Not yours to fix — #253 carries the one-line fix.

Security/privacy: nothing new. The function is pure local string classification over output that is already printed, and `sequencer_addr` is only influenced by local project config.

---

src/commands/wallet_support.rs
```diff
@@ -300,13 +300,59 @@ pub(crate) fn is_connectivity_failure(text: &str) -> bool {
...
+///   "error sending request … operation timed out") are already caught by
+///   [`is_connectivity_failure`] above.
+/// - A confirmation timeout ("transaction not found in preconfigured amount of
+///   blocks") means the sequencer *was* reached but the tx did not settle in
+///   time; it is excluded so the caller's dedicated check still fires, even if
+///   that message is later reworded to mention a connection or the URL.
+pub(crate) fn sequencer_connectivity_failure(text: &str, sequencer_addr: &str) -> bool {
+    if is_connectivity_failure(text) {
+        return true;
+    }
```
weboko, line 359: **Ordering defeats the guard this comment promises.** `is_connectivity_failure` runs first and returns early, so `is_confirmation_timeout_failure` never gets a chance when the reworded message carries one of that list's generic tokens (`"http error"`, `"error sending request"`). Measured on this branch:

```
text = "transaction not found in preconfigured amount of blocks (http error contacting http://127.0.0.1:3040)"
is_confirmation_timeout_failure = true
sequencer_connectivity_failure  = true
```

At the call site that means `wallet.rs` L380 pre-empts L394, so a tx that reached the sequencer but did not settle is reported as "sequencer appears unavailable … run `localnet start`", and the `"status":"pending"` JSON with the tx id is dropped. Hoisting the guard restores the documented contract:

```suggestion
pub(crate) fn sequencer_connectivity_failure(text: &str, sequencer_addr: &str) -> bool {
    if is_confirmation_timeout_failure(text) {
        return false;
    }
    if is_connectivity_failure(text) {
        return true;
    }
```

---

src/commands/wallet_support.rs
```diff
@@ -300,13 +300,59 @@ pub(crate) fn is_connectivity_failure(text: &str) -> bool {
...
+    if is_confirmation_timeout_failure(text) {
+        return false;
+    }
+    let lower = text.to_lowercase();
+    let host_port = sequencer_addr
+        .split_once("://")
+        .map_or(sequencer_addr, |(_, rest)| rest)
+        .trim_end_matches('/')
+        .to_lowercase();
```
weboko: **`localhost` alias dropped versus both the prior art and the pre-change behaviour.** `doctor`'s `is_localnet_connectivity_failure` — cited in the PR description as getting this right — matches `127.0.0.1:{port}` **and** `localhost:{port}`. This matches only the single resolved string, and `sequencer_addr` is taken from `wallet_config.json` when that file sets it, so the config form and the wallet's output form can legitimately differ:

```
addr = "http://127.0.0.1:3040"
text = "Error: connection timed out contacting localhost:3040"
-> false   (the removed hard-coded `localhost:3040` needle used to catch this)
```

Two smaller things in the same expression: the match is an unanchored substring, so a configured `:3040` also matches a message naming `:30401`; and any path survives `trim_end_matches('/')`, so `sequencer_addr = "http://127.0.0.1:3040/rpc"` produces the needle `127.0.0.1:3040/rpc`, which output naming only the authority never contains. Deriving the port and matching both host aliases on a delimiter boundary addresses all three.

---

src/commands/wallet_support.rs
```diff
@@ -549,6 +595,78 @@ mod tests {
...
+        // Collision guard, pinned independently of the confirmation-timeout
+        // wording: even when the message carries BOTH the connect-timeout phrase
+        // and the sequencer URL, a confirmation timeout must not be reclassified
+        // as connectivity — the caller's dedicated check owns it. This fails if
+        // someone drops the `is_confirmation_timeout_failure` early-return.
+        let confirmation = "transaction not found in preconfigured amount of blocks; \
+             connection timed out contacting http://127.0.0.1:3040";
+        assert!(
+            !sequencer_connectivity_failure(confirmation, default_addr),
+            "confirmation timeout must not be reclassified as connectivity"
```
weboko: This test pins that the guard **exists** but not **where it sits**: it passes today only because `"connection timed out"` is absent from `is_connectivity_failure`'s needle list. Deleting the early-return fails it, but moving it below the transport check — which is the state of the code today — still passes.

Adding a variant that uses a token from that list would pin the ordering as well:

```rust
// Same guard, but with a token that `is_connectivity_failure` matches, so
// this also pins that the confirmation-timeout check runs *before* it.
assert!(
    !sequencer_connectivity_failure(
        "transaction not found in preconfigured amount of blocks \
         (http error contacting http://127.0.0.1:3040)",
        default_addr
    ),
    "confirmation timeout carrying a transport token must not be connectivity"
);
```

---

weboko (commented):
### Re-review after the follow-up commit (`002b519`) — all three points addressed ✅

I re-checked `sequencer_connectivity_failure` and the tests against my earlier comments. The fresh commit resolves each one:

1. **Ordering guard hoisted.** `is_confirmation_timeout_failure(text)` now returns `false` *first*, before `is_connectivity_failure`, so a reworded confirmation-timeout carrying a generic transport token (`http error`, `error sending request`) no longer pre-empts the caller's dedicated check. Reproduced the old failure, confirmed it's gone.
2. **`localhost` alias restored + port derivation + boundary anchoring.** The classifier now derives the port from the resolved URL and matches both `127.0.0.1:{port}` and `localhost:{port}`; `sequencer_port` strips scheme and any path so `…:3040/rpc` yields `3040`; and `contains_on_port_boundary` rejects a trailing-digit extension so `:3040` no longer matches `:30401`. All three sub-points covered.
3. **Ordering pinned by test.** The new case uses `http error`, a token `is_connectivity_failure` *does* match, so merely moving the guard below the transport check (not just deleting it) now fails the test.

**Verification on my side:** `cargo test --lib sequencer_connectivity` passes; the four deploy/topup call sites are correctly routed through the new classifier with the resolved `sequencer_addr` (which carries the configured `[localnet] port` and any `wallet_config.json` override). `cargo fmt --check` is clean on all files this PR touches — the only remaining fmt failure is the pre-existing `src/process.rs:632` from #34, which this PR does not touch (and which #253 fixes), so your CI note is accurate.

Also ran the shared default-template `deploy` / `wallet topup` happy path end-to-end against a real sequencer — no false "sequencer unreachable" classification. Looks good to me.

---

src/commands/wallet_support.rs
```diff
@@ -300,13 +301,167 @@ pub(crate) fn is_connectivity_failure(text: &str) -> bool {
...
+    let authority = authority.split(['/', '?', '#']).next().unwrap_or(authority);
+
+    let Ok(mut addrs) = authority.to_socket_addrs() else {
+        return true;
+    };
+    let Some(addr) = addrs.next() else {
+        return true;
+    };
+
+    TcpStream::connect_timeout(&addr, Duration::from_millis(300)).is_ok()
```
weboko: **The probe answers "is a port open", but the hint it gates claims "our sequencer is serving".**

Those differ exactly where `sequencer_unreachable_hint` says they do — its own third line is *"Another project's sequencer may already be running and may not match this project."* Reproduced with a plain `TcpListener` on `:3040` that accepts and never replies:

```
$ lgs wallet topup
error: wallet topup failed while checking account initialization: Error: Request timeout
Hint: verify the destination with `logos-scaffold wallet -- account get ...`
```

Handshake completes → "reachable" → hint suppressed → user debugs their account instead of the port. (`master` behaves the same, so not a regression — just the case this machinery is best placed to fix.)

The better corroborator already lives in this file: `rpc_get_last_block_id(sequencer_addr)` returns `RpcReachabilityError::Connectivity(..)`, is what `deploy`'s `preflight_sequencer_reachability` (`deploy.rs:469`) uses, verifies the sequencer actually answers `getLastBlockId`, and reuses `map_ureq_error`. It needs no new socket code.

src/commands/wallet_support.rs
```diff
@@ -300,13 +301,167 @@ pub(crate) fn is_connectivity_failure(text: &str) -> bool {
...
+/// actually-unreachable one — so this settles it the way message-scraping
+/// cannot: if the TCP handshake completes, something is listening, whatever
+/// the wallet's timeout means it is not "unreachable".
+///
+/// Uses a short, fixed timeout (this only runs after the wallet has already
+/// failed, so a few hundred extra milliseconds is an acceptable cost for a
+/// correct classification). If the address can't be parsed or resolved, this
+/// returns `true` (assume reachable) so a wallet-side address quirk is never
+/// misreported as our own connectivity problem.
+fn probe_sequencer_reachable(sequencer_addr: &str) -> bool {
```
weboko: **This reimplements `port_open`.** `src/process.rs:744` is already `pub(crate) fn port_open(addr: &str) -> bool` doing `TcpStream::connect_timeout`, used in eight places (`doctor_checks.rs:133`, `testnode/mod.rs:271`, `localnet.rs` x6). This is a ninth implementation that diverges on every axis:

| | timeout | on unparseable | DNS |
| --- | --- | --- | --- |
| `port_open` | 500 ms | `false` (fail closed) | no |
| `rpc_get_last_block_id` | 1 s connect | — | via ureq |
| this | **300 ms** | `true` (fail open) | **yes** |

300 ms is now the most aggressive timeout in the repo for this question — a legitimate handshake that exceeds it produces a false "sequencer appears unavailable". Please reuse one primitive rather than adding a third set of semantics, and keep the fail-open choice explicit at the call site since it inverts `port_open`'s.

src/commands/wallet_support.rs
```diff
@@ -300,13 +301,167 @@ pub(crate) fn is_connectivity_failure(text: &str) -> bool {
...
+    let authority = sequencer_addr
+        .split_once("://")
+        .map_or(sequencer_addr, |(_, rest)| rest);
+    let authority = authority.split(['/', '?', '#']).next().unwrap_or(authority);
+
+    let Ok(mut addrs) = authority.to_socket_addrs() else {
```
weboko: Two edges in this line specifically.

**Userinfo silently disables the probe.** `to_socket_addrs` is handed `user:pass@127.0.0.1:3040`, fails to resolve, and returns `true` — "reachable" — with nothing listening:

```
probe(http://127.0.0.1:35199)           = false   # correct, nothing listening
probe(http://user:pass@127.0.0.1:35199) = true    # wrong, same dead port
```

Not hypothetical here: `report.rs:1238` redacts `user@host` out of sequencer URLs precisely because credential-bearing URLs occur.

**The DNS leg is unbounded.** The doc comment above promises "a short, fixed timeout … a few hundred extra milliseconds", but 300 ms bounds only `connect_timeout`. Blocking `getaddrinfo` on a hostname `sequencer_addr` (reachable — it comes from `wallet_config.json`) has no bound. I could not get it to hang here (unresolvable names returned in 2–8 ms), so this is the design and the comment's promise, not a reproduced bug. `port_open`'s parse-only approach avoids both.

src/commands/wallet_support.rs
```diff
@@ -300,13 +301,167 @@ pub(crate) fn is_connectivity_failure(text: &str) -> bool {
...
+    };
+
+    TcpStream::connect_timeout(&addr, Duration::from_millis(300)).is_ok()
+}
+
+/// Extract the port from a resolved sequencer URL, dropping the scheme and any
+/// path/query/fragment so `http://127.0.0.1:3040/rpc` yields `3040` — wallet
+/// output naming only the authority never carries the path, so keeping it would
+/// poison the needle.
+fn sequencer_port(sequencer_addr: &str) -> Option<String> {
```
weboko: These four lines are verbatim identical to `probe_sequencer_reachable` above (strip scheme, cut at `/?#`). `report.rs:1215-1236` has a third variant and `config.rs:706` a fourth. Worth one shared helper — which is also where userinfo stripping would live once, fixing the edge flagged above for both callers at the same time.

weboko (commented):
### Re-review after `e26608a` (the socket-probe commit)

Rebuilt the full environment from scratch for this pass — Nix 2.35.2 (flakes), Rust 1.94.1, r0vm 3.0.5 via the DOGFOODING curl fallback (rzup still dies on `InvalidCertificate(UnknownIssuer)`, as documented), risc0 guest toolchain 1.91.1, and a real LEZ sequencer built from the pinned `cf3639d8`. `cargo test --lib` → **565 passed / 0 failed**. `cargo clippy --all-targets` introduces **no new warnings** (the two candidates are pre-existing in `basecamp.rs` / `report.rs`).

The new commit changes the shape of the fix substantially — from a text matcher to something that opens a socket — so I re-ran the scenarios the rerun guidance maps to (`D3`+`D6` for deploy, `D4` for wallet flows), plus `D1` as the bootstrap prerequisite, against a real sequencer.

**DOGFOODING results — all green:**

| Scenario | Result |
| --- | --- |
| `D1` | `new` → `setup` (wallet seeded from preconfigured account) → `localnet start` (ready, pid) → `build` → `deploy` **5/5, 0 failed** → `wallet topup` exit 0 → `wallet -- check-health` `✅All looks good!` |
| `D3` | `deploy hello_world` OK; `--program-path --json` → single flat object, pure JSON, `program_id` present; `deploy nonexistent_program` lists the 5 discovered programs; localnet-down path prints the sequencer-unavailable hint |
| `D4` | `wallet list`, `topup --dry-run`, `topup`, `wallet -- account list/check-health` all behave as documented |
| `D6` | Both runners submitted real txs; account A `data` → `Hola mundo!`, account B `data` → `dogfood-test-message`, distinct `program_owner`s, exit 0 |

---

### The central claim checks out, and the probe demonstrably earns its keep

I verified the load-bearing premise independently rather than taking the comment's word for it:

- `jsonrpsee-core-0.26.0/src/client/error.rs:55` → `#[error("Request timeout")]`. The wallet reaches this through `sequencer_service_rpc` (`client` feature). So yes — the pinned wallet really does emit a bare `Error: Request timeout` naming neither the connection nor the address, and no text matcher can classify it.
- I reproduced the SYN-blackhole case for real (`iptables -I INPUT -p tcp --dport 3040 -j DROP`) and A/B'd the two binaries:

```
master:  error: wallet topup failed while checking account initialization: Error: Request timeout
         Hint: verify the destination with `logos-scaffold wallet -- account get ...`   ← misleading

PR 257:  error: wallet topup failed during account preflight: Error: Request timeout
         sequencer appears unavailable at http://127.0.0.1:3040
         Run `logos-scaffold localnet start`.                                            ← correct
```

That is a genuine, user-visible improvement, and it is not reachable by any amount of message-scraping. The `is_confirmation_timeout_failure` hoist from `002b519` also still holds. Nice work.

I also confirmed the four old call sites are fully migrated, with `is_connectivity_failure` surviving only as the transport-token matcher for `map_ureq_error` — exactly as the description says.

---

### The main thing I'd like changed: the probe answers the wrong question

`probe_sequencer_reachable` asks *"is anything listening on this port?"*. The hint it gates asks *"is **our sequencer** serving?"*. Those differ precisely in the case `sequencer_unreachable_hint` was written for — its own third line reads **"Another project's sequencer may already be running and may not match this project."**

Reproduced end-to-end with a plain `TcpListener` on `:3040` that accepts and never replies (i.e. a foreign process squatting the port):

```
$ lgs wallet topup            # PR 257, non-sequencer listener on :3040
error: wallet topup failed while checking account initialization: Error: Request timeout
Hint: verify the destination with `logos-scaffold wallet -- account get --account-id Public/Cbg...`
```

The TCP handshake completes, the probe reports "reachable", the hint is suppressed, and the user is sent to debug their *account* when the actual fault is the port. To be fair: **master does the same thing here, so this is not a regression** — it's the one case the new machinery is best placed to fix and currently doesn't.

The fix is already in this file. `rpc_get_last_block_id(sequencer_addr)` returns `RpcReachabilityError::Connectivity(..)` and is what `deploy`'s own `preflight_sequencer_reachability` (`deploy.rs:469`) uses — I watched it produce the correct hint on the localnet-down `D3` run. It verifies the sequencer actually answers `getLastBlockId`, reuses `map_ureq_error`, and needs no new socket code at all.

### Second: this reimplements `port_open`

`src/process.rs:744` is already `pub(crate) fn port_open(addr: &str) -> bool` doing `TcpStream::connect_timeout`, used in eight places (`doctor_checks.rs:133`, `testnode/mod.rs:271`, `localnet.rs` ×6). The new probe is a ninth implementation that diverges on all three axes:

| | timeout | on unparseable | DNS |
| --- | --- | --- | --- |
| `port_open` | 500 ms | `false` (fail closed) | no — `SocketAddr::parse` |
| `rpc_get_last_block_id` | 1 s connect | — | via ureq |
| `probe_sequencer_reachable` | **300 ms** | `true` (fail open) | **yes** — `to_socket_addrs` |

300 ms is now the most aggressive timeout in the repo for this question; a legitimate handshake that exceeds it yields a false "sequencer appears unavailable". Whichever primitive you land on, please reuse one rather than adding a third set of semantics — and keep the fail-open choice explicit at the call site, since it's the opposite of `port_open`'s.

### Two concrete edges in the current probe

**Userinfo silently disables it.** `to_socket_addrs` gets handed `user:pass@127.0.0.1:3040`, fails to resolve, and the function returns `true` — "reachable" — with nothing listening. Demonstrated:

```
probe(http://127.0.0.1:35199)                 = false   ← correct, nothing listening
probe(http://user:pass@127.0.0.1:35199)       = true    ← wrong, same dead port
```

Not hypothetical for this codebase: `report.rs:1238` redacts `user@host` out of sequencer URLs precisely because credential-bearing URLs occur.

**The DNS leg is unbounded.** The doc comment promises "a short, fixed timeout … a few hundred extra milliseconds", but the 300 ms bounds only `connect_timeout`. Blocking `getaddrinfo` on a hostname `sequencer_addr` — which is reachable, since `sequencer_addr` comes from `wallet_config.json` — has no bound at all. I did not get it to hang here (unresolvable names returned in 2–8 ms), so I'm flagging the design and the comment's promise, not a reproduced bug. `port_open`'s parse-only approach sidesteps this entirely.

### Scope worth stating in the PR

The ordinary "sequencer is down" case never reaches the probe. I captured the wallet's real stderr with localnet stopped:

```
Error: client error (Connect)
Caused by:
    0: tcp connect error
    1: Connection refused (os error 111)
```

`tcp connect error` is matched by `is_connectivity_failure` at step 2, so the classifier short-circuits (0.11 s, no probe). The probe only decides SYN-blackhole / frozen-process cases. That's a real case worth fixing, but it bounds how much risk the probe should be allowed to introduce — which is my argument for reusing an existing primitive rather than adding new semantics.

### Consistency

`doctor::is_localnet_connectivity_failure` (`doctor.rs:482`) is text-only and requires address **and** transport token. This one is probe-based and requires neither. The description cites `doctor` as the model, but the two have now diverged into different answers to the same question. Worth either converging them or noting in the doc comment why they intentionally differ.

Also on the two-tier logic: every `CONNECT_TIMEOUT_PHRASES` entry contains `timeout`/`timed out`, so the fast path only ever changes the answer when the probe would have said "reachable". Your own test pins the asymmetry — `other_port` + reachable → not connectivity, our-port + reachable → connectivity. Defensible (explicit text about our address is stronger evidence), but it deserves one line saying it's deliberate.

### DOGFOODING.md landed in the wrong scenario

The single doc edit is at line 1694, inside **`T4`. Real Committed User Transaction**. But the rerun guidance for this change points at **`D3`+`D6`** ("Changes to deploy behavior…", line 1712) and **`D4`** ("Changes to wallet flows…", line 1713). Two failure signals now depend on probe behaviour and are unchanged:

- `D3` line 358 — "If localnet is unreachable, deploy should fail with a sequencer-unavailable hint instead of a vague wallet error."
- `D4` — "Connectivity failures during topup should mention localnet/sequencer reachability rather than only raw wallet output."

Both are exactly what the foreign-listener case above breaks, and a dogfooder following either would have no way to know a live probe is now involved. The sentence you did add also describes only the port derivation from `5dbcdab` — it says nothing about the socket probe from `e26608a`.

Relatedly, the PR body says "Applicable DOGFOODING deploy/wallet scenarios exercise the same error paths" without naming `D3`/`D4`/`D6` or reporting results; CONTRIBUTING asks for "Applicable DOGFOODING scenarios rerun". My run above is green, so this is a bookkeeping fix, not a blocker.

### Nits

- ~60 lines of doc comment for ~25 lines of code, carrying review-thread narrative into permanent source ("observed in review of PR #257", "see PR #257 review, reproduced against a real SYN-blackholed sequencer"). The invariant belongs in the doc comment; the archaeology belongs in the commit message.
- `sequencer_connectivity_failure_corroborates_bare_timeout_with_socket_probe` binds port 0, drops the listener, then probes — the ephemeral port can be reassigned in between. Rare, but it's a real flake window.
- `contains_on_port_boundary` I checked closely: panic-free and terminating (`find` returns byte offsets that land on char boundaries; `start` advances by at least `needle.len()`). No issue — noting it so you know it was looked at.

---

### Summary

The diagnosis, the direction, and the SYN-blackhole fix are all correct and independently verified, and the full `D1`/`D3`/`D4`/`D6` set is green against a real sequencer on this head. I'm not approving only because of the foreign-listener gap plus the duplicated reachability primitive — both resolve the same way, by corroborating with `rpc_get_last_block_id` (or at minimum `port_open`) instead of a new raw socket, which would also delete the userinfo and DNS edges for free. The DOGFOODING relocation to `D3`/`D4` should ride along.

CI note: your read is accurate — `cargo fmt --check` fails at `src/process.rs:632` identically on `master` (I swapped the file between branches to confirm), and this PR doesn't touch that file. Note the `CI` run on `e26608a` is currently `action_required`, so nothing has actually executed on this head; the results above are all local.

---

DOGFOODING.md
```diff
@@ -1691,7 +1691,7 @@ HEAD0=$("$SCAFFOLD_BIN" test-node blocks head --url "$URL" --json | jq -r .block
 
 ### Failure Signals / Common Pitfalls
 
-- `wallet topup` failing with a sequencer-unreachable hint means the test-node is not on the wallet's expected port — start it with `--port <localnet.port>`.
+- `wallet topup` failing with a sequencer-unreachable hint means the test-node is not on the wallet's expected port — start it with `--port <localnet.port>`. The hint keys off the resolved `sequencer_addr` (from `wallet_config.json` when that file sets it, otherwise the configured `[localnet] port`), not a hardcoded `:3040`; if it fires while the sequencer *is* up on that resolved port, treat it as a real connectivity fault, not a port mismatch.
```
weboko: This edit landed inside **`T4`. Real Committed User Transaction**, but the rerun guidance for this change points at **`D3`+`D6`** ("Changes to deploy behavior…", line 1712) and **`D4`** ("Changes to wallet flows…", line 1713). Two failure signals now depend on probe behaviour and are unchanged:

- `D3` line 358 — "If localnet is unreachable, deploy should fail with a sequencer-unavailable hint instead of a vague wallet error."
- `D4` — "Connectivity failures during topup should mention localnet/sequencer reachability rather than only raw wallet output."

Both are exactly what the foreign-listener case breaks, and a dogfooder reading either has no way to know a live TCP probe is now involved. The sentence added here also describes only the port derivation from `5dbcdab` — nothing about the socket probe from `e26608a`.

src/commands/wallet_support.rs
```diff
@@ -300,13 +300,134 @@ pub(crate) fn is_connectivity_failure(text: &str) -> bool {
         "network is unreachable",
         "error sending request",
         "http error",
```
weboko, line 302: **Removing these two needles regresses `deploy`'s preflight, and is why the new probe can't see a blackholed sequencer.**

`is_connectivity_failure` is also called by `map_ureq_error` (L628), which is how `preflight_sequencer_reachability` decides between a clean bail and "continue anyway". `ureq` embeds the request URL in its `Display`, so on `master` the `127.0.0.1:3040` needle matched *every* transport error against the default port. Measured, blackholed sequencer on 3040:

```
ureq msg: "http://127.0.0.1:3040/: Connection Failed: Connect error: connection timed out"
  master: is_connectivity_failure = true  -> Connectivity -> deploy bails with the unreachable hint
  PR #257: is_connectivity_failure = false -> Other        -> "warning: probe failed; continuing"
```

The same miss makes `probe_sequencer_reachable` return `true` for a blackholed sequencer, so `sequencer_connectivity_failure` never fires.

Removing the needles is right — they were a hardcoded-port hack. The fix is to stop asking prose for something `ureq` reports structurally. In `map_ureq_error`:

```rust
ureq::Error::Transport(transport) => {
    let msg = transport.to_string();
    // ureq reports both "connection refused" and a connect *timeout* as
    // ConnectionFailed; its prose ("Connect error: connection timed out")
    // matches none of the needles below, and a blackholed sequencer is
    // exactly the case the connectivity hint exists for.
    match transport.kind() {
        ureq::ErrorKind::ConnectionFailed | ureq::ErrorKind::Dns => {
            RpcReachabilityError::Connectivity(msg)
        }
        // Io covers a read timeout against a listener that *did* accept, so
        // it stays "something answered" and a slow sequencer is not
        // misreported as unreachable.
        _ if is_connectivity_failure(&msg) => RpcReachabilityError::Connectivity(msg),
        _ => RpcReachabilityError::Other(msg),
    }
}
```

That fixes the blackhole case and restores the `deploy` preflight in one place, and leaves `is_connectivity_failure` doing what it is actually good at: reading wallet *subprocess* text.

---

src/commands/wallet_support.rs
```diff
@@ -300,13 +300,134 @@ pub(crate) fn is_connectivity_failure(text: &str) -> bool {
...
+///
+/// `Connectivity` errors are treated as unreachable; anything else (bad
+/// JSON, wrong RPC surface, an HTTP status) means something answered, which
+/// is not a connectivity failure — mirrors `preflight_sequencer_reachability`,
+/// which continues rather than failing on non-connectivity RPC errors.
+fn probe_sequencer_reachable(sequencer_addr: &str) -> bool {
+    match rpc_get_last_block_id(sequencer_addr) {
+        Ok(_) => true,
+        Err(RpcReachabilityError::Connectivity(_)) => false,
+        Err(RpcReachabilityError::Other(_)) => true,
```
weboko, line 409: **`Other => true` is what swallows the case this PR exists for.**

A connect timeout lands here as `Other`, not `Connectivity` (see my note on L302), so this returns `true` and the caller concludes the sequencer is fine. Verified against the real functions:

```
rpc_get_last_block_id("http://10.255.255.1:3040")
  = Other("… Connection Failed: Connect error: connection timed out")   <-- reported REACHABLE
sequencer_connectivity_failure("Error: Request timeout", "http://10.255.255.1:3040") = false
```

The mapping itself is defensible once `map_ureq_error` classifies connect failures correctly — I'd fix it there rather than here, so `preflight_sequencer_reachability` benefits too.

Worth noting the doc comment above claims this asks "whether *our* sequencer answers RPC, not merely whether some process has the port open". With `Other => true` that isn't quite true: a foreign HTTP 404, a silent `accept()`, and raw garbage bytes all land in `Other` and read as reachable, so in practice the probe behaves like a port-open check for every squatter shape I tested. Either soften the comment or make the probe `matches!(rpc_get_last_block_id(addr), Ok(_))`.

---

src/commands/wallet_support.rs
```diff
@@ -549,6 +670,199 @@ mod tests {
...
+    // Port 1 is a privileged port with nothing listening in this test
+    // environment (matches the existing convention in
+    // `rpc_get_last_block_id_returns_connectivity_error_when_unreachable`
+    // below) — a connection to it is refused immediately, so this is
+    // deterministic and fast, no need to actually wait out a timeout. Unlike
+    // the classification test above, this one exercises the REAL probe
+    // end-to-end via the public `sequencer_connectivity_failure`.
+    #[test]
+    fn sequencer_connectivity_failure_corroborates_bare_timeout_with_rpc_probe() {
+        let addr = "http://127.0.0.1:1";
```
weboko, line 904: **This test is named for the timeout case but exercises the refused case, which is why the suite is green while the real scenario fails.**

Port 1 gives `Connection refused`, as the comment above acknowledges ("no need to actually wait out a timeout"). Refused was already classified as connectivity before this PR via the generic transport needles — so this asserts a path that never needed the probe, and never touches the connect-*timeout* path the probe was added for.

A blackholed address reproduces finding 1 as a red test, and costs only the configured 1s connect timeout:

```rust
// 10.255.255.1 is RFC1918 and unrouted here, so SYNs are dropped rather
// than refused: this is a real connect *timeout*, bounded by the agent's
// 1s timeout_connect. Refused (port 1) does not exercise this path — it
// was already classified before the probe existed.
let addr = "http://10.255.255.1:3040";
```

Keep the port-1 case too; it is a good control. They fail differently, which is the point.

---

DOGFOODING.md
```diff
@@ -356,6 +356,7 @@ Both deploy paths honor `--json`, with a different shape each. `--program-path -
 
 - If either `--json` path starts emitting a guaranteed-null `tx` key, or mixes command echoes and the human-readable summary into the JSON stream, record that as a machine-readability regression.
 - If localnet is unreachable, deploy should fail with a sequencer-unavailable hint instead of a vague wallet error.
+- That hint is corroborated via an RPC call (`getLastBlockId`) against the resolved `sequencer_addr`, not merely an open TCP port — a foreign process squatting on the configured port must not suppress it.
```
weboko: **This promises a guarantee the code does not currently deliver.** I tested every squatter shape against the real classifier, and all three read as reachable, so the hint *is* suppressed:

```
foreign HTTP server (404) on the port    -> REACHABLE (Status → Other)
foreign TCP squatter (accept, no reply)  -> REACHABLE (Unexpected EOF → Other)
foreign non-HTTP squatter (garbage)      -> REACHABLE (Bad Status → Other)
```

End-to-end with the real wallet and a silent squatter on 3040, `wallet topup` printed `Error: Request timeout` and no unreachable hint.

If you keep `Other => reachable` (reasonable — it mirrors `preflight_sequencer_reachability` and tolerates a version-skewed sequencer), the runbook line should describe what the probe actually distinguishes:

```suggestion
- That hint is corroborated by an RPC call (`getLastBlockId`) against the resolved `sequencer_addr` rather than by wallet wording alone. The probe separates "refused/unreachable" from "something answered" — it does not verify the responder is *our* sequencer, so a foreign process squatting on the configured port will still suppress the hint.
```

Same applies to the matching line at L422.

---

weboko (commented):
### Re-review of `7b6f0d2` on a fully provisioned environment — dogfooding is green, but the fix still does not fire

Environment: Nix 2.35.2, Rust 1.94.1, r0vm 3.0.5 + risc0 guest toolchain `r0.1.91.1`, real LEZ `sequencer_service` and `wallet` built from the pinned `cf3639d8`, circuits v0.4.1. `cargo test --lib` on this head: **565 passed, 0 failed**; `cargo fmt --check` clean on every file this PR touches.

The full happy path is green on this head, with a real sequencer executing real transactions: `new` → `setup` → `localnet start` → `build` → `wallet topup` (real pinata claim, r0vm-executed) → `deploy` (5/5 programs submitted) → `deploy <name>` → `deploy --program-path --json` → `cargo run --bin run_hello_world`, with account `data` decoding to `Hola mundo!` and a non-null `program_owner`. Nothing this PR changes broke any of that.

Moving from `CONNECT_TIMEOUT_PHRASES` to a real RPC probe is the right direction, and dropping the socket code in favour of `rpc_get_last_block_id`/`RpcReachabilityError` addressed my last round cleanly. But I ran the classifier against the real wallet again, and the outcome is unchanged from `master`.

---

#### 1. Blocking: the probe reports a blackholed sequencer as *reachable*, so the branch still never fires

Same reproduction as my previous round — an unreachable sequencer whose SYNs are dropped rather than refused (firewall, wrong host, host down), on the configured port:

```
$ jq '.sequencer_addr = "http://10.255.255.1:3040"' .scaffold/wallet/wallet_config.json   # blackholed, port 3040

=== MASTER ===
error: wallet topup failed while checking account initialization: Error: Request timeout
Hint: verify the destination with `logos-scaffold wallet -- account get --account-id Public/Cbg…`.

=== PR #257 (7b6f0d2) ===
error: wallet topup failed while checking account initialization: Error: Request timeout
Hint: verify the destination with `logos-scaffold wallet -- account get --account-id Public/Cbg…`.
```

Byte-identical. No "sequencer appears unavailable" hint — the exact case the PR sets out to fix.

The root cause is one layer below the code you changed. `ureq` reports a connect timeout as `ErrorKind::ConnectionFailed`, and its `Display` is:

```
http://10.255.255.1:3040/: Connection Failed: Connect error: connection timed out
```

None of `is_connectivity_failure`'s needles match that string — `"connecterror"` has no space, and `"tcp connect error"` is a different phrasing. So `map_ureq_error` returns `Other`, and `probe_sequencer_reachable` maps `Other → true`. Confirmed by instrumenting the real functions:

```
REVIEW: rpc_get_last_block_id("http://10.255.255.1:3040") = Other(… Connect error: connection timed out)  <-- probe reports REACHABLE
REVIEW: sequencer_connectivity_failure("Error: Request timeout", "http://10.255.255.1:3040") = false
REVIEW: is_connectivity_failure("… Connection Failed: Connect error: connection timed out") = false
```

`connection refused` still works (I verified the hint appears with localnet stopped), but that already worked on `master` via the generic transport needles. The blackhole case — the only one the new branch exists for — is exactly the one the probe cannot see.

#### 2. Regression: removing the `:3040` needles degrades `deploy`'s preflight

`is_connectivity_failure` is not only used by your new code — `map_ureq_error` calls it, which is how `preflight_sequencer_reachability` decides between a clean bail and "continue anyway". Dropping `"127.0.0.1:3040"` / `"localhost:3040"` changes that path for a blackholed sequencer on the default port:

| | `is_connectivity_failure(ureq msg)` | result |
|---|---|---|
| `master` | `true` (the `127.0.0.1:3040` needle matched the URL ureq embeds) | `Connectivity` → `deploy` bails with the unreachable hint |
| this PR | `false` | `Other` → `warning: sequencer reachability probe failed …; continuing with wallet submission mode`, then a confusing downstream failure |

No test covers this, which is why the suite stayed green.

#### Both are fixed by one change: classify on `ErrorKind`, not on prose

`ureq` already tells us structurally what happened, so the string sniffing can go away for the cases that matter. Measured kinds:

```
http://10.255.255.1:3040   -> kind=ConnectionFailed  (connect timeout)
http://127.0.0.1:1         -> kind=ConnectionFailed  (connection refused)
http://no-such-host.invalid-> kind=Dns
```

Suggested change in `map_ureq_error` — this makes the blackhole case classify, restores the `deploy` preflight behaviour, and keeps `is_connectivity_failure` as the fallback for wallet *subprocess* text (its real job). `Io` is deliberately left out so a reachable-but-slow sequencer that trips the 2s read timeout still counts as reachable, keeping your `does_not_misclassify_a_reachable_but_slow_sequencer` test honest.

#### 3. The new test is named for the timeout case but exercises the refused case

`sequencer_connectivity_failure_corroborates_bare_timeout_with_rpc_probe` uses `http://127.0.0.1:1`, which is *refused*, not timed out — as its own comment says ("no need to actually wait out a timeout"). Refused was already classified before this PR. That substitution is precisely why the suite is green while the real scenario fails. A test against a blackholed address (`http://10.255.255.1:3040` — RFC1918, unrouted, ~1s to the configured connect timeout) reproduces finding 1 as a red test today.

#### 4. `DOGFOODING.md` promises a squatter guarantee the code does not deliver

Both added lines claim a foreign process on the configured port "must not suppress" the hint. I tested all three squatter shapes against the real classifier; every one is treated as reachable, so the hint *is* suppressed:

```
foreign HTTP server (404) on the port    -> REACHABLE (Status → Other)
foreign TCP squatter (accept, no reply)  -> REACHABLE (Unexpected EOF → Other)
foreign non-HTTP squatter (garbage)      -> REACHABLE (Bad Status → Other)
```

End-to-end with the real wallet and a silent squatter on 3040: `wallet topup` printed `Error: Request timeout` with no unreachable hint.

`Other → reachable` is a defensible, conservative choice (it mirrors `preflight_sequencer_reachability`, and it protects against a version-skewed sequencer that renames the method). If you keep it, please soften the two runbook lines to match — the probe distinguishes *refused* from *answering*, not *ours* from *someone else's*. If the squatter guarantee is what you actually want, `probe_sequencer_reachable` has to be `matches!(rpc_get_last_block_id(addr), Ok(_))`, and the doc comment's "not merely whether some process has the port open" becomes true. Either is fine — they just need to agree.

---

#### Non-blocking notes

- **`deploy`'s new call site is nearly unreachable.** `preflight_sequencer_reachability` runs at `deploy.rs:147` and bails on `Connectivity` before the loop, so by the time `sequencer_connectivity_failure` runs at `:239` the sequencer answered RPC moments ago. It only earns its keep if the sequencer dies mid-deploy — worth a one-line comment, since it also costs up to ~3s per failing program.
- **`contains_on_port_boundary` anchors only the right edge.** `…:3040` won't match `…:30401`, but `foo127.0.0.1:3040` still matches. Harmless in practice; noting it so the asymmetry is deliberate.
- **Doc comments are load-bearing and good** — the `doctor` divergence rationale in particular is the kind of thing that stops a future reader from "unifying" the two. No objection to the length.
- **Security/privacy:** nothing new. The probe POSTs to the same `sequencer_addr` `deploy`'s preflight already contacts, carries no credentials, and adds no new outbound surface.

#### Process notes, not yours to fix

- CI has still never run on this PR — all five runs for `fix/sequencer-connectivity-by-port` sit at `action_required` (the fork first-time-contributor gate). A maintainer needs to click **Approve and run workflows**.
- `master` is *still* red on `cargo fmt --check` (`src/process.rs:632`, inherited from the setsid daemonize change); #253 carries the fix but has not landed. `validate` will go red here for that reason alone, independent of your diff.

---

src/commands/wallet_support.rs
weboko: Re-measured on `7cd8cba`: **the case I reported here is fixed**, but one sub-case of it survives.

Fixed — connect timeouts now classify structurally, and it is load-bearing. Blackholed sequencer, `deploy`:

```
master : warning: sequencer reachability probe failed (Connect error: connection timed out); continuing
         -> wasted wallet submission -> "Error: Transaction submission error / inspect sequencer logs and retry"
#257   : error: cannot deploy programs: ... Connect error: connection timed out
         sequencer appears unavailable at http://10.255.255.1:3051
```

Remaining: `ErrorKind::Io` still falls through to the text fallback, which no longer has the URL needle — so a **foreign listener that accepts and never replies** now loses the hint on the default port. Measured, squatter on `127.0.0.1:3040`, `deploy hello_world`:

```
master : error: cannot deploy programs: http://127.0.0.1:3040/: Network Error: ... timed out reading response
         sequencer appears unavailable at http://127.0.0.1:3040
         Another project's sequencer may already be running and may not match this project.
#257   : warning: sequencer reachability probe failed (... timed out reading response); continuing
         FAIL hello_world deployment failed
           Error: Error: Transaction submission error
           Hint: inspect sequencer logs and retry.
```

That third hint line is written for exactly this case, and it is the D3 signal ("deploy should fail with a sequencer-unavailable hint instead of a vague wallet error").

Scope it honestly before deciding: master only got this right **on the default port, by accident of the hardcoded needle**. On a custom port master warns-and-continues identically:

```
squatter on 127.0.0.1:3051 -> master and #257 both: "probe failed; continuing" -> vague wallet error
```

So this is a narrow default-port regression, not a broad one, and it is the mirror image of the false positive the PR set out to kill — I don't think `Io => Connectivity` is right either, since a genuinely slow-but-live sequencer produces the same kind. The cheapest fix that loses nothing is to leave the classification alone and make `preflight_sequencer_reachability`'s non-connectivity warning carry the pointer, e.g. append *"a foreign process may be holding the port; check `logos-scaffold localnet status`"*. Your call whether that belongs here or in a follow-up — the blackhole fix is the valuable half and it works.

weboko (approved):
CI is red, else looks good

danisharora099 (approved):
Approving. I read the full diff and the four rounds with @weboko, ran the
suite (566/566) and re-measured the three probe outcomes locally: blackholed
address gives `ConnectionFailed` after 1.0s, port 1 gives refused in under a
millisecond, and an accept-and-never-reply listener gives
`Other(timed out reading response)` after 2.0s. All match what you and
@weboko reported, and the port parser handles the edges I threw at it
(no port, IPv6, userinfo, path suffix, empty string) without a panic path.

Two things before merge, both small:

1. `DOGFOODING.md` line 1694 on master is deleted with no replacement: the
   T4 bullet starting "`wallet topup` failing with a sequencer-unreachable
   hint means the test-node is not on the wallet's expected port" and ending
   with the `--port <localnet.port>` instruction. That is the T4 test-node bullet, unrelated to this PR. It looks like the
   move to D3/D4 took the whole bullet with it instead of just the sentence
   you had appended in `002b519`. Please restore the original text verbatim.

2. The title and description describe the first revision (phrase list plus
   address co-signal, three files, custom-port framing). What ships is an RPC
   probe corroborating timeout-worded failures, structural
   `ErrorKind::ConnectionFailed | Dns` classification in `map_ureq_error`,
   four files, and a fix for the blackholed sequencer on the default port as
   well. Since this becomes the squash commit, rewrite both to match. Linking
   #118 item 2 as @weboko suggested is worth doing in the same pass.

One minor note on
`sequencer_connectivity_failure_corroborates_bare_timeout_when_blackholed`:
`10.255.255.1` being unrouted is true here and on your machine, but on a
runner inside a 10/8 VPC it may come back refused or no-route instead. Both
still map to `ConnectionFailed`, so the test stays green either way; it just
stops exercising the timeout path it is named for. A comment saying so is
enough, no code change needed.

+1 to leaving `ErrorKind::Io` for the follow-up you already agreed with
@weboko. `Io => Connectivity` would misreport a slow live sequencer, and the
default-port squatter hint was only ever right on master by accident.

Rebase on master for CI: #258 fixed the `process.rs:632` fmt deviation, so
`validate` should go green once the fork workflow is approved.

src/commands/wallet_support.rs
```diff
@@ -300,13 +300,141 @@ pub(crate) fn is_connectivity_failure(text: &str) -> bool {
...
+///   trusted without a live RPC round-trip.
+/// - Fallback: any other message that merely says "timeout" is corroborated
+///   against `sequencer_addr` via `probe`, since real wallets don't reliably
+///   name the connection or the address.
+fn sequencer_connectivity_failure_with_probe(
+    text: &str,
+    sequencer_addr: &str,
+    probe: impl Fn(&str) -> bool,
+) -> bool {
+    if is_confirmation_timeout_failure(text) {
```
weboko: **This early-return is justified as a hand-off, but only one of its four call sites has anything to hand off to.**

The doc comment says the confirmation timeout is *"excluded first so the caller's dedicated check still owns it."* Checking every caller:

| call site | submits a tx? | has an `is_confirmation_timeout_failure` branch? |
| --- | --- | --- |
| `wallet.rs:299` — preflight (`check destination wallet initialization`) | no | no — but read-only, so nothing to lose |
| `wallet.rs:340` — `auth-transfer init` | **yes** | **no** |
| `wallet.rs:380` — pinata claim | yes | yes (`wallet.rs:394`) ✅ |
| `deploy.rs:243` — program submission | **yes** | **no** (`grep` finds no match in the file) |

At `wallet.rs:380` the guard is a genuine hand-off. At `wallet.rs:340` and `deploy.rs:243` it is a drop — both submit transactions, so both can genuinely produce the phrase, and neither has a branch to catch it.

The regression is specifically the case where the wallet emits **both** the confirmation line and a transport token, which is what a sequencer dying mid-operation looks like once `combined` is `stdout + "\n" + stderr`. Measured on `6f2ed30` with both classifiers exactly as they exist on this branch, so this is the guard alone and not the `:3040` needle removal:

```rust
let combined = "transaction not found in preconfigured amount of blocks\n\
                error sending request for url (http://127.0.0.1:3040/): Connection refused (os error 111)";

is_connectivity_failure(combined)                                     // -> true   (what deploy used on master)
sequencer_connectivity_failure_with_probe(combined, addr, |_| false)  // -> false  (this branch)
```

master: `Hint: sequencer appears unavailable at http://127.0.0.1:3040`
#257:   `Hint: inspect sequencer logs and retry.`

Reachable, not theoretical. Both paths spawn `wallet.wallet_binary` (`deploy.rs:212`, and the init command in `wallet.rs`), and in the pinned LEZ (`cf3639d8`) both reach the poller:

```
auth-transfer init  -> cli/programs/native_token_transfer.rs:95  wallet_core.poll_native_token_transfer(tx_hash)
                    -> lib.rs:309                                self.poller.poll_tx(hash)
                    -> poller.rs:63                              anyhow::bail!("Transaction not found in preconfigured amount of blocks")
```

Note this is also the exact scenario the new comment at `deploy.rs:239-242` says the call is there for (*"if the sequencer dies mid-deploy"*) — so on that path the call is currently disarmed for its stated purpose.

The fix I'd prefer is to stop asserting the hand-off from inside the shared helper and let each caller own its own ordering — i.e. drop the early-return here and put `is_confirmation_timeout_failure` **before** the connectivity check at `wallet.rs:380`, where the dedicated branch already exists. That keeps `wallet.rs:380` behaviour bit-identical, and restores master's behaviour at the other three. Alternatively keep the guard but give `deploy` and the init path their own confirmation-timeout branches.

Either way I don't think the current shape should merge: the comment documents a contract that three of the four callers cannot honour.

---

src/commands/wallet_support.rs
```diff
@@ -300,13 +300,141 @@ pub(crate) fn is_connectivity_failure(text: &str) -> bool {
...
+        return false;
+    }
+    if is_connectivity_failure(text) {
+        return true;
+    }
+    let lower = text.to_lowercase();
+
+    if let Some(port) = sequencer_port(sequencer_addr) {
+        let mentions_sequencer = [format!("127.0.0.1:{port}"), format!("localhost:{port}")]
```
weboko: Minor, and only bites off-localnet: the fast path matches on **port alone**, under hardcoded `127.0.0.1`/`localhost` aliases, and never consults the configured *host*. So a message about a different machine that happens to share the port takes the zero-probe path:

```rust
sequencer_connectivity_failure_with_probe(
    "Error: connection timed out contacting localhost:3040",
    "http://10.0.0.5:3040",   // our sequencer is remote...
    |_| true,                 // ...and the probe says it is reachable
)  // -> true
```

The probe would have answered correctly here; the fast path returns before it is consulted. `doctor`'s `is_localnet_connectivity_failure` has the same alias assumption, but it is explicitly scoped to *localnet* and takes a `localnet_port`, so the assumption holds there — `sequencer_addr` here is whatever `wallet_config.json` resolved to, which is not necessarily loopback.

Cheap fix if you want it: only take the fast path when the configured host is itself a loopback alias, and otherwise fall through to the probe (which is already the strictly better signal). Happy to see this deferred — it is not reachable from a default localnet setup.

---

src/commands/deploy.rs
```diff
@@ -236,7 +236,11 @@ pub(crate) fn deploy_for_project(
         if !output.status.success() {
             let summary = summarize_command_failure(&output.stdout, &output.stderr);
             let combined = format!("{}\n{}", output.stdout, output.stderr);
-            let connectivity_failure = is_connectivity_failure(&combined);
+            // `preflight_sequencer_reachability` already bails on a
+            // Connectivity error before this loop starts, so by the time we
+            // get here the sequencer answered RPC moments ago. This call
+            // mainly earns its keep if the sequencer dies mid-deploy.
+            let connectivity_failure = sequencer_connectivity_failure(&combined, &sequencer_addr);
```
weboko: Minor, and specific to this call site rather than the wallet ones: this is inside the per-program loop and the failure arm ends in `continue`, not a bail. So when the sequencer dies partway through a multi-program deploy, every remaining program fails, and each one issues its own probe asking the identical question and getting the identical answer.

Bounded — `rpc_get_last_block_id` pins `timeout_connect(1s)`/`timeout_read(2s)` — so it is a few seconds per program, not a hang. (The three `wallet.rs` call sites don't have this shape: each of them bails, so a `topup` run pays for at most one probe.)

Not worth blocking on, but hoisting the result to at most once per `deploy_for_project` invocation would be cheap and would also stop the per-program results from being able to disagree with each other within a single run.

---

DOGFOODING.md
```diff
@@ -356,6 +356,7 @@ Both deploy paths honor `--json`, with a different shape each. `--program-path -
 
 - If either `--json` path starts emitting a guaranteed-null `tx` key, or mixes command echoes and the human-readable summary into the JSON stream, record that as a machine-readability regression.
 - If localnet is unreachable, deploy should fail with a sequencer-unavailable hint instead of a vague wallet error.
+- That hint is corroborated by an RPC call (`getLastBlockId`) against the resolved `sequencer_addr` rather than by wallet output alone. The probe separates "refused/unreachable" from "something answered" — it does not verify the responder is our sequencer, so a foreign process squatting on the configured port will still suppress the hint.
```
weboko, line 359: This note is accurate now — I verified the squatter claim against the real classifier rather than taking it on trust, and all three shapes do suppress the hint:

```
accept-never-replies  -> rpc=Other  hint_shown=false
foreign-http-404      -> rpc=Other  hint_shown=false
non-http-garbage      -> rpc=Other  hint_shown=false
```

The gap is the other direction: **the runbook still has no way to reproduce the failure mode this PR exists to fix.** D3's only unreachable-sequencer signal is *"If localnet is unreachable"*, which in practice means `localnet stop` — a **refused** connection, which master already classified correctly. The blackhole case (SYN dropped: firewalled, wrong host, host down) is a distinct mode, it is the one the `ErrorKind` change actually repairs, and `grep` finds nothing in the file that exercises it:

```
$ grep -c "10.255.255\|blackhole\|SYN" DOGFOODING.md
0
```

So a dogfooder running D3 exactly as written passes on both master and this branch, and the regression this PR fixes stays invisible to the runbook — which is the "if the product surface changes and this runbook does not, the runbook is wrong" case from the maintenance note at the top.

Worth one more bullet here, since the recipe is two lines and deterministic (bounded by the existing `timeout_connect(1s)`, so it is fast, not a 30s hang):

```suggestion
- That hint is corroborated by an RPC call (`getLastBlockId`) against the resolved `sequencer_addr` rather than by wallet output alone. The probe separates "refused/unreachable" from "something answered" — it does not verify the responder is our sequencer, so a foreign process squatting on the configured port will still suppress the hint.
- Two unreachable shapes are worth distinguishing, and only the first is covered by stopping localnet. **Refused** (`localnet stop`, nothing listening) fails fast. **Blackholed** (SYN dropped — firewalled host, wrong host, host down) is reproduced by pointing the project's sequencer at an unrouted address, e.g. `10.255.255.1:3040`, and re-running `deploy`. Both must fail with the sequencer-unavailable hint. If the blackholed one instead prints `warning: sequencer reachability probe failed (...); continuing` and then a vague wallet error, that is the regression this signal protects — the preflight classified a connect *timeout* as "something else" and burned a wallet submission against a dead address.
```

---

weboko (commented):
## Re-review of `6f2ed30` — the blackhole fix is real and I reproduced it; one new blocking finding in `deploy`

New commit since my 2026-08-26 pass (`6f2ed30`, force-pushed over the `7cd8cba` I measured last time), so this is a re-review of the diff, not a status note. Environment built from scratch and verified end-to-end: Nix 2.35.2 (flakes on), Rust 1.94.1, risc0 3.0.5 (`r0vm` + `cargo-risczero`, matching the pinned LEZ `cf3639d8`'s `risc0-zkvm = 3.0.5`), guest toolchain `r0.1.91.1`.

### The core fix works — independently reproduced

The `ErrorKind`-based classification in `map_ureq_error` does what the PR says. Same call, same address, both branches:

```
master : Other(http://10.255.255.1:3040/: Connection Failed: Connect error: connection timed out)
#257   : Connectivity(http://10.255.255.1:3040/: Connection Failed: Connect error: connection timed out)
```

That is the whole ballgame for `preflight_sequencer_reachability`: `Other` warns-and-continues into a wasted wallet submission and a vague error, `Connectivity` bails with the hint. Bounded at ~1s by the existing `timeout_connect(1)`. This half is good and I'd take it as-is.

### My previous points, re-measured

- **DOGFOODING overclaim (my 2026-08-26 note on line 359) — fixed, and I verified the new wording rather than taking it on trust.** All three squatter shapes against the real classifier:

  ```
  accept-never-replies  -> rpc=Other  hint_shown=false
  foreign-http-404      -> rpc=Other  hint_shown=false
  non-http-garbage      -> rpc=Other  hint_shown=false
  ```

  The new text ("it does not verify the responder is our sequencer, so a foreign process squatting on the configured port will still suppress the hint") is now accurate.
- **Guard-ordering test (2026-08-21, line 784) — fixed.** The suite now pins ordering with a `"http error"`-carrying variant, so moving the guard below the transport check fails, not just deleting it.
- **Test named for timeout but exercising refused (2026-08-26, line 857) — fixed.** The `10.255.255.1` blackhole test is added alongside, and it does exercise the timeout path here (measured: `TimeoutError` after 3.00s on a raw connect in this container, so the test is not silently degenerating to the refused case).
- **DOGFOODING edit landed only in T4 (2026-08-26, line 1694) — fixed.** D3 (359) and D4 (421) now carry the reachability note and T4's original bullet is restored.
- **Bespoke socket probe / `port_open` duplication (2026-08-25) — resolved** by routing through `rpc_get_last_block_id`; the userinfo edge went away with it.
- **Squatter loses the hint on the default port (my 2026-08-26 "remaining" note) — still open, and I still think that is fine as a follow-up.** `ErrorKind::Io` continues to fall through to the text fallback. It is now honestly documented rather than papered over, which was my actual complaint.

### Checks

`cargo fmt --check`, `cargo check`, `cargo test` all clean **on the merge result** (566 lib + 182 + 3 + 5, 0 failures). Note the branch is 3 commits behind `master` and its merge-base is `20b9d99`, so `cargo fmt --check` fails *on the branch in isolation* — purely because `00448b7` (the `process.rs` format fix) is not in it. Merges clean and formats clean once merged, and CI checks out the merge commit, so this is not a defect; flagging only so it is not mistaken for one. CI has not actually run on this head — the run is `action_required` (fork PR awaiting approval), so nothing here has been machine-validated yet.

### Dogfooding — D3/D4 connectivity signals, real sequencer toolchain

Full provisioning per the runbook (`test-node doctor --json | jq .ok` → `true`, all checks pass; real `sequencer_service` 25 MB and LEZ-local `wallet` 21 MB built from `cf3639d8`; `setup complete`, default wallet seeded).

**D4, refused (`localnet` stopped)** — hint fires, as on master. Classified from text (`tcp connect error` / `Connection refused` are both in the needle list), no probe needed:

```
error: wallet topup failed during account preflight: Error: client error (Connect)
sequencer appears unavailable at http://127.0.0.1:3040
```

**D4, blackholed (`sequencer_addr` → `10.255.255.1:3040`, SYN dropped)** — this is the case the PR exists for, and it now works:

```
error: wallet topup failed during account preflight: Error: Request timeout
sequencer appears unavailable at http://10.255.255.1:3040
```

Worth flagging what the wallet actually printed: **`Error: Request timeout`, verbatim** — no address, no "connection", nothing a text matcher can key on. That is exactly the string the new unit tests assert against, so the fixture is real rather than invented, and it confirms the `mentions_timeout` → probe fallback is genuinely load-bearing here, not decorative. On master this text matches no needle and there is no probe, so the hint does not fire.

Total wall time 61s, essentially all of it the wallet's own `seq_poll_timeout: 30s` retries; the added probe contributes ~1s of that.

**D3, blackholed** — the clearest demonstration of what this PR buys. Same project, same dead address, the two binaries side by side:

```
master : $ wallet deploy-program …/hello_world.bin      <- submits anyway
         FAIL hello_world deployment failed
           Error: Error: Transaction submission error
           Hint: inspect sequencer logs and retry.

#257   : error: cannot deploy programs: http://10.255.255.1:3040/: Connection Failed: Connect error: connection timed out
         sequencer appears unavailable at http://10.255.255.1:3040
         Run `logos-scaffold localnet start`.
```

master burns a wallet submission against a dead address and then misattributes the result; `#257` bails at preflight in **1.009s** and names the actual problem. `wallet topup` shows the same shape — master's advice is actively misleading (*"verify the destination with `wallet -- account get …`"* when the destination is fine and the sequencer is gone), and `--json` moves from `"reason": "preflight"` to `"reason": "connectivity"`, so machine consumers get the right category too.

**Controls, against a live sequencer (no false positives)** — this is the part I most wanted to see, given the change widens what counts as connectivity:

```
wallet topup   -> auth-transfer init + pinata claim both executed by the real sequencer -> "wallet topup complete", exit 0
deploy hello_world                    -> OK submitted, program_id 5f912d3d…, exit 0
deploy --program-path … --json        -> {"program":"…","program_id":"d97bed76…","status":"submitted"}, exit 0 (single object, no null `tx` key)
```

Note the happy-path `topup` goes through `auth-transfer init` — the call site I flag above — and is unaffected, so that finding is about the failure path only.

**Squatter, at the CLI rather than in a unit test**, to check the new D3/D4 wording against reality. Foreign listener on `127.0.0.1:3040` that accepts and never replies:

```
FAIL hello_world deployment failed
  Error: Error: Transaction submission error     <- no sequencer-unavailable hint
```

Hint suppressed, exactly as the new text now says it will be. So the docs are accurate, and my 2026-08-26 "remaining" sub-case is confirmed still open — which I'm content to leave to a follow-up, as I said then.


### What I'd want changed before this merges

One thing, and it is not the part of the PR I was worried about last round:

**The `is_confirmation_timeout_failure` early-return is documented as a hand-off to "the caller's dedicated check", but only one of the four call sites has one.** `wallet.rs:380` does (`wallet.rs:394`). `wallet.rs:340` (`auth-transfer init`) and `deploy.rs:243` both submit transactions, both can produce that phrase, and neither has a branch to catch it — so for them the guard is a drop, not a hand-off. Details and a measured before/after are inline on `wallet_support.rs:346`. Notably this disarms `deploy.rs:243` for the exact case the new comment above it says the call is there for.

Three smaller notes inline (the fast path ignores the configured host; a multi-program `deploy` re-probes once per remaining program; and D3/D4 still have no way to reproduce a blackholed sequencer, which is the very mode this PR fixes). None of those block.

### Security / privacy

Nothing I'd hold the PR for. Two things worth having said out loud:

- `wallet topup` did not previously make its own network calls — it shelled out to the wallet binary and read the text. It now issues a `getLastBlockId` POST to `sequencer_addr` on error paths. `sequencer_addr` comes from `wallet_config.json`, i.e. it is project-controlled, so this is a (small) new outbound surface reachable by cloning a hostile project. It is bounded (1s connect / 2s read), carries no credentials, and the body is a fixed JSON-RPC payload whose response is reduced to a bool — and `deploy` already did exactly this via `preflight_sequencer_reachability`, so the trust boundary is not new, only the command that touches it.
- No new secret exposure: the hint and `summarize_command_failure` print the same address and first stderr line they did before.

### Verdict

The `ErrorKind` half is a real fix, well tested, and the docs are now honest about what the probe does and does not prove — that is a clear improvement over both master and the last revision. I'm not approving on this pass only because of the guard finding above, which is a straightforward change. Fix that and I'd approve.

---

PR #264, opened 2026-09-01
fix(localnet): make [localnet].port actually take effect

### LINKED ISSUE #263: [localnet].port is a no-op, and fixing only the sequencer half breaks deploy
## What happens

`[localnet].port` in `scaffold.toml` does nothing. Set it to anything other than `3040` and the sequencer still binds `3040`.

The sequencer takes its RPC port from a CLI flag, not from the config file:

```rust
// lez sequencer/service/src/main.rs
#[clap(short, long, default_value = "3040")]
port: u16,
```

`run_server` binds that value. It never reads the `port` key that scaffold's `prepare_sequencer_config` patches into `sequencer_config.json`, and `localnet start` does not pass `--port`. `test-node` does (`src/testnode/mod.rs`), which is why it works there.

At the default port the flag and the clap default agree, so this is invisible. It shows up on a machine where something else already holds 3040: every `localnet start` dies with a bare `Error: Failed to build RPC server / Address already in use` no matter what port you configure.

## Why the one-line fix is not enough

Passing `--port` makes the sequencer honour the setting — and then `deploy` breaks, because the wallet is still pointed at the old port:

```
error: cannot deploy programs: http://127.0.0.1:3040/: Connection Failed: Connection refused
```

`load_wallet_runtime` takes `sequencer_addr` from the wallet config verbatim, and the seeded config carries the vendored LEZ debug value `http://127.0.0.1:3040`. `default_sequencer_http_url_for_project` (which does derive from `[localnet].port`) is only the fallback for when the key is absent:

```rust
let sequencer_addr = wallet.sequencer_addr.clone()
    .unwrap_or_else(|| default_sequencer_http_url_for_project(project));
```

`doctor` already knows: it emits `WARN | wallet network config | wallet may point to non-local sequencer` with a remediation telling you to hand-edit `sequencer_addr`. With the sequencer moved and the wallet not, `doctor` also goes to `FAIL | wallet usability` — the wallet CLI panics on a refused connection to 3040.

So today both halves are wrong in the same direction and cancel out. Fix one and a previously-working project breaks.

## Suggested fix

Do both together:

1. `localnet start` passes `--port <configured>`, mirroring `test-node`.
2. Wallet seeding writes `sequencer_addr` from `default_sequencer_http_url_for_project`, so the wallet follows `[localnet].port`. (Or `load_wallet_runtime` prefers the project port over a loopback address whose port disagrees — but rewriting the config once at setup is less surprising than silently overriding a value the user can see.)

Then `doctor`'s existing `wallet network config` check becomes meaningful instead of a standing warning, and `D2` can gain a real assertion that a non-default port works end to end.

## Note

`tests/cli.rs::localnet_start_patches_config_and_uses_configured_port` currently asserts `--port` must **not** appear, which locks the current behaviour in. That assertion needs inverting as part of the fix.

Found while rerunning `D1`/`D3` for #259 on a box where another project's localnet held 3040. I had the sequencer half in #260 and reverted it there — it is out of scope for that PR and, on its own, a regression.

erhnysr:
Fixes #263.

`[localnet].port` in `scaffold.toml` did nothing — two bugs canceled out in opposite directions.

**Sequencer side**: the pinned `sequencer_service` binds the port given by its `--port` CLI flag (clap default `3040`) — confirmed by reading `sequencer/service/src/main.rs` and `lib.rs::run()` directly at the pinned SHA (`logos-execution-zone` @ `cf3639d8`). It never reads the `sequencer_config.json` `port` key that `prepare_sequencer_config` patches for this purpose. `start_localnet` never passed `--port`, so the sequencer always bound `3040` regardless of the configured port. (`test_node` already does this correctly — a separate code path.)

**Wallet side**: `prepare_wallet_home` seeds a fresh wallet home by copying the vendored LEZ debug `wallet_config.json` verbatim. That file hardcodes `sequencer_addr: http://127.0.0.1:3040`. Since the key is present, `load_wallet_runtime`'s fallback to `default_sequencer_http_url_for_project` (which does follow `[localnet].port`) never triggers.

As the issue notes, fixing only one side breaks the other — moving the sequencer without fixing the wallet breaks `deploy` with `Connection refused`. This PR does both:
- `start_localnet` now passes `--port <configured>`, mirroring `test_node`'s existing pattern.
- `prepare_wallet_home` now takes the project's configured sequencer address and writes it into the freshly-seeded wallet config, instead of leaving the vendored `3040` default.
- Fixed a stale doc comment on `prepare_sequencer_config` that incorrectly claimed the pinned sequencer doesn't accept `--port`.

`tests/cli.rs::localnet_start_patches_config_and_uses_configured_port` previously locked in the old behavior (asserted `--port` must **not** be passed, per the issue's own note); inverted here, with the fake sequencer now binding `--port` like the real one. Added two regression tests for the wallet-seeding half in `state.rs`.

- Full suite: 587/587 lib tests pass, 186/186 cli integration tests pass.
- No behavior change when `[localnet].port` is left at the default `3040`.

### DIFF
--- a/src/commands/localnet.rs
+++ b/src/commands/localnet.rs
@@ -351,6 +351,8 @@ fn start_localnet(
     sequencer_cmd
         .current_dir(lez)
         .arg(&patched_config_path)
+        .arg("--port")
+        .arg(localnet_port.to_string())
         .env("RUST_LOG", "info")
         .env("RISC0_DEV_MODE", if risc0_dev_mode { "1" } else { "0" });
 
@@ -693,8 +695,14 @@ fn build_status_report(
 /// Produce a sequencer config patched for scaffold's localnet — port set to
 /// the project's configured port and `max_block_size` widened so the bundled
 /// deploy flow fits in a single block — and return the absolute path to the
-/// patched file. The pinned LEZ version does not accept `--port` as a CLI flag
-/// — it reads everything from the config file passed as its first argument.
+/// patched file.
+///
+/// The pinned sequencer binary ignores this file's `port` key for RPC
+/// binding: it always binds the `--port` CLI flag (clap default `3040`),
+/// which `start_localnet` also passes explicitly so it matches this
+/// patched config's port. The `port` key is still written here for any
+/// tooling that inspects the config file directly, and so the file stays
+/// internally consistent with the port the sequencer is actually bound to.
 ///
 /// The patched copy is written under `dest_dir` (the project's
 /// `.scaffold/state/`), **not** back into the vendored LEZ checkout. Writing
--- a/src/commands/run.rs
+++ b/src/commands/run.rs
@@ -20,7 +20,9 @@ use crate::commands::run_state::{
 };
 use crate::commands::setup::ensure_default_wallet_seeded;
 use crate::commands::wallet::{cmd_wallet_topup_inner, TopupOutcome};
-use crate::commands::wallet_support::set_wallet_home_env;
+use crate::commands::wallet_support::{
+    default_sequencer_http_url_for_project, set_wallet_home_env,
+};
 use crate::constants::{
     DEFAULT_RUN_LOCALNET_TIMEOUT_SEC, FRAMEWORK_KIND_LEZ_FRAMEWORK, SPEL_BIN_REL_PATH,
     WALLET_BIN_REL_PATH,
@@ -548,7 +550,11 @@ fn segment_match(pat: &[u8], seg: &[u8]) -> bool {
 fn reseed_after_wipe(project: &Project) -> DynResult<()> {
     let lez = resolve_repo_path(project, &project.config.lez, "lez")?;
     let wallet_home = project.root.join(&project.config.wallet_home_dir);
-    prepare_wallet_home(&lez, &wallet_home)?;
+    prepare_wallet_home(
+        &lez,
+        &wallet_home,
+        &default_sequencer_http_url_for_project(project),
+    )?;
     ensure_default_wallet_seeded(&project.root, &wallet_home, &lez.join(WALLET_BIN_REL_PATH))
 }
 
@@ -1708,7 +1714,10 @@ mod tests {
         {
             let lez = baseline.path().join("lez");
             let wallet_home = baseline.path().join(".scaffold/wallet");
-            prepare_wallet_home(&lez, &wallet_home).expect("baseline prepare");
+            // Matches make_test_project's default localnet.port (3040) below,
+            // so this stays byte-equivalent to the post-reset path.
+            prepare_wallet_home(&lez, &wallet_home, "http://127.0.0.1:3040")
+                .expect("baseline prepare");
             ensure_default_wallet_seeded(
                 baseline.path(),
                 &wallet_home,
--- a/src/commands/setup.rs
+++ b/src/commands/setup.rs
@@ -11,9 +11,9 @@ use crate::state::prepare_wallet_home;
 use crate::DynResult;
 
 use super::wallet_support::{
-    first_public_address_in_listing, first_public_wallet_address, read_default_wallet_address,
-    set_wallet_home_env, summarize_command_failure, wallet_password, wallet_state_path,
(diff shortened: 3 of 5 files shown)


weboko (commented):
Verified the fix end-to-end on a non-default port, and confirmed the premise against the pinned LEZ source.

**Premise checks** (LEZ `cf3639d8`): `sequencer/service/src/main.rs` parses `#[clap(short, long, default_value = "3040")] port` and hands it to `sequencer_service::run(config, port)`, which binds `SocketAddr::from(([0,0,0,0], port))` in `run_server`. Nothing reads the config file's `port` key for RPC binding. Both halves of the PR description hold.

**What I ran** (fresh default-template project, `[localnet].port = 3141`, real sequencer + wallet):

- `setup` → seeded `.scaffold/wallet/wallet_config.json` with `"sequencer_addr": "http://127.0.0.1:3141"`, all four `initial_accounts` and the other keys intact.
- `localnet start` → `sequencer_service …/sequencer_config.json --port 3141`; 3141 open, 3040 closed.
- `doctor` → `PASS | sequencer port 3141`, `PASS | wallet network config | wallet points to local sequencer`, `PASS | wallet usability`. That row is a standing WARN on `master` at any non-default port, so this is the check becoming meaningful, exactly as #263 predicted.
- `wallet topup` → `Network: local sequencer (http://127.0.0.1:3141)`; `deploy hello_world` → `Succeeded: 1`.
- Full suite green: 587 lib / 186 CLI / 3 test-node / 5 doctests, including the inverted `localnet_start_patches_config_and_uses_configured_port`.

---

**The gap: projects that already ran `setup` are left broken, and the error sends them the wrong way**

`prepare_wallet_home` only writes `sequencer_addr` when it is seeding a *fresh* wallet home — and `prepare_wallet_home_does_not_touch_existing_wallet_config` deliberately locks that in. But the population this bug affects is precisely the projects that already ran `setup` with `[localnet].port != 3040`. Today those work by accident (both halves wrong in the same direction). After this change the sequencer moves and the wallet does not, which is the regression #263 warned about — just relocated from "everyone" to "everyone who upgrades".

Reproduced on the same project by restoring the wallet config to the pre-PR on-disk state (`sequencer_addr = http://127.0.0.1:3040`) while localnet ran on 3141:

```
$ lgs deploy hello_world_with_move_function
error: cannot deploy programs: http://127.0.0.1:3040/: Connection Failed: Connection refused (os error 111)
sequencer appears unavailable at http://127.0.0.1:3040
Run `logos-scaffold localnet start`.
Another project's sequencer may already be running and may not match this project.
```

Localnet *is* running, and the advice is to start it again. `doctor` does say the real thing (`WARN | wallet network config` with a remediation naming the file and the value), but nothing in the failing command points there, so the natural next step is a loop of `localnet start` / `localnet status` that never mentions the port mismatch.

Either fix would close it:

1. **Detect and say so** — in `load_wallet_runtime` (or the deploy preflight), when `sequencer_addr` is a loopback URL whose port disagrees with `[localnet].port`, replace the hint with that fact: *"wallet is configured for 127.0.0.1:3040 but `[localnet].port` is 3141 — update `.scaffold/wallet/wallet_config.json` or run `lgs doctor`"*. Lowest risk, and keeps the PR's "don't silently rewrite a value the user can see" stance.
2. **Repair on `setup`** — rewrite an existing `sequencer_addr` when it is loopback with a disagreeing port (leave any non-loopback address alone). Fixes it without the user reading anything, at the cost of the invariant the new test asserts.

I'd take (1); (2) alone would still leave anyone who never re-runs `setup` stuck. Whichever you pick, it seems worth a line in #263 about the upgrade path, since the issue explicitly framed "fix one and a previously-working project breaks" as the thing to avoid.

**Minor:** seeding now round-trips the config through `serde_json`, so the written file is re-indented (4→2 spaces) and its keys come out alphabetically reordered (`initial_accounts` first, `sequencer_addr` last). Harmless, but this is a file `doctor` tells people to hand-edit and that some will have diffed before; `serde_json`'s `preserve_order` feature would keep it recognisable if that matters to you.

---

PR #266, opened 2026-09-04
feat(basecamp): opt into the inspector build via `setup --inspector`

### LINKED ISSUE #265: basecamp: allow selecting bin-bundle-dir-inspector via [repos.basecamp].attr
## Summary

`bin-bundle-dir-inspector` — the Basecamp output with the QML inspector compiled in — cannot be selected through scaffold today. It is the one build a UI end-to-end test suite needs, and it is the only thing keeping raw `nix build` in an otherwise fully `lgs`-driven project.

The gap is smaller than it first appears: the *pin* and the *attr* are already expressible in `scaffold.toml`. What is missing is the dev-vs-portable **classification** of that attr.

## Why this matters

`lgs basecamp develop --help` states its own purpose as:

> Verb-set symmetry so contributors stop reaching for raw `nix`.

That is exactly the goal here. In [`fryorcraken/logos-radicle-module`](https://github.com/fryorcraken/logos-radicle-module) we converted both CI workflows and all local dev docs to `lgs` — `build`, `build-portable`, `install`, `launch`, `develop`, `paths`, `doctor`. The single remaining raw invocation is:

```bash
nix build 'github:logos-co/logos-basecamp/aa237766baf61404e12da86b7303cb41065464c9#bin-bundle-dir-inspector' -o result-bundle
```

Notably, that rev is byte-identical to scaffold's own `DEFAULT_BASECAMP_PIN`. We are asking for the *same commit* scaffold already pins, just a different flake attr.

## What already works

From `src/constants.rs`:

```rust
/// Basecamp commit pin — `logos-basecamp` tag `0.2.3` …
/// Projects can override via `[repos.basecamp].pin` in `scaffold.toml`.
pub(crate) const DEFAULT_BASECAMP_PIN: &str = "aa237766baf61404e12da86b7303cb41065464c9";
pub(crate) const BASECAMP_ATTR: &str = "app";
```

and `basecamp docs` documents a per-system attr map:

```toml
[repos.basecamp.attr]
aarch64-darwin = "bin-macos-app"
aarch64-linux  = "bin-appimage"
```

So this ought to be enough to build it:

```toml
[repos.basecamp]
attr = "bin-bundle-dir-inspector"
```

## What blocks it

```rust
/// `[repos.basecamp].attr` values that select the portable distribution stack.
/// Anything else (including unrecognised attrs) is treated as dev.
pub(crate) const BASECAMP_PORTABLE_ATTRS: &[&str] =
    &["bin-macos-app", "bin-appimage", "bin-bundle-dir"];
```

`bin-bundle-dir-inspector` is not in that list, so it falls through to the dev classification. That picks:

- **the wrong XDG subpath** — `Logos/LogosBasecampDev` instead of `Logos/LogosBasecamp` (`BASECAMP_XDG_APP_SUBPATH_DEV` vs `..._PORTABLE`), and
- **the wrong lgpm variant** — `cli` (accepts `<host>-dev` `.lgx` variants) instead of `cli-portable` (accepts bare `<host>`).

So `setup` would build the correct binary and then seed profiles that binary cannot load. The failure mode is the quiet one: Basecamp logs a single line —

```
Warning: module 'radicle' … was installed for variant 'linux-amd64-dev' which is
not supported on this platform and will not be loadable:
supported variants [linux-x86_64, linux-amd64].
```

— declines to load the module, and the UI opens to nothing. A test suite then times out on its first step with nothing explaining why.

## Proposed change

Add the attr to the portable list:

```rust
pub(crate) const BASECAMP_PORTABLE_ATTRS: &[&str] =
    &["bin-macos-app", "bin-appimage", "bin-bundle-dir", "bin-bundle-dir-inspector"];
```

`bin-bundle-dir-inspector` is `bin-bundle-dir` plus a compile-time flag; it is the same distribution stack by construction, so classifying it as portable looks correct rather than a special case. Happy to send a PR if you agree with the shape.

### Alternative, if the list is meant to stay closed

A `[repos.basecamp].stack = "portable" | "dev"` escape hatch would let projects classify an attr scaffold does not know about, which also covers future variants without another constant bump.

## Out of scope for this issue

Two adjacent things, noted only so they are not conflated:

1. **`lgs basecamp build` has no `--print-build-logs` passthrough.** Converting our UI-test workflow to `build-portable` meant losing it, so a failed module build in that job now reports less than the raw `nix build` did. Separate ask; can file if useful.
2. **`lgs basecamp setup` rewrites `scaffold.toml` and drops its comments.** Our config carries load-bearing ones (why `runtime_dir` must be the session's real runtime dir — the 108-byte `sun_path` cap that made every module segfault, and Qt locating the Wayland socket through `XDG_RUNTIME_DIR`). Losing them silently re-opens a bug that cost a full debugging session. Also happy to file separately.

## Environment

- `logos-scaffold` 0.3.1 (released, from crates.io)
- Basecamp pin `aa237766baf61404e12da86b7303cb41065464c9` (tag 0.2.3) — scaffold's own default
- Linux x86_64

fryorcraken:
Closes #265.

## What this does

`bin-bundle-dir-inspector` — the Basecamp output with the QML inspector compiled in — could not be selected through scaffold, leaving a sitometres UI end-to-end suite as the last thing reaching for raw `nix build` in an otherwise fully `lgs`-driven project.

The issue proposed appending the attr to `BASECAMP_PORTABLE_ATTRS`. That fixes the build selection but leaves the opt-in as a flake attr name the user has to know exists and spell correctly — for an output upstream deliberately marks as *not* a release artifact. This surfaces it as an intent instead:

```console
$ lgs basecamp setup --inspector
inspector build selected (.#bin-bundle-dir-inspector) — test-only, not a release artifact
cloning basecamp at aa237766...
```

- `--inspector` selects the build and persists the choice, so later runs stay on it without repeating the flag.
- `--no-inspector` restores the default. It only undoes *our* attr — a hand-set `bin-appimage` survives, and scaffold says so rather than appearing to act.
- Neither flag leaves the configured attr alone, so a plain `setup` re-run never silently drops the opt-in.
- Both edit **only the current host's entry**. `setup` builds for the machine it runs on, so a Linux developer's toggle must not discard the project's `aarch64-darwin` mapping.

## What this unblocks

The consumer is a **sitometres** UI end-to-end suite, which drives Basecamp headlessly through the QML inspector (via `logos-qt-mcp`). The inspector is a compile-time feature, deliberately off in every shipping output, so no release artefact can be driven this way.

Downstream context: [fryorcraken/logos-radicle-module#4](https://github.com/fryorcraken/logos-radicle-module/pull/4) converted both CI workflows to `lgs`, with its UI-test job pointing sitometres' `--app-dir` at `.scaffold/basecamp/portable/` and running 18/18 steps green. The single raw `nix build` it could not convert is the inspector bundle.

## The classification was the actual bug

Confirmed against the pinned flake (`aa23776`, tag 0.2.3): `bin-bundle-dir-inspector` is `dirBundler appDistributedWithInspector`, built from the same `portable = true` derivation as `bin-bundle-dir` and differing only by `enableInspector`. It is the portable stack by construction — the issue's reasoning holds.

Classified as dev it picked `LogosBasecampDev` for the XDG subpath and the `cli` lgpm variant, so Basecamp declined to load every module with one `variant ... not supported on this platform` line and opened to an empty UI.

Three adjacent instances of that same quiet failure, all reachable from the new flag and fixed here:

1. **Stale `[repos.lgpm].attr`.** Only defaulted when empty, so toggling the stack on an already-set-up project left `cli` against a portable build. The two are one choice and now move together; an attr scaffold doesn't manage is preserved but warned about.
2. **Choice lost on a failed build.** Persisted only after the long basecamp build, so an interrupt reverted it. An explicit flag now writes both attrs *before* the build.
3. **`--no-inspector` was a silent no-op** on a project with a per-platform `attr` map — it compared the scalar, not what `effective_attr` resolves. Now compares the effective attr, and reports when it declines.

## Also included

From the issue's follow-up comment: **`--print-output` on `basecamp build` / `build-portable`.** Reuses the flag `install` already exposes rather than adding `--print-build-logs` as a second name for the same thing. As the follow-up notes, there was no log-carrying module build left in CI: a failing build reported the derivation failure without the compiler output explaining why.

Docs also record `.scaffold/state/basecamp.state` and its three keys. `setup` has always written it and nothing documented it — which left `--inspector` telling a harness author to build something while leaving them to derive its path, and that path is exactly what they must not derive (the correct entry point under the nix output varies by basecamp generation and stack).

That contract now lives in `docs/basecamp-module-requirements.md` — the doc `basecamp docs` prints — rather than only in `docs/commands.md`. Dogfooding this PR downstream showed why: a consumer looking for "where does `setup` leave the binary" goes to `basecamp docs`, finds the module contract, and reconstructs a cache path by hand when it isn't there. The section covers the three things a consumer cannot infer — read it rather than reconstruct it, it is a read-only nix out-link whose real ELF is a dot-prefixed sibling on portable builds, and it moves whenever the stack moves (so `--inspector` invalidates a cached path).

## Breaking change for library embedders

`BasecampCommand::Setup` goes from a unit variant to a struct variant so it can carry the tri-state. `BasecampCommand` is a public enum on a published crate, so any embedder constructing it stops compiling. The CLI is unaffected, and nothing in `tests/` exercises the API surface, so CI does not catch this.

```rust
// before
project.basecamp(BasecampCommand::Setup)?;

// after — `None` is the previous behaviour: leave the configured stack alone
project.basecamp(BasecampCommand::Setup { inspector: None })?;
```

Recorded in `ADR.md` and in the rustdoc on the variant itself. **No version bump in this PR** — this repo bumps in standalone `chore:` commits, so one is expected to follow separately. `#[non_exhaustive]` is deliberately not added: it would itself be a breaking change and is a separate decision.

## Split out of this PR

The issue's other aside — `setup` preserving `scaffold.toml` comments — is a change to the config writer that **every** scaffold command goes through, so it is on its own branch (`fix/preserve-scaffold-toml-comments`) with its own PR. It is unrelated to selecting a basecamp build, and reviewing them together would let a problem in either block the other.

## Testing

- **796 tests green**: 596 lib + 192 CLI + 3 API + 5 doc, 0 failures.
- `cargo fmt --check` clean; `cargo clippy --all-targets` warning count unchanged from the base commit (**62** pre-existing, 0 new — an earlier revision of this description said 64, which was wrong). None of the 62 is in a file this PR touches.
- Unit tests cover the attr selection round trip, per-host map editing (including that a sibling platform's entry survives both directions), the lgpm realignment both ways, and the inspector attr's stack classification — asserted via the XDG subpath and variant key, not just a boolean.
- Verified end to end locally: `--inspector` persists `attr = "bin-bundle-dir-inspector"` and `lgpm.attr = "cli-portable"` even when the build is interrupted.

The earlier `LOGOS_SCAFFOLD_E2E_NIX=1` gate is gone (review feedback). `ensure_nix_present` is only a `which("nix")`, so a stub `nix` on PATH satisfies it and an unresolvable `[repos.basecamp]` source fails the clone in milliseconds — which is well inside the pre-build window the test is actually about. `basecamp_setup_inspector_persists_both_attrs_before_the_build` now runs unconditionally in CI in 0.01s with no network and no nix store, alongside a sibling covering the per-platform-map case.

Docs updated per CONTRIBUTING: `docs/commands.md` for the flags, `docs/basecamp-module-requirements.md` (what `basecamp docs` prints) for the `.scaffold/state/basecamp.state` contract, and `DOGFOODING.md` for `B1` / `B5` plus the rerun-guidance rows.


weboko (commented):
Reviewed at `a806193`. Verified on Linux (`x86_64-linux`) with Nix 2.35.2 and a real `nix` on PATH.

**What I checked out as correct.** The classification argument holds against the pinned flake. In `logos-basecamp@aa23776`, `flake.nix:230` is `binBundleDirInspector = withMainProgram (dirBundler appDistributedWithInspector)`, and `appDistributedWithInspector` (`flake.nix:180-186`) is the same `./nix/app.nix` import as `appDistributed` with `portable = true; enableInspector = true`. `nix/app.nix:201` shows `portable` is what drives `-DLOGOS_PORTABLE_BUILD`, and it is also what selects the unwrapped `$out/bin/LogosBasecamp` layout — so adding `bin-bundle-dir-inspector` to `BASECAMP_PORTABLE_ATTRS` (and with it `Logos/LogosBasecamp` rather than `Logos/LogosBasecampDev`, and `cli-portable`) is right, not a guess. `nix eval .#bin-bundle-dir-inspector.drvPath` resolves cleanly at that pin.

The flag itself behaves as documented end to end: `setup --inspector` writes `attr = "bin-bundle-dir-inspector"` and `[repos.lgpm].attr = "cli-portable"` before the build starts, a later `--no-inspector` moves both back to `app`/`cli`, `--no-inspector` on a project that never opted in declines with a note instead of silently doing nothing, and `--inspector --no-inspector` resolves last-flag-wins through clap's `overrides_with`. 797 tests pass (600 lib / 189 CLI / 3 API / 5 doc), `cargo fmt --check` is clean, and `cargo clippy --all-targets` reports zero warnings in any file this PR touches.

The findings below are all in the `save_project_config` half of the change.

---

### 1. `save_project_config` now panics on a `scaffold.toml` that `parse_config` accepts (blocking)

Switching the writer from "render into an empty document" to "merge over the parsed original" means `write_config_into` now inherits whatever *shape* the user's file has, and the `.expect("… table")` calls in it stop being structurally unreachable. Previously `serialize_config` always started from `DocumentMut::new()`, so every `doc.entry(k).or_insert(Item::Table(..))` genuinely inserted a `Item::Table` and the `expect`s could not fire. They can now.

The trigger is a root-level **inline** table for any section whose parser tolerates a non-table value by falling back to defaults. `parse_config` reads these through `Item::as_table`, which returns `None` for an inline table, so the file parses fine — and then the writer reaches for `as_table_mut()` on the inline value and aborts.

Repro (`x86_64-linux`, PR head):

```toml
wallet = { home_dir = ".scaffold/wallet" }

[scaffold]
version = "0.2.0"
# ... otherwise minimal, valid v0.2.0 config
```

```console
$ lgs basecamp setup --inspector
inspector build selected (.#bin-bundle-dir-inspector) — test-only, not a release artifact

thread 'main' panicked at src/config.rs:807:27:
wallet table
```

Confirmed for four sections, each aborting at its own `expect`:

| inline key at document root | panic site |
| --- | --- |
| `wallet = { … }` | `src/config.rs:807` `expect("wallet table")` |
| `framework = { … }` | `src/config.rs:815` `expect("framework table")` |
| `localnet = { … }` | `src/config.rs:827` `expect("localnet table")` |
| `circuits = { … }` | `src/config.rs:839` `expect("circuits table")` |

`src/config.rs:821` (`expect("idl table")`) is the same shape via `[framework]` + `idl = { … }`, and `ensure_subtable`'s `expect("parent is a table")` becomes reachable the same way for `modules = { … }` once the config actually carries a module (`basecamp modules` rather than `setup`).

Two things make this worth fixing before merge rather than filing as a follow-up. It is an abort with a backtrace, not a scaffold-level error — exactly the "raw stack trace with no scaffold-side hint" that `DOGFOODING.md` B1 calls a UX regression. And it lands on hand-edited files specifically, which is the population this change exists to be kind to; the file that gets a comment written into it is the file most likely to have been hand-shaped in other ways too.

The unparseable-input fallback doesn't cover it, since the file parses. Either coerce a table-like item into a `Table` before writing, or treat "modelled key exists but isn't a `Table`" as the same case as unparseable input and fall back to a fresh render — but please make it a returned error or an explicit fallback rather than an `expect`, since these are now reachable from file content.

### 2. `[basecamp.profiles]` is the one conditional emitter the sweep missed (`src/config.rs:932`)

`env` and `env_append` both got the `else { basecamp_table.remove(...) }` branch; `profiles` did not. Same defect the PR describes: clearing the profiles while `[basecamp]` still exists leaves the old tables stranded in the file.

Confirmed with a test in `config::tests`:

```rust
let original = minimal_v0_2_0()
    + "\n[basecamp]\nport_base = 41000\n\n[basecamp.profiles.alice]\nruntime_dir = \"/run/user/1000\"\n";
let mut cfg = parse_config(&original).unwrap();
cfg.basecamp.as_mut().unwrap().profiles.clear();
let updated = update_config(&original, &cfg).unwrap();
// fails: `[basecamp.profiles.alice]` is still in `updated`;
// serialize_config(&cfg) emits only `[basecamp]\nport_base = 41000`.
```

`update_config_clears_keys_reset_to_defaults` doesn't catch it because it sets `cfg.basecamp = None`, which short-circuits to `doc.remove("basecamp")` at line 979 and never exercises "section kept, profiles emptied". Worth extending that test with the mixed case — it is the vacuous-coverage shape the PR write-up already identified once.

While in there: `write_repo_ref` is only called under `if let Some(repo) = &cfg.basecamp_repo` / `cfg.lgpm_repo` with no `else` removing a stale `[repos.basecamp]` / `[repos.lgpm]`. Nothing sets either back to `None` today, so it is latent rather than a live bug, but it is the last gap in the "every conditional emitter has a removal branch" invariant the PR states.

### 3. An unparseable or unreadable `scaffold.toml` is silently overwritten

`update_config` does `existing.parse().unwrap_or_default()`, and `save_project_config` does `fs::read_to_string(&path).unwrap_or_default()`. The fallback policy is defensible, but it is silent and there is no backup, so a file that was one stray bracket away from valid is replaced by a from-scratch render — every comment and unmodelled section gone, no message. That is the exact loss the rest of this PR is built to prevent, and the user gets no signal it happened.

`unwrap_or_default()` on the read also collapses "no file yet" and "file exists but I couldn't read it" (permissions, I/O error) into the same path; the second case silently truncates a file that was there.

A one-line `eprintln!` naming the fallback would be enough for the parse case. For the read, distinguishing `NotFound` from other errors and propagating the rest would be better. `init`'s migration already writes a `scaffold.toml.bak`; matching that here would be the belt-and-braces option.

### 4. `DOGFOODING.md` isn't updated, and CONTRIBUTING requires it

`CONTRIBUTING.md:44` — "If the change affects user-facing behavior, update `README.md` and `DOGFOODING.md` in the same PR." `docs/commands.md` is updated (thank you — the `--inspector` paragraph is genuinely good), but nothing in the runbook is.

Concretely missing:

- **B1** drives `basecamp setup` with no flags. Nothing exercises `--inspector` / `--no-inspector`, the `[repos.lgpm].attr` realignment, or the persist-before-build ordering — and B1's "Expected Success Signals" already enumerate per-platform `attr` behavior, so the inspector attr belongs alongside it.
- **B5** covers `basecamp build` / `build-portable` without mentioning `--print-output`.
- Nothing anywhere asserts the new **comment-preservation contract**. This one matters most: it silently changes the behavior of *every* command that rewrites `scaffold.toml`, not just the ones this PR touches, and the failure mode it introduced mid-review (keys reset to defaults being stranded) is invisible unless a scenario looks for it. B1 already says to capture a `scaffold.toml` excerpt as evidence — a check that a comment placed before `setup` is still there afterward, and that a value reset to its default is actually gone, would be cheap and would have caught findings 1–3.

The "Minimum Rerun Guidance" table should probably also gain a row for changes to the config writer, since the blast radius is every config-writing command rather than just `basecamp`.

### 5. Minor: `basecamp_setup_inspector_persists_both_attrs_before_the_build` does real network + Nix work in `cargo test`

The doc comment frames this as local-only coverage that no-ops in CI, and that's accurate for CI. But on any developer machine that *does* have Nix — the machines most likely to be working on this code — a plain `cargo test` now shells out to `basecamp setup --inspector`, which clones the pinned basecamp repo into the shared cache root and starts a Nix build of a Qt-heavy closure. It's the only test in `tests/cli.rs` that reaches the network or the Nix store, and it uses the real default cache root rather than a temp one.

The assertion is worth keeping. Gating it behind an explicit opt-in env var, and pointing `LOGOS_SCAFFOLD_CACHE_ROOT` at the test's own tempdir, would make it safe to leave in the default run.

---

Findings 1 and 2 are code defects with reproductions; 3 and 4 are behavior/documentation gaps that follow from the same change. I'd hold approval on 1 in particular — an abort on a `scaffold.toml` that parses is a worse outcome than the comment loss the change fixes.

I could not complete a full `basecamp setup` here: the container's proxy answers `403` for Nix's `github:` flake inputs, and a direct-egress run gets past evaluation but then builds the Qt closure from source, which is the environment limitation `DOGFOODING.md` B1 documents as the usual B-series blocker. Everything above the Nix build — attr selection, `lgpm` realignment, persist ordering, config round-tripping — was verified against the real binary.

---

src/commands/basecamp.rs
```diff
@@ -283,6 +314,124 @@ fn format_flake_ref(repo: &RepoRef) -> String {
...
+        Some(true) => {
+            if basecamp_repo.effective_attr(system) == BASECAMP_ATTR_INSPECTOR {
+                return false;
+            }
+            if basecamp_repo.attr_platform.contains_key(system) {
+                basecamp_repo
+                    .attr_platform
+                    .insert(system.to_string(), BASECAMP_ATTR_INSPECTOR.to_string());
+            } else {
+                basecamp_repo.attr = BASECAMP_ATTR_INSPECTOR.to_string();
```
danisharora099: **Blocking: this write is lost whenever the project already has a per-platform map.**

`write_repo_ref` (`src/config.rs:1053`) emits only the inline map when `attr_platform` is non-empty and drops the scalar. So on a project like:

```toml
[repos.basecamp]
attr = { aarch64-darwin = "bin-macos-app" }
```

a Linux developer running `setup --inspector` gets `lgpm.attr = "cli-portable"` on disk but never `bin-bundle-dir-inspector`. The next plain `setup` defaults the scalar back to `app` and builds a dev basecamp against portable lgpm, which is the exact empty-UI mismatch this flag exists to prevent. A later `--no-inspector` reports "nothing to undo".

The unit tests only cover a map that already contains the current host, so this path is untested.

Suggested fix: when the map is non-empty, always write this host's key into the map and never touch the scalar. That also turns the `--no-inspector` branch below into a plain `remove(system)` with no scalar-repair heuristic.

src/commands/basecamp.rs
```diff
@@ -133,7 +134,16 @@ pub(crate) fn basecamp_for_project(project: Project, action: BasecampAction) ->
         BasecampAction::Develop { module, dev_shell } => {
             cmd_basecamp_develop(project, module, dev_shell)
         }
-        BasecampAction::Build { variants, module } => cmd_basecamp_build(project, variants, module),
+        BasecampAction::Build {
+            variants,
+            module,
+            print_output,
+        } => {
```
danisharora099, line 141: **Blocking: `--print-output` is a no-op on this path.**

The only nix spawn reachable from `cmd_basecamp_build` is `run_build_portable_nix` (around line 2087), which calls `cmd.output()` and never consults `print_output_enabled()`. Output is captured and discarded on success, and only trimmed stderr surfaces on failure, exactly as before the flag existed. The help text is also wrong on the other half: `build` never writes a `.scaffold/logs` file.

`basecamp_build_accepts_print_output_flag` only checks `--help`, so nothing catches this.

Either route that spawn through `run_logged` the way `setup` and `install` do, or drop the flag from this PR.

DOGFOODING.md
```diff
@@ -1060,21 +1060,42 @@ test -f scaffold.toml || "$SCAFFOLD_BIN" init
...
 "$SCAFFOLD_BIN" basecamp doctor
 "$SCAFFOLD_BIN" basecamp doctor --json
 "$SCAFFOLD_BIN" basecamp setup
+
+# Inspector stack opt-in, and back. Each grep is the evidence for the step
+# above it; run them even if the build is out of reach in this environment,
+# since the attrs are written before the build starts.
+"$SCAFFOLD_BIN" basecamp setup --inspector
+grep -n 'attr' scaffold.toml
+cat .scaffold/state/basecamp.state         # basecamp_bin must have moved
```
danisharora099: This assertion is false on Linux, and a dogfooder following it will file a bug that is not one.

`build_basecamp_app` always writes the out-link to `<cache>/basecamp/<pin>/app-result` regardless of attr, and both the 0.2.x dev launcher and the portable bundle resolve to `bin/LogosBasecamp`. So `basecamp_bin` is byte-identical before and after `--inspector`.

The real hazard is the inverse of what the docs describe: the shared symlink gets re-pointed, so a consumer holding the old path silently execs the new build, and the displaced stack loses its GC root so the next toggle back rebuilds. Keying the out-link by attr (`app-result-<attr>`) would make this line true and let both stacks coexist.

docs/basecamp-module-requirements.md
```diff
@@ -22,6 +22,26 @@
...
+- **It is a nix out-link, so it is read-only**, and on a portable build the real ELF sits beside it as a dot-prefixed sibling (`.LogosBasecamp.elf`). A harness that needs a writable copy, or one that probes for a specific file layout, has to account for that — copy the tree out rather than expecting to modify it in place.
+- **It moves when the stack moves.** `setup --inspector` rebuilds at a different flake attr, so `basecamp_bin` points somewhere else afterwards. Re-read the file after any `setup`; caching the path across a stack change silently execs the previous build.
```
danisharora099: Same as the DOGFOODING comment: the path does not change between stacks today, because the out-link is keyed by pin only and both stacks resolve to `bin/LogosBasecamp`. Either key the out-link by attr so this becomes true, or reword this bullet to say the symlink is re-pointed in place.

src/commands/basecamp.rs
```diff
@@ -283,6 +314,124 @@ fn format_flake_ref(repo: &RepoRef) -> String {
...
+                }
+                return false;
+            }
+            // Drop this host's map entry, so whatever the scalar says applies
+            // to it again. Then fix the scalar only if what now resolves is
+            // still the inspector (it was the scalar that selected it) or is
+            // nothing at all (the map was the only source for this host).
+            basecamp_repo.attr_platform.remove(system);
+            let resolved = basecamp_repo.effective_attr(system);
+            if resolved.is_empty() || resolved == BASECAMP_ATTR_INSPECTOR {
```
danisharora099: The round trip loses a hand-set portable attr, silently.

Start from `attr = "bin-appimage"`, `lgpm.attr = "cli-portable"`. `--inspector` overwrites the scalar. `--no-inspector` then lands here, sets `app`, and `align_lgpm_attr` drags lgpm to `cli`. The user is now on the dev stack with no note, because the "left alone" note above only fires when the effective attr was not the inspector. A per-host map entry gets the same treatment: deleted rather than restored. `inspector_flags_touch_only_the_current_host_entry` asserts this outcome as correct, and `docs/commands.md` says hand-set values survive.

Restoring the previous value would mean remembering it somewhere, so I am not asking for that here. But the docs should not claim the opposite, and a note when `--no-inspector` lands on a different stack than the user started from would help.

src/commands/basecamp.rs
```diff
@@ -211,13 +225,30 @@ fn cmd_basecamp_setup(mut project: Project) -> DynResult<()> {
...
+    // to be interrupted or to fail on an unrelated step, and losing the choice
+    // there would leave the next plain `setup` silently back on the other
+    // stack — the same quiet mismatch the flag exists to prevent. The attrs
+    // alone are inert config; nothing downstream depends on the build having
+    // finished. Keyed on the stack having actually moved rather than on a flag
+    // merely being present: a `--no-inspector` that declined to touch a
+    // hand-set attr has nothing to persist, and rewriting `scaffold.toml` to
+    // say exactly what it already said is churn the user did not ask for. A
+    // plain `setup` still writes once, at the end.
+    if stack_changed {
```
danisharora099, line 248: The comment above says nothing downstream depends on the build having finished. That is not quite right: `launch`, `paths`, and `doctor` derive the XDG subpath and variant key from config, but take the binary from `basecamp.state`, which does not record the attr.

After an interrupted `setup --inspector`, config says portable while the binary on disk is still the dev launcher. `launch` seeds portable profiles and execs the dev app, and `doctor` reports PASS because the path exists.

Recording the built attr in `basecamp.state` and having `doctor` compare it to `effective_attr` would close this, and also answers the "which build is this?" question from the `basecamp_bin` comments.

Smaller: on a toggle this early save plus the unconditional save at the end write `scaffold.toml` twice with identical bytes. Guarding the trailing save with `!stack_changed` would keep it to one.

docs/commands.md
```diff
@@ -177,13 +177,14 @@ logos-scaffold help
...
...ger library is what the app uses to read the modules `lgpm` installed. Existing projects keep whatever they pinned in `scaffold.toml` — bumping means editing both pins and re-running `setup`. Runtime config (`port_base`, `port_stride`) is in `[basecamp]`. Setup also records what it built as `key=value` lines in `.scaffold/state/basecamp.state` — `pin`, `basecamp_bin`, `lgpm_bin` — which is the supported way to find the binary from a script. Use `basecamp_bin` verbatim rather than reconstructing a path: which file under the nix output is the right entry point differs by basecamp generation and by stack — on a dev `#app` build it is a `/bin/sh` launcher that exports `QT_PLUGIN_PATH` / `QML2_IMPORT_PATH`, and picking the raw binary beside it yields an app that cannot find its Qt platform plugin or QML imports. `setup` resolves that for you. `lgs doctor` prints the same path for humans. [basecamp-module-requirements.md](./basecamp-module-requirements.md#what-setup-leaves-behind--scaffoldstatebasecampstate) — which `basecamp docs` prints offline — carries the full contract for script consumers, including the read-only out-link layout and the fact that the path moves when the stack does.
```
danisharora099: Two corrections in this paragraph:

- `lgs doctor` does not print the binary path. Only `lgs basecamp doctor` reads `basecamp.state`.
- The `--inspector` bullet says the flag persists "overriding a per-platform attr map, since the inspector output is built for every supported system", and a few sentences later says both flags "edit only the current host's entry". The code does the second. The first reads as if the whole map is replaced.

src/api/mod.rs
```diff
@@ -645,7 +645,14 @@ impl CreateProjectOptions {
 #[derive(Clone, Debug)]
 pub enum BasecampCommand {
     /// Build/install the pinned basecamp binary and seed profiles.
-    Setup,
+    Setup {
```
danisharora099, line 654: This changes `BasecampCommand::Setup` from a unit variant to a struct variant on the public API, so any embedder writing `BasecampCommand::Setup` stops compiling. Fine if intended, but it deserves a version bump or at least a note, since nothing in `tests/` exercises the API surface and CI will not notice.

tests/cli.rs
```diff
@@ -3340,6 +3340,99 @@ fn basecamp_build_portable_inside_empty_project_emits_hint_to_capture_first() {
...
+/// Opt-in via `LOGOS_SCAFFOLD_E2E_NIX=1`, because on a host that *does* have
+/// nix this shells out to a real `setup`: a clone of the pinned basecamp repo
+/// and the start of a Qt-heavy closure build. That is not something a plain
+/// `cargo test` should do, and it is the only test in this file that reaches
+/// the network or the nix store. When it does run, `LOGOS_SCAFFOLD_CACHE_ROOT`
+/// points at the test's own tempdir so the clone lands there and not in the
+/// developer's shared `~/.cache/logos-scaffold` — B1's "do not pollute the
+/// user's home" rule applies to the test suite too.
+#[test]
+fn basecamp_setup_inspector_persists_both_attrs_before_the_build() {
```
danisharora099, line 3468: This could run unconditionally in CI. `ensure_nix_present` is only a `which("nix")`, and the file already stubs binaries on PATH via `write_wallet_stub`. A stub `nix` plus a `[repos.basecamp]` source that fails to clone immediately would prove the write ordering in milliseconds, and the two `contains` assertions stay as they are. Then the opt-in gate and the nix caveat in the PR description both go away.

DOGFOODING.md
```diff
@@ -1060,21 +1060,42 @@ test -f scaffold.toml || "$SCAFFOLD_BIN" init
...
+"$SCAFFOLD_BIN" basecamp setup            # no flag: must stay on the inspector
+grep -n 'attr' scaffold.toml
+"$SCAFFOLD_BIN" basecamp paths alice | grep -i module_root
+"$SCAFFOLD_BIN" basecamp setup --no-inspector
+grep -n 'attr' scaffold.toml
+"$SCAFFOLD_BIN" basecamp setup --inspector --no-inspector   # must be rejected
```
danisharora099: This line is meant to exit non-zero, but unlike the other expected-failure lines in this scenario it has no `|| true`. Under a `set -e` capture script the block aborts here. Same for the `setup --inspector` line above on a host where the clone fails: the comment says to run the greps regardless, but the greps come after a command that may exit non-zero.

danisharora099 (changes requested):
Reviewed at `f207de3`.

The in-memory logic in `apply_inspector_selection` and `align_lgpm_attr` is careful and the unit tests around it are good. The problems are at the seams: where that logic meets the config writer, the build path, and the docs. Two of them undercut the PR's own goal, so I am holding on those.

**Blocking**

1. `--inspector` is silently not persisted when the project has a per-platform attr map without an entry for this host. The config writer drops the scalar whenever a map exists. Inline comment on `basecamp.rs:354`.
2. `--print-output` on `build` / `build-portable` does nothing. The module build path never reads the flag. Inline comment on `basecamp.rs:141`.

**Should fix here, all about the same state file**

3. Docs and runbook say `basecamp_bin` moves when the stack moves. It does not; the out-link is keyed by pin only. Inline on `DOGFOODING.md:1073` and `basecamp-module-requirements.md:39`.
4. The `--inspector` then `--no-inspector` round trip replaces a hand-set portable attr with the default. Inline on `basecamp.rs:376`.
5. Persisting before the build can leave config and `basecamp.state` on different stacks, and `doctor` cannot tell. Inline on `basecamp.rs:248`.

**Minor**

- `docs/basecamp-module-requirements.md:249`: the `LOGOS_DATA_DIR` row lists three portable attrs. The inspector attr now qualifies too.
- `BasecampCommand::Setup` changed from a unit variant to a struct variant on the public API with no version bump. Inline on `api/mod.rs:648`.
- The E2E test could run in CI with a stub `nix` on PATH. Inline on `tests/cli.rs:3405`.
- The last line of the new DOGFOODING block has no `|| true`. Inline on `DOGFOODING.md:1079`.

The existing verification on this thread ran on a project with a plain scalar attr, so it does not exercise point 1, and nobody checked that `--print-output` actually streamed anything. Happy to re-review once 1 and 2 are in.

src/commands/basecamp.rs
```diff
@@ -283,6 +318,156 @@ fn format_flake_ref(repo: &RepoRef) -> String {
...
+                }
+            }
+            // Undoing `--inspector` restores the default stack, not whatever
+            // the user had before it — scaffold does not remember a displaced
+            // hand-set attr like `bin-appimage`, and inventing one would be a
+            // guess. What it can do is not leave the move silent: landing on a
+            // different stack than the user started from is exactly the sort
+            // of quiet change this feature exists to surface.
+            let restored = basecamp_repo.effective_attr(system);
+            if restored != BASECAMP_ATTR {
```
danisharora099, line 408: Non-blocking. This condition fires in the opposite case from the one it was added for. The lossy round trip lands on the default after displacing a hand-set value, and there `restored == BASECAMP_ATTR`, so the note stays silent. In the scalar branch above `attr` is always set to the default, so the note never fires there. In the map branch it fires only when the scalar is non-default, which cannot be serialized alongside a map. In practice it is close to unreachable.

Printing the note unconditionally on a successful undo would be simpler and would actually cover the case it describes. Fine as a follow-up.

src/commands/basecamp.rs
```diff
@@ -320,7 +505,14 @@ fn build_basecamp_app(
...
 ) -> DynResult<PathBuf> {
-    let link = out_dir.join("app-result");
+    // Key the out-link by attr, not by pin alone. A single shared
+    // `app-result` would be re-pointed by every stack toggle: a consumer
+    // holding the path from a previous `setup` would silently exec the other
+    // build, and the displaced stack would lose its GC root, so toggling back
+    // would rebuild from scratch. Per-attr links let both stacks coexist under
+    // one pin and make `basecamp_bin` genuinely move when the stack does,
+    // which is what `basecamp.state`'s consumers are told to rely on.
+    let link = out_dir.join(format!("app-result-{attr}"));
```
danisharora099, line 515: Non-blocking. `lgpm-result` in `build_lgpm` just below is still one shared link across `cli` and `cli-portable`, so it has the same re-pointing behaviour this change fixes for basecamp: a toggle replaces the other stack's lgpm and drops its GC root. Much smaller cost since lgpm builds in seconds, but keying it the same way (`lgpm-result-<attr>`) would keep the two consistent and make `lgpm_bin` in `basecamp.state` distinct per stack too. One-liner for a follow-up.

danisharora099 (approved):
Re-reviewed at `27ffa8b`. Both blockers are fixed and covered by tests that fail against the previous code, and the out-link change makes the docs true rather than just rewording them. Thanks for taking the E2E test suggestion further than I asked, running it in CI unconditionally is the right outcome.

Verified locally: `cargo fmt --check` clean, lib tests green, the three new CLI tests pass in about a second. One CLI test failed on my first full run (`deploy_program_path_json_includes_program_id`, a ureq panic) but passes on rerun and on master, so that is a pre-existing flake unrelated to this PR.

Two non-blocking notes inline. Deferring the `basecamp.state` attr field to a follow-up is fine by me. The version bump is a maintainer call, and the ADR entry gives them what they need to make it.

PR #267, opened 2026-09-04
fix(config): preserve scaffold.toml comments on every rewrite

fryorcraken:
Split out of #266, which carried this alongside unrelated `basecamp --inspector` work. That PR keeps the inspector feature; this one is the comment-preservation fix on its own.

## What was wrong

`save_project_config` rendered `scaffold.toml` from an empty `toml_edit` document, so **every user comment was dropped on every write**. Exactly two commands persist config, both in `src/commands/basecamp.rs` — `basecamp setup` (`cmd_basecamp_setup`) and `basecamp modules` (`cmd_basecamp_modules`, module capture) — and both silently reformatted the file and threw the comments away.

That matters more than cosmetics. These comments are routinely load-bearing — why a `runtime_dir` override exists, why a pin is held back — so losing them silently re-opens the bug they were written to document.

## The fix, and the subtlety it introduces

Seeding the writer with the parsed existing file preserves comments, key ordering, and sections scaffold doesn't model. But it changes the writer's contract from **"render the model"** to **"merge the model over the file"**, and that inverts the meaning of every conditional emit.

The codebase's optional keys were all written as `if non_default { assign }` with no `else`. Under a render that's correct — a skipped assignment means the key is simply absent from the output. Under a merge it is a silent data bug: the skipped assignment leaves the *old* value in the file, so a value reset to its default quietly persisted and was read back on the next load. Every one of those now has a matching removal branch, and every map-backed table (`modules`, `basecamp.env`, `profiles`, `run.profiles`, …) retains down to the keys the config still carries.

The inverse rule is the sharper one, because getting it wrong destroys user content rather than merely staling it:

> **A removal may be driven by the model only where the parser is _total_** — where it errors on anything it cannot represent, so that "absent from the model" genuinely means "absent from the file".

Several of scaffold's parsers are not total, and a blanket `remove` driven by them would delete whatever the user wrote:

- `parse_basecamp_runtime` returns `None` for a `[basecamp]` whose keys are all unmodelled, *skips* an `env_append` entry whose list is empty, and *drops* a fully-default `[basecamp.profiles.<name>]`.
- `parse_run` maps an all-default `[run]` (or `[run.watch]`) to the default value, indistinguishable from absent.

For those the model's key set is a strict subset of the file's. `remove_owned_child`, `clear_owned_basecamp_keys` and `prune_run_profiles` therefore clear only the keys scaffold owns and drop the table only if nothing is left. Blanket removal survives only where the parse genuinely is total — `[basecamp.env]` and a profile's `env`, via `parse_string_map`, which rejects every key it cannot model.

The rule and its rationale are documented on `write_config_into`, where the next person adding an optional key will hit it.

## Two defects found in review (fixed in `e2690eb`)

`@weboko` reviewed the above and found two holes in it. Both were real; both are now fixed.

**1. The writer aborted on a `scaffold.toml` that parses fine.** Every parser in `config.rs` reads its section through `Item::as_table`, which is `None` for an *inline* table. So `wallet = { home_dir = "…" }` is a valid file that the reader accepts (falling back to the default). The writer then reached the same key via `entry(...).or_insert(...)` — which returns the **occupied** item, not a fresh insert alongside it — so `.as_table_mut()` was `None` and `.expect("wallet table")` aborted the process with a backtrace. `DOGFOODING.md` B1 calls a raw panic on file content a UX regression, and this one was reachable from a file a person could reasonably write.

The repro turned out sharper than first reported: the panic needs the inline table to be the *only* definition of that key. With a `[wallet]` header also present, `or_insert` finds the real table and the write instead emits a duplicate key. Both outcomes are wrong.

Fixed with a `coerce_to_table` helper that promotes an inline table into a real `Table` in place — every key survives the syntax change, unmodelled ones included — and replaces a non-table scalar (nothing to keep, and the parser already ignored it) with an empty table. Every reachable site routes through it now, root sections directly and nested ones via `child_table` / `ensure_subtable`, which carried the same `expect`. No `expect` on file content remains in the writer.

**2. An unparseable `scaffold.toml` was overwritten with no diagnostic.** `update_config` did `existing.parse().unwrap_or_default()`, so a file one stray bracket from valid was silently replaced by a from-scratch render — every comment and unmodelled section gone, exit 0, no message. That is precisely the loss this PR exists to prevent.

The fallback stays (refusing would wedge every command that persists config), but it is no longer quiet: `update_config_reporting` returns a `RewriteOutcome`, and `save_project_config` warns on stderr naming the parse error and copies the original to a timestamped sibling `scaffold.toml.bak-YYYY-MM-DD-HHMMSS_NNNNNNNNN` before overwriting.

This is **fail-closed**: if the backup cannot be written, the whole write aborts and `scaffold.toml` is left exactly as the user left it. Re-running a command is recoverable; the comments in that file are not. The timestamp means each destructive rewrite keeps its own copy rather than the second one colliding with the first, and the names sort oldest-first in a directory listing.

The stamp runs to nanoseconds, which is load-bearing rather than gratuitous precision. At millisecond resolution two consecutive `save_project_config` calls genuinely share a stamp — that made the suite flaky, and would equally have failed a legitimate back-to-back write in the field. Nanoseconds make a collision the anomaly it was assumed to be, so the name is claimed in exactly one attempt via `create_new` and a taken name is a hard abort, with no retry or suffix in the picture. (`sync_all` runs before the overwrite, so a close-time IO error can't be reported as a successful backup. The timestamp is computed in-tree via Hinnant's `civil_from_days` plus `localtime_r`, rather than adding a date crate to name a file.)

> **Scope correction (see `dd1c2da`).** The paragraphs above describe this machinery as user-facing recovery. It isn't, and the docs that said so have been corrected. Both `save_project_config` callers get their `Project` from `load_project`, which parses the same file and hard-errors first — so `RewriteOutcome::RenderedFresh` is reachable only if the file changes between load and write. On a broken file the user gets `error: invalid scaffold.toml: …`, exit 1, no warning and no `.bak-*`. The backup path stays as the TOCTOU guard it actually is; README, `docs/commands.md` and DOGFOODING B1 now document the refusal that ships. Making it genuinely reachable means letting the config-writing commands recover from an unparseable file instead of `load_project` bailing — a larger change, and a separate PR.

## Three further defects found in review (fixed in `dd1c2da`)

A second review pass by `@weboko` found three more. All three were real.

**1. (Blocking) Promoting a commented inline table wrote a file that does not parse.** The `coerce_to_table` fix above removed the panic, but replaced it with something worse at the root-level sites. `child_table` clears the promoted key's decor first; the seven root sections and `ensure_subtable`'s parent coercion called `coerce_to_table(doc.entry(..).or_insert(..))` directly — and `coerce_to_table` takes `&mut Item`, so it *structurally cannot* reach the key, which is exactly where a preceding own-line comment lives. The result:

```toml
# why this override exists
circuits = { version = "0.4.1" }
```

rewrote to

```toml
[# why this override exists
circuits ]
```

— invalid TOML, written with exit 0 and a "captured modules" success message, after which every later command fails on the file scaffold just wrote. That is strictly worse than the panic it replaced, which at least aborted before writing and left the file intact.

The guard is hoisted out of `child_table` into `section_table`, keyed on a small `TableParent` trait over `DocumentMut` and `Table` so both root and nested sections route through one place and a new call site cannot miss it. `coerce_to_table` now has exactly one caller.

Clearing the decor is *not* the whole fix, though: it keeps the file parseable while deleting the comment, which is the thing this PR exists to preserve. So `section_table` lifts the comment lines out of the key's decor and re-homes them onto the promoted table's header decor — `# why this override exists` ends up on its own line above a clean `[circuits]`, and only the inline-specific whitespace is dropped.

**2. Trailing comments on scaffold-owned keys were deleted on every write.** `table[k] = value(v)` installs a fresh `Item`, discarding the old value's suffix decor — which is where `toml_edit` keeps a trailing comment. Own-line comments live in the key's decor and survived, which is why this hid behind the tests that did exist.

```diff
-port = 3040  # chosen to dodge the corp VPN range
-risc0_dev_mode = true  # keep proofs cheap locally
+port = 3040
+risc0_dev_mode = true
```

It applies to keys whose value is rewritten *identically*, so a no-op run stripped them too. And `pin = "aa2377…"  # held back until the new IDL spec lands` is precisely the load-bearing comment this PR's own framing cites as its reason to exist. Fixed with `set_preserving_suffix`, which carries the old suffix onto the new value. Carrying the decor rather than skipping unchanged assignments is deliberate — the latter handles only the majority and leaves the changed-value case still losing its comment.

> **Correction (`c680438`).** This paragraph originally claimed "all 38 assignment sites route through it". That was wrong: five array- and inline-table-valued sites were missed, so the bug was unfixed for `run.watch.include` / `.exclude`, both `post_deploy` sites, and `basecamp.env_append`. All 42 sites now route through the helper.

**3. The docs described a feature the binary doesn't expose.** Covered by the scope correction above.

## Two more defects, found reviewing the review (fixed in `c680438` and `86492d2`)

The fixes above were themselves reviewed, which turned up two more. Both were real, and the second is the most serious defect in the whole PR.

**1. Five assignment sites still dropped trailing comments (`c680438`).** The "all 38 sites" claim above was simply false. The five missed are array- or inline-table-valued — `run.watch.include`, `run.watch.exclude`, `run.post_deploy`, `run.profiles.<n>.post_deploy`, `basecamp.env_append.<k>` — which is why a conversion pass shaped around scalars skipped them. `set_preserving_suffix` takes an `Item` and `Item::as_value` covers arrays and inline tables, so they needed no special handling, only to be called. Demonstrated by a test where `include`'s comment died while the scalar `debounce_ms` beside it survived.

The same commit narrows the salvaged suffix to its **first line**. toml_edit's value suffix runs to the next key, so `port = 3040  # a` followed by an own-line `# b` gives `port` a suffix of `"  # a\n# b"` — and `# b` belongs to the *next* key. Carrying it wholesale round-trips today but would relocate `# b` as soon as the value rendered differently.

**2. An inline-table section was read as absent, silently discarding the user's values (`86492d2`).** This one is a genuine data-loss bug, and it is in the **reader**, not the merge:

```toml
wallet = { home_dir = "CUSTOM/PATH", note = "keep me" }
```

Every parser in `config.rs` reached its section through `Item::as_table`, which is `None` for an inline table. So this valid file was read as *no `[wallet]` section at all*, the model took scaffold's default, and the next rewrite faithfully wrote that default over the user's override — exit 0, no warning. The override was silently dead.

Note what this means for the earlier fixes: the promotion machinery was never the problem here. `section_table` / `coerce_to_table` were already promoting correctly and keeping unmodelled keys; by the time the writer ran, `cfg` simply no longer carried the user's values. The fix is a `section_like` helper reading through `Item::as_table_like`, which accepts both forms, applied across the reader. (`parse_attr_platform` already did this — the codebase was inconsistent rather than uniformly wrong.) `ensure_subtable` also had to carry a salvaged comment down to its first child, since `set_implicit(true)` means no parent header is emitted to hang it on.

Verified end-to-end on a real project: `home_dir = "CUSTOM/PATH"` and the unmodelled `note = "keep me"` both survive a rewrite, under exactly one `[wallet]` definition, with the comment above a clean header.

Scope worth stating: the silent path is specific to **optional or defaulted** sections. An inline `repos` fails loudly with `missing [repos.lez]` rather than losing data.

**Why the tests missed it.** The table-driven inline-table test asserted `rewritten.contains("keep me")` — which passed *because of* the bug, since the user's abandoned text was still sitting in the file. An assertion meant to prove preservation was satisfied by the loss. Those rows now walk the reparsed document and assert one definition per section, so they test the semantic property rather than substring presence. Two rows changed as fallout: `modules` needed a well-formed entry (a bare `note` is an invalid module and is correctly rejected), and the `framework.idl` row was dropped — it built a root-level `idl` key scaffold does not model and asserted nothing about promotion. Nested coverage moved to a dedicated test.

## Testing

- `cargo build`, `cargo fmt --check` — clean.
- `cargo test --all-targets` — **819 passed, 0 failed**.
- `cargo clippy --all-targets` — no new warnings (unchanged at 64, none in the touched files).

Beyond the basic preservation cases, the tests pin the two invariants that make the merge contract safe, deliberately in tension with each other:

- a rewrite of a config whose optional sections have all been reset to defaults must agree with a fresh render (catches a missing `else { remove }`), and
- a section holding only unmodelled keys, an empty `env_append` list, a profile of only unmodelled keys, and unmodelled keys sitting inside a table scaffold owns must all survive (catches an over-eager removal).

For the third review round, the sharpest lesson was that **the inline-table tests passed against corrupting code.** They exercised promotion but planted no comment on the promoted key, so the one piece of decor that breaks the header was never present. `with_inline_section` now always plants one, which turns both existing tests into genuine regressions, and a further test pins the *rendered shape* (`# comment\n[circuits]`) rather than just asserting the comment appears somewhere — the weaker check passes on `[# comment\ncircuits ]`. One row of the table is exempted with a comment explaining why: for `framework.idl` the helper produces a root-level `idl` key rather than a child of `[framework]`, so its comment is orphaned by the fixture, not the writer; the real child-table promotion is covered separately. The trailing-comment test covers three shapes deliberately — a value that changes, one rewritten identically, and a non-string — so the fix cannot be quietly string-specific. All verified to fail against the pre-fix writer.

For the two earlier review defects: a table-driven test covering an inline table at every reachable section, an inline-child-table test, and a non-table-scalar test — all three verified to panic against the pre-fix writer — plus `save_project_config` disk tests: backup written and byte-identical to what was discarded, a second rewrite keeping its own backup, no backup on the merge path, an unwritable directory aborting the rewrite with `scaffold.toml` untouched, a taken name aborting, the backup-name shape, and `civil_from_days` against known dates (leap days, 1900 vs 2000, pre-epoch). One test claims 50 backup names in a row and asserts each is distinct and a bare unsuffixed stamp — since name-claiming has no retry, an `Ok` proves the first derived name was free, so a regression to coarser precision fails there rather than resurfacing as load-dependent flakiness.

## Docs

Per `CONTRIBUTING.md`, `README.md` gains a section on how scaffold edits `scaffold.toml`, and `DOGFOODING.md` B1 gains the runbook check `@weboko` asked for. B1 is the only scenario that drives a real `save_project_config`, and those checks are the contract these functions exist to hold:

- a planted own-line comment and unmodelled `[team.notes]` section must survive two rewrites;
- a **trailing** comment on a scaffold-owned key must survive too — it lives in a different place to an own-line one and fails independently;
- a value reset to its default must actually be gone from the file;
- an inline table with a comment above it must come back as a clean `[section]` with the comment on the line above — checked by reparsing, not just grepping;
- a broken file must produce a hard error, an untouched file, and **no** backup.

The inline-table block carries a warning about two fixture traps, because a first attempt at it passed against the corrupting code twice over: the inline form must replace *every* `[circuits]` stanza `init` writes (two definitions of one key is a duplicate-key error, and the writer then targets the real table and promotes nothing), and it must sit above the first `[header]` (a bare `circuits = { … }` after `[localnet]` is a key *inside* `[localnet]`). Verified end-to-end against the built binary on a real `lgs init` project.


weboko (commented):
Reviewed at `0efd792`. I built the branch, ran the unit suite (623 pass, `cargo fmt --check` clean), and exercised the write path end-to-end through `basecamp modules --path` (the one config-persisting command that doesn't need a full basecamp nix build) on a real `lgs init` project.

The core fix works and I could confirm it against a file a person actually wrote: an own-line comment and an unmodelled `[team.notes]` section both survive repeated rewrites, and the inverse half holds too — a hand-deleted `port_base` is not resurrected on the next write. The merge-vs-render analysis in `write_config_into`'s doc comment is the right framing and the totality rule is worth the space it takes.

Three things came out of the run. The first is blocking.

---

## 1. (Blocking) Promoting a root-level inline table corrupts `scaffold.toml` when the key carries a comment

`coerce_to_table` fixes the panic, but the promotion it replaces it with emits structurally invalid TOML if the inline-table key has an own-line comment above it. The comment gets absorbed into the section header.

Repro on a fresh `lgs init` project — `[wallet]` replaced by the inline form the reader already tolerates:

```toml
# why this override exists
wallet = { home_dir = ".scaffold/w", note = "mine" }
```

After `lgs basecamp modules --path mods/foo.lgx` (exit **0**, "captured modules" printed):

```toml
[# why this override exists
wallet ]
home_dir = ".scaffold/wallet"
note = "mine"
```

The file no longer parses. Every subsequent scaffold command fails with `error: invalid scaffold.toml: TOML parse error`, and because `load_project` is the gate (see #2), there is no way back in without hand-editing. I reproduced the same corruption for `wallet`, `localnet`, `circuits` and `framework`; `repos` / `modules` reach the same code through `ensure_subtable`. Without the comment the output is valid but still cosmetically off — `[wallet ]`, with a stray space.

This is the failure mode `child_table` already documents and guards against:

> Carrying an inline key's decor across renders `"[\nframework.idl ]"` — not the section the user wrote, and not even the same file.

`child_table` (src/config.rs:1567) clears `key.leaf_decor_mut()` / `key.dotted_decor_mut()` before promoting. The root-level sites do not: src/config.rs:858, 923, 931, 940, 952, 997, 1197 all call `coerce_to_table(doc.entry(...).or_insert(...))` directly, and `coerce_to_table` takes `&mut Item`, so it structurally cannot reach the key's decor — that decor is where a preceding own-line comment lives. `ensure_subtable`'s parent coercion (src/config.rs:1592) has the same gap.

Suggest hoisting the guard out of `child_table` into a helper keyed on `(&mut Table | &mut DocumentMut, name)` and routing the root sections and `ensure_subtable`'s parent through it, so the decor reset can't be missed at a new call site. Worth a test that asserts the rewritten output *reparses*, not just that it contains the expected keys — the current inline-table tests (`update_config_survives_an_inline_table_for_every_modelled_section`, `update_config_survives_an_inline_child_table`) pass on a file that a later `parse_config` would reject, because no comment is planted on the promoted key.

This matters more than the panic it replaces: the panic aborted before writing and left the file intact, whereas this writes corruption and reports success.

## 2. (Blocking, mostly documentation) The unparseable-file fallback can't be reached, but README and `DOGFOODING.md` document it as user-facing behaviour

The backup, the fail-closed abort, and the nanosecond stamps are careful work, but `RewriteOutcome::RenderedFresh` is unreachable from either writer in practice. Both callers of `save_project_config` — `cmd_basecamp_setup` (src/commands/basecamp.rs:267) and `cmd_basecamp_modules` (src/commands/basecamp.rs:2123) — get their `Project` from `load_project`, which runs `parse_config` on the same file and hard-errors first. So the only window in which `save_project_config` sees an unparseable `scaffold.toml` is a TOCTOU one: the user breaks the file after load and before write.

Running the B1 steps this PR adds confirms it:

```
$ printf '\n[[[ broken\n' >> scaffold.toml
$ lgs basecamp setup
error: invalid scaffold.toml: TOML parse error at line 51, column 3
$ echo $?
1
$ ls scaffold.toml.bak-*
ls: cannot access 'scaffold.toml.bak-*': No such file or directory
```

Same for `basecamp modules`. The runbook tells the reviewer to expect a `could not be parsed` warning on stderr and exactly one `.bak-*`, then two after the second break, then a non-zero exit with an untouched file from the read-only-directory check. None of that happens — the three added negative-path blocks in B1 will be reported as failures by anyone who runs them. The README section ("If the file cannot be parsed as TOML — a stray bracket from a hand-edit — scaffold cannot merge into it, so it writes a fresh one instead") describes behaviour a user cannot observe; what they actually get is a hard error.

Two coherent resolutions, either is fine:

- Keep the machinery as the TOCTOU safety net it is, and scope the docs to match — drop the three negative-path blocks from B1 and the README paragraph, or reword them as an internal invariant rather than a user-facing recovery path.
- Or make it genuinely reachable, by letting the two config-writing commands recover from an unparseable file instead of `load_project` bailing. That is the larger change and probably a separate PR.

Worth deciding explicitly, because right now the PR body, the README and the runbook all describe a feature that the shipped binary doesn't expose.

## 3. (Non-blocking) Trailing comments on scaffold-owned keys are dropped

Every same-line comment on a key scaffold owns is deleted, including on keys whose value doesn't change:

```diff
-build = "nix-flake"  # trailing on repos key
-port = 3040  # chosen to dodge the corp VPN range
-risc0_dev_mode = true  # trailing on unchanged bool
+build = "nix-flake"
+port = 3040
+risc0_dev_mode = true
```

`table["k"] = value(v)` installs a fresh `Item` and discards the existing value's suffix decor, which is where toml_edit keeps a trailing comment. Own-line comments survive because they live in the key's decor, which is left alone.

I'd flag this because the PR's own framing is what makes it matter: `pin = "aa2377…"  # held back until the new IDL spec lands` is exactly the load-bearing comment the PR cites as the reason this fix exists, and it is silently lost. The README says comments are "left exactly where they were" and docs/commands.md says they "survive", so the docs currently overclaim.

`update_config_keeps_a_comment_on_a_key_it_reassigns` (src/config.rs:2474) doesn't catch it — it plants the comment on the line above `[repos.spel]`, i.e. in the section header's decor, not on a reassigned key. Extending that test to a `pin = "…"  # comment` form would pin the real case.

Cheapest fix is probably to carry the old value's decor onto the new one on reassign; skipping the assignment when the rendered value is unchanged would also handle the majority of it.

---

Happy to re-check once 1 is fixed. 2 and 3 are both partly a docs question, so your call on scope — but I don't think the README and B1 text should ship describing the unparseable-file recovery as-is.


---

weboko (commented):
Re-reviewed at `c680438`, focused on the two commits added since my last pass (`dd1c2da`, `c680438`). I rebuilt the branch, ran the suite (`817 passed, 0 failed`, `cargo fmt --check` clean, clippy surfaces nothing new in the code added here beyond one `as i64` redundant-cast nit at `src/project.rs:313`, in a repo that already carries 60-odd warnings), and re-ran the relevant B1 blocks end-to-end against the built binary through `basecamp modules --path` on a real `lgs init` project — the one config-persisting command that does not need a full basecamp nix build. (`basecamp setup` itself is out of reach here: nix flake fetching answers 403 in this container, as B1's preconditions anticipate.)

**All three findings from the previous round are fixed, and I could confirm each against a file rather than only against the tests.**

1. Promoting a commented inline table now writes a clean section. On a fixture with `# LOAD-BEARING: …` above `circuits = { version = "0.4.1", note = "hand-written" }`, the rewrite exits 0 and produces the comment on its own line above `[circuits]`, the file reparses (`basecamp modules --show` succeeds), and the unmodelled `note` key survives. Hoisting the decor reset into `section_table` behind `TableParent` is the right shape — `coerce_to_table` having exactly one caller is what makes the guard unskippable at a new call site, which was the actual defect. Re-homing the comment onto the header decor rather than just clearing it is better than what I asked for.

2. Trailing comments survive. `pin = "aa2377…"  # held back until the new IDL spec lands`, `port = 3040  # chosen to dodge the corp VPN range`, and a trailing comment on the last key of a section all come back untouched after a write, alongside the own-line comment and the unmodelled `[team.notes]`; the second write is a byte-for-byte no-op. I specifically checked the case the first-line truncation in `set_preserving_suffix` is there for — a trailing comment followed by an own-line comment owned by the next key, and a trailing comment on the last key with footer comments at EOF — and both stay attached to the key they went in with.

3. The docs now match the binary. `printf '\n[[[ broken\n' >> scaffold.toml` gives `error: invalid scaffold.toml: TOML parse error at line 50, column 3`, exit 1, the file byte-identical, and no `.bak-*`. I verified the invariant the docs now rest on: `save_project_config` has exactly two callers (`src/commands/basecamp.rs:267`, `:2123`), both reached from the `load_project()` at `src/commands/basecamp.rs:106`, so `RenderedFresh` really is TOCTOU-only. Scoping the docs to the refusal that ships, and keeping the backup as the internal guard it is, is the right call for this PR.

Nothing blocking. Two things I turned up while probing the new merge contract, both minor and neither a regression against `master`.

---

## 1. (Non-blocking) An array-of-tables at a modelled section is deleted silently

`[[wallet]]` is valid TOML, and the reader tolerates it the same way it tolerated an inline table — `Item::as_table` is `None` for an `ArrayOfTables`, so the parse falls back to the default. The writer then destroys it. On a real project:

```toml
# LOAD-BEARING: two wallets, see issue 500
[[wallet]]
home_dir = ".scaffold/w1"
note = "hand-written"

[[wallet]]
home_dir = ".scaffold/w2"
```

after `basecamp modules --path mods/foo.lgx` (exit **0**, "captured modules" printed):

```toml
[wallet]
home_dir = ".scaffold/wallet"
```

Both entries, the unmodelled `note`, and the comment above them are gone. `[[modules]]` behaves the same once the model carries any module — which `basecamp modules` capture guarantees — so it is reachable through the command this PR's own runbook exercises, not only through a contrived section. (`[[repos]]` is safe: the reader rejects it with `missing [repos.lez]` before the writer runs.)

The cause is `coerce_to_table`'s `None => Table::new()` arm (src/config.rs:1571). Its comment — "nothing to keep, and the parser already ignored it" — is true of the scalar case it was written for, but an `ArrayOfTables` is the one non-table shape that *does* hold user content, and `section_table` has already cleared the key's decor by the time it gets there, so the explanatory comment goes with it.

This is the same class as the inline-table finding, one notch less severe: the output still parses, so the user is not wedged, and `master` discarded the whole file anyway so nothing regresses. But it is squarely against the rule `write_config_into` now documents — a removal driven by a parser that is not total — and the inline path next door was made lossless precisely to honour it. I'd lean toward a hard error (`scaffold.toml: [wallet] is an array of tables; scaffold expects a table`) rather than a lossless merge: there is no sensible merge of two `[[wallet]]` entries into one model, and refusing keeps the file intact, which is the outcome this PR argues for throughout. A test alongside `update_config_rebuilds_a_modelled_key_that_is_not_a_table` would pin whichever you pick.

## 2. (Minor) Comments inside a multi-line array value are still dropped

Adjacent to the suffix fix, and not covered by it — the array is rebuilt wholesale, so only the suffix survives:

```toml
[run.watch]
include = [
  "src/**",  # the sources
  "lib/**",
]
```

comes back as `include = ["src/**", "lib/**"]`. `set_preserving_suffix` carries the comment *after* the closing bracket, so the trailing case your new test covers is fine; it is only comments interior to the array. Same for `post_deploy` and `env_append` lists. Low-frequency, and clearly out of scope for a fix here — but README currently says comments are "left exactly where they were" without qualification, and this is the one shape where they are not. A clause on the README bullet would settle it.

---

One observation, not a finding: `toml_edit` normalises CRLF to LF on parse, so a `scaffold.toml` authored on Windows has its line endings rewritten wholesale on every save. That predates this PR (the from-scratch render did the same) and is a library property rather than something to fix here — but `update_config_promotes_a_commented_inline_section_in_a_crlf_file` reads as if the write path is CRLF-faithful, and it isn't. Worth a line in that test's doc comment so the next person doesn't rely on it.

Neither point blocks. The three defects from the last round are genuinely fixed, the tests added for them fail against the pre-fix writer, and the runbook now describes behaviour the binary actually has.

---

weboko (commented):
Reviewed the commits added since my last pass — `dd1c2da`, `c680438`, `86492d2` — and re-ran the runbook checks this PR added to `DOGFOODING.md` B1 against a real project.

## Verification

Built at `86492d2`; `cargo build`, `cargo fmt --check`, `cargo test` all clean (630 + 186 + 8 tests pass), clippy unchanged. Because `basecamp setup` needs Nix with unrestricted GitHub flake access (not available in my container), I drove `save_project_config` through the other writer, `basecamp modules --path <file>.lgx`, which reaches the identical code path.

Every B1 config check passes:

- own-line comment and unmodelled `[team.notes]` survive two rewrites; the second rewrite is byte-identical
- trailing comment on `port = 3040` survives a reassignment
- `port_base` deleted by hand is not resurrected
- a broken file is refused (`invalid scaffold.toml: TOML parse error … line 50, column 3`), left byte-identical, with no `.bak-*`
- a commented inline `circuits = { … }` promotes to `# comment` / `[circuits]` on separate lines, reparses, exactly one definition

And the fix in `86492d2` is real and works. With

```toml
# why the wallet lives outside the default
wallet = { home_dir = "CUSTOM/PATH", note = "keep me" }
```

`86492d2` keeps `home_dir = "CUSTOM/PATH"` and the unmodelled `note`, under one `[wallet]` header with the comment re-homed above it. The same fixture against `c680438` rewrites `home_dir` to `.scaffold/wallet` with exit 0 — the silent loss the commit describes, confirmed rather than taken on trust.

## Blocking: widening the reader wedged the old-schema migration path

`86492d2` routed `detect_old_schema_markers` through `section_like`, so an inline `basecamp = { … }` or `repos = { … }` is now *detected* as carrying old-schema keys. `src/migrate.rs` was not widened to match: it reaches the same sections through `Item::as_table_mut` / `Item::as_table`, which are `None` for an inline table.

- `src/migrate.rs:83` — `doc.get_mut("repos").and_then(Item::as_table_mut)` (the `lssa` → `lez` rename)
- `src/migrate.rs:145` — `doc.get_mut("basecamp").and_then(Item::as_table_mut)` (`pin`/`source`/`lgpm_flake`)
- `src/migrate.rs:215-217` — the same, plus `modules_item.as_table()`, for `[basecamp.modules.*]` → `[modules.*]`

Detection fires, migration is a no-op, and the project becomes permanently unusable. Minimal repro, already at schema v0.2.0 so nothing else is stale:

```toml
basecamp = { modules = { foo = { flake = "./x#lgx", role = "project" } } }

[scaffold]
version = "0.2.0"

[repos.lez]
source = "https://github.com/logos-blockchain/logos-execution-zone.git"
pin = "cf3639d8252040d13b3d4e933feb19b42c76e14a"

[repos.spel]
source = "https://github.com/logos-co/spel.git"
pin = "73fc462eb8f0a4d00f1a846437c627ec2e523f83"
```

```
$ lgs doctor
error: scaffold.toml uses an old schema. Run `logos-scaffold init` to migrate to v0.2.0; existing settings are preserved.
$ lgs init
scaffold.toml at … is already at schema v0.2.0.
$ lgs doctor
error: scaffold.toml uses an old schema. Run `logos-scaffold init` to migrate to v0.2.0; existing settings are preserved.
```

`init` writes nothing (file byte-identical), every other command refuses, and the error names the one command that cannot fix it. There is no way out except hand-editing the file. `repos = { lssa = { … }, lez = { … }, spel = { … } }` behaves the same way.

I bisected this to the PR rather than assuming: `master` and `c680438` both accept the same file and run normally, `86492d2` does not. So this is a regression introduced by the last commit, and it is worse than the pre-existing behaviour it replaced — before, the inline section was silently ignored; now it blocks the CLI outright with unactionable advice.

The PR's own framing is what makes this a defect rather than a judgement call: it argues the two syntaxes are semantically identical and that the reader has no business distinguishing them. Applying that to detection but not to the rewriter leaves the two halves disagreeing about what the file says, which is exactly the class of bug the PR set out to close.

`toml_edit` 0.22 has `Item::as_table_like_mut`, so `migrate.rs` can be widened the same way `config.rs` was. One caveat for whichever way you go: the `[basecamp.modules.*]` move clones `Table` values (`moved_modules: Vec<(String, Table)>`), so an inline child needs converting rather than cloning — probably by promoting the section to a real table before migrating it, which would also mean the file that comes out is normalised the way the writer normalises everything else.

Please also add a regression test for the combination. The existing coverage exercises inline sections against the reader and the writer, but nothing drives an inline old-schema section through `detect_old_schema` → `migrate` → re-parse, which is why this got through a fairly thorough test pass.

## Non-blocking

- `src/config.rs:1107`, `:1481`, `:1524` — `profiles.retain(|_, item| !item.as_table().is_some_and(Table::is_empty))` has the same blind spot: an empty profile written inline (`profiles = { alice = {} }`) is not pruned, so a profile reset to defaults survives the rewrite that is supposed to drop it. Much smaller than the above, and no data is lost.
- The README section added here says key ordering is "left exactly where they were", but a section written inline is now normalised to `[header]` form on the next write. That is defensible behaviour and the `coerce_to_table` docs explain it well; it just isn't in the user-facing docs, and it is the one case where a rewrite visibly changes something the user wrote. Worth a sentence in the README block and in the B1 expectations.

Everything else in this PR held up under testing. The merge-contract work, the removal branches, and the backup/fail-closed path all behave as documented, and B1 is a genuinely good runbook addition. It is the detection/migration split that needs fixing before this lands.


---

weboko:
I just noticed this battle of bots 🤣

PR #269, opened 2026-09-12
fix(deploy): discover guest binaries under methods/guest/target

### LINKED ISSUE #231: `lgs deploy` guest-binary auto-discovery misses `methods/guest/target/…`

aegonmyy:
A guest crate that carries its own `[workspace]` builds independently of the parent and emits its `.bin` artefacts to `methods/guest/target/riscv32im-risc0-zkvm-elf/<profile>/`, which the two existing search roots never walked. `deploy` then reported `missing program binary` for a guest that had just built successfully, and the only workaround was `--program-path` with the full path spelled out.

This adds `methods/guest/target` to `GUEST_BIN_SEARCH_ROOTS` so the walk covers the self-workspaced guest layout. The `docker` profile directory has no `release` component and is matched through the existing debug fallback bucket, which already accepts paths without one.

Covered by `discovers_guest_owned_workspace_layout`, verified to fail against the old constant. No API-surface change: the discovery walk is shared by the CLI and `logos_scaffold::api`, so both front ends pick the new root up from the same path.

Closes #231 (my own issue, filed while building an LP-0017 submission against a guest crate with its own workspace).


weboko (commented):
Verified this end-to-end rather than by reading. The change is correct and the regression test is genuine; my only substantive concern is that it duplicates an already-open PR.

## Verification

Built at `8644f8f`. `cargo build`, `cargo fmt --check`, `cargo test` clean (586 + 186 + 8 pass), clippy at 61 warnings (60 on master, the delta is in an untouched file).

`discovers_guest_owned_workspace_layout` is a real regression test — reverting `GUEST_BIN_SEARCH_ROOTS` to the two-element constant makes it fail at `deploy.rs:1094` and pass again with the fix.

Then the user-visible path, against a real localnet (`lgs setup`, RISC Zero toolchain installed, sequencer running). I built the template's guest, moved its ELFs to `methods/guest/target/riscv32im-risc0-zkvm-elf/release/`, and deleted `target/riscv-guest/` so the self-workspaced root was the only place the binaries existed:

```
# master
{"deploys":[{"error":"missing program binary","program":"hello_world","status":"failed"}]}
error: deploy failed: missing program binary

# this PR
{"deploys":[{"program":"hello_world","program_id":"211b41fb…2fc60f1f","status":"submitted"}]}
```

So the reported symptom reproduces on `master` and is fixed here, through the wallet to a live sequencer. (The pinned sequencer then crashes on the deploy inscription — `Sequencer failed unexpectedly: Main loop exited unexpectedly`. That is not this PR: it reproduces identically on the `master` binary with `--program-path`, which bypasses discovery entirely, for the same `program_id`.)

## This duplicates #262

#262, "fix(deploy): discover guest binaries from an independently-workspaced guest crate", makes the same change: the same three-element `GUEST_BIN_SEARCH_ROOTS`, a rewritten doc comment on the same constant, and a regression test built from the same `methods/guest/target/riscv32im-risc0-zkvm-elf/docker` fixture. Both close #231, both are open, and both touch only `src/commands/deploy.rs`.

They will conflict with each other — same constant, same hunk — so whichever lands first leaves the other needing a rebase for no benefit. Worth deciding between them before either merges. On the diffs alone, this one is the better of the two: the assertion pins the full discovered path rather than just the filename, so it would still fail if discovery started matching on a different root, and the doc comment tracks the three roots against `METHODS_DIR` more explicitly. #262 is also a commit behind master. If you agree, closing #262 in favour of this one (with a note on the issue) is the tidier outcome; the reverse is fine too, but the duplication should not survive the merge.

## Non-blocking: the `docker` profile ranks below a stale `release` build

The doc comment and the new test both establish `docker` as a supported profile directory, picked up through the debug fallback bucket because the path has no `release` component. That is true and the test passes, but the fallback is unconditional: any `release` match wins over any non-`release` match regardless of depth or mtime. So when both exist:

```
methods/guest/target/riscv32im-risc0-zkvm-elf/docker/my_program.bin   (fresh)
target/riscv-guest/riscv32im-risc0-zkvm-elf/release/my_program.bin    (stale)
```

`discover_program_binaries` returns the stale `release` one. I confirmed this with a test against this branch — the stale ELF is what `deploy` would submit.

This is pre-existing, not introduced here, and out of scope for a fix in this PR. I raise it because this is the PR that makes `docker` an explicitly tested layout, and #260 (opt-in deterministic guest builds) turns `docker` into a path users will actually take — at which point "deploy silently submitted the previous non-deterministic ELF" becomes a plausible and hard-to-notice failure. A follow-up issue would be enough.

## Minor

The self-workspaced guest layout is now supported but is not mentioned anywhere in the docs — `DOGFOODING.md:238` and README both describe only the default-template `target/riscv-guest/...` tree. Neither claims to enumerate every layout, so nothing is stale; a sentence in D3 noting that a guest crate with its own `[workspace]` is discovered too would save the next person the debugging round-trip that produced #231.

No concerns on security, error handling, or API surface: the new root is a fixed project-relative path, the walk still runs with `follow_links(false)`, program names are still validated before they reach the search map, and both the CLI and `logos_scaffold::api` pick the root up from the same shared function.


---

aegonmyy:
@weboko

